PR #1, opened 2018-12-11
Add circle CI to app operator

tomahawk28:
I can't assign the team in here,

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,8 @@ jobs:
         chmod +x ./architect
         ./architect version
     - run: ./architect build
+    - store_test_results:
+        path: /tmp/results
     - deploy:
         command: |
           if [ "${CIRCLE_BRANCH}" == "master" ]; then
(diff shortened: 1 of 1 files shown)


rossf7:
In the repo settings you need to add the `Employees` team as well as `bots`.

.circleci/config.yml
```diff
@@ -0,0 +1,19 @@
...
+    machine: true
+    steps:
+    - checkout
+
+    - run: |
+        wget -q $(curl -sS -H "Authorization: token $RELEASE_TOKEN" https://api.github.com/repos/giantswarm/architect/releases/latest | grep browser_download_url | head -n 1 | cut -d '"' -f 4)
+        chmod +x ./architect
+        ./architect version
+    - run: ./architect build
+    - deploy:
```
pipo02mix: What about add support JUnit reports? we have in other operator and it helps to give more info when something fails
```suggestion
    - store_test_results:
        path: /tmp/results
    - deploy:
```

PR #3, opened 2018-12-14
Version endpoint

tomahawk28:
Now added `version` endpoint

```
> curl http://127.0.0.1:8000/
{"description":"The app-operator manages apps in Kubernetes clusters.","git_commit":"n/a","go_version":"go1.11.2","name":"app-operator","os_arch":"darwin/amd64","source":"https://github.com/giantswarm/app-operator","version_bundles":[{"changelogs":[{"component":"app-operator","description":"Added initial version.","kind":"added"}],"components":[],"name":"app-operator","version":"0.1.0"}]}
```


service/healthz/service.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// Config represents the configuration used to create a healthz service.
+type Config struct {
+	K8sClient kubernetes.Interface
+	Logger    micrologger.Logger
+}
+
+// New creates a new configured healthz service.
+func New(config Config) (*Service, error) {
```
xh3b4sd: The health service should not be necessary at all. I think it is nowhere used in the code. All we need is the health endpoint. It provides the general availability check which is all we want here. Reason is that we do not want to have the k8s health checks in the operators in order to prevent cascading failures. If you like you could document this for the coming generations of developers in https://github.com/giantswarm/sig-operator, as you just ran into the topic. <3

Where did you get the usage of the k8s health service from? When this is used somewhere still we should remove it from there as well.

tomahawk28: @xh3b4sd It's from chart-operator, https://github.com/giantswarm/chart-operator/blob/master/service/healthz/service.go 
Is it possible to create an endpoint w/o service?

rossf7: @xh3b4sd Thanks I think I missed this while I was away. 

@tomahawk28 You can remove both the service and endpoint for `healthz`. Please also remove it from `chart-operator`. 

Documenting in the `sig-operator` repo is a good idea and helps everyone but especially future devs.

xh3b4sd: The endpoint should stay. Only the service can be removed. That works because the basic service is defaulted if no service is provided. See https://github.com/giantswarm/microendpoint/blob/f77c569259aef9cea5e38d0d7d94bf1d7d07d136/endpoint/healthz/endpoint.go#L48-L59.

service/service.go
```diff
@@ -1,12 +1,13 @@
 package service
 
 import (
-	"github.com/spf13/viper"
 	"sync"
 
 	"github.com/giantswarm/app-operator/flag"
+	"github.com/giantswarm/microendpoint/service/version"
```
rossf7: app-operator import above should be in a 3rd block.

PR #4, opened 2018-12-14
Add dockerfile

tomahawk28:


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM alpine:latest
```
rossf7: We don't use latest tags because we want building our images to be reproducible. We should use the `3.8` tag.

tomahawk28:
I created quay repo but I'm not sure whether it should be private or not 
https://quay.io/repository/giantswarm/app-operator

README.md
```diff
@@ -0,0 +1,7 @@
+[![CircleCI](https://circleci.com/gh/giantswarm/app-operator.svg?style=shield)](https://circleci.com/gh/giantswarm/chart-operator) [![Docker Repository on Quay](https://quay.io/repository/giantswarm/chart-operator/status "Docker Repository on Quay")](https://quay.io/repository/giantswarm/app-operator)
```
indyfree: The badges partly reference `chart-operator` instead of app-operator

tomahawk28:
I'm still getting this error from quay. 
```A build step failed: ADD failed: stat /var/lib/docker/tmp/docker-builder156319217/app-operator: no such file or directory```

I thought `architect` do that job but looks like it didn't add `app-operator`

rossf7:
The Quay repo is still configured as an automated build. We don't use those because `architect` does our image builds.

Deleting the build trigger should fix it. In future you just need to create the Quay image as an empty repo.

screen shot 2018-12-17 at 12 42 05 pm

tomahawk28:
it's now updated. https://quay.io/repository/giantswarm/app-operator?tab=tags 
Can I merge it?

tomahawk28:
🤦‍♂️ Sorry, but thanks for updating the setting

PR #5, opened 2018-12-19
Add App controller

tomahawk28:
Just now added a bare controller. 
minikube is not working at the moment from my localhost, so can't check CRD creation.


tomahawk28:
Please check 46d185dadadb2a21a9c0003f300b7c7f266cf5e9 only if you want to see only controller part.

tomahawk28:
If no resource app resource is there, there is an interface issue between controller, so it should be there to solve that issue. 
I solved the issue with minikube, I just totally delete `~/.minikue` and restart it :)

tomahawk28:
And my thought was reviewers can focus on code changes on operator sources, not in `vendor`. 
That's why I propose looking into 46d185d than other commits.

rossf7:
Good deleting minikube and recreating it is a good strategy. We do that when running e2e tests locally.

What error do you get if there is no resource? I'm pretty sure I've created a resource set without a resource before.

tomahawk28:
If no resource, `operatorkit` reported an issue (run-time error) it cannot run with empty resource_set. So I added up everything that it need.

rossf7:
> And my thought was reviewers can focus on code changes on operator sources, not in vendor.
That's why I propose looking into 46d185d than other commits.

Its very unusual to look in vendor for a PR. Sometimes I check the Gopkg.toml but usually thats it.

All the code in the PR needs to be reviewed. So looking at individual commits doesn't really help. How we create PRs and reviews takes some getting used to. Lets chat next time we sync.

rossf7:
Thanks! That should be fixable. I'll give it a try after standup.

rossf7:
@tomahawk28 You're right the empty structure is needed. Its this error check in operatorkit which is needed. Checking on the rest of the code now.

https://github.com/giantswarm/operatorkit/blob/894bebfb2a6835b15394c329617a442a0725939f/controller/resource_set.go#L55-L57

Gopkg.toml
```diff
@@ -31,10 +31,18 @@ required = [
   "github.com/docker/distribution/reference",
   "k8s.io/api/core/v1",
   "k8s.io/apimachinery/pkg/apis/meta/v1",
-  "k8s.io/apiserver/pkg/endpoints/request"
+  "k8s.io/apiserver/pkg/endpoints/request",
+  "github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
```
rossf7: `apiextensions` isn't needed. The required block is only needed when a package isn't used in our code.

service/controller/app.go
```diff
@@ -0,0 +1,122 @@
+package controller
+
+import (
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
+	"github.com/giantswarm/app-operator/service/controller/v1"
```
rossf7: This import should be in a separate block.

service/controller/app.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/giantswarm/operatorkit/client/k8scrdclient"
+	"github.com/giantswarm/operatorkit/controller"
+	"github.com/giantswarm/operatorkit/informer"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	"k8s.io/client-go/kubernetes"
+	//"github.com/giantswarm/app-operator/service/controller/v1"
```
rossf7: This should be removed.

service/controller/app.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+const AppControllerSuffix = "-app"
```
rossf7: I don't think this is needed. Only chart-operator seems to have it. Lets remove it and just use the project name as the controller name.

service/controller/app.go
```diff
@@ -0,0 +1,122 @@
...
+
+type App struct {
+	*controller.Controller
+}
+
+func NewApp(config AppConfig) (*App, error) {
+	var err error
+
+	if config.G8sClient == nil {
+		return nil, microerror.Maskf(invalidConfigError, "config.G8sClient must not be empty")
```
rossf7:
```suggestion
return nil, microerror.Maskf(invalidConfigError, "%T.G8sClient must not be empty", config)
```

This is outdated. We should use the new approach. It makes it easier to read stack traces when debugging.

service/controller/v1/resource/app/resource.go
```diff
@@ -0,0 +1,80 @@
+package app
```
rossf7: This resource should be called `chart` not `app`. This is because app-operator will watch app CRs and create chart CRs in the tenant cluster.

service/controller/v1/resource_set.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/giantswarm/app-operator/service/controller/v1/key"
+	"github.com/giantswarm/app-operator/service/controller/v1/resource/app"
+)
+
+const (
+	// ResourceRetries presents number of retries for failed Resource
+	// operation before giving up.
+	ResourceRetries uint64 = 3
+)
```
rossf7: This constant isn't used anymore. So we should remove it.

service/service.go
```diff
@@ -3,10 +3,16 @@ package service
 import (
 	"sync"
 
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
+	"github.com/giantswarm/app-operator/service/controller"
```
rossf7: Should be in the last block.

service/service.go
```diff
@@ -79,5 +138,7 @@ func New(config Config) (*Service, error) {
 func (s *Service) Boot() {
 	s.bootOnce.Do(func() {
 		// Insert service startup logic here.
+		// Start the controller.
```
rossf7: Lets remove the `Insert service startup logic here.` comment above. Its not needed anymore.

service/controller/v1/resource/app/resource.go
tomahawk28: Changed 👍

service/controller/app.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/giantswarm/operatorkit/client/k8scrdclient"
+	"github.com/giantswarm/operatorkit/controller"
+	"github.com/giantswarm/operatorkit/informer"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/giantswarm/app-operator/service/controller/v1"
+)
+
+type AppConfig struct {
```
rossf7:
```suggestion
 type Config struct {
```

rossf7: This is outdated.

service/controller/v1/key/key.go
```diff
@@ -0,0 +1,45 @@
+package key
+
+import (
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/microerror"
+)
+
+const (
+	VERSION_BUNDLE = "giantswarm.io/version-bundle"
```
rossf7:
```suggestion
 	versionBundleLabel = "giantswarm.io/version-bundle"
```

service/controller/v1/key/key.go
```diff
@@ -0,0 +1,45 @@
...
+
+	if customObjectPointer == nil {
+		return v1alpha1.App{}, microerror.Maskf(emptyValueError, "empty value cannot be converted to CustomObject")
+	}
+
+	return *customObjectPointer, nil
+}
+
+func VersionBundleVersion(customObject v1alpha1.App) string {
+	if val, ok := customObject.ObjectMeta.Annotations[VERSION_BUNDLE]; ok {
```
rossf7: Cool that it uses the annotation. Just needs the constant renaming.

service/controller/v1/resource/chart/resource.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/giantswarm/operatorkit/controller"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// Name is the identifier of the resource.
+	Name = "appv1"
+)
+
+// Config represents the configuration used to create a new app resource.
```
rossf7: Should be chart resource. There are a few comments that need updating.

service/controller/app.go
tomahawk28: @rossf7 You mean we don't need to change it as `Config`?

service/controller/v1/key/key.go
rossf7: Actually I think this should be `versionBundleAnnotation`. 😆 Missed it as this is one of our first uses of annotations!

service/controller/app.go
rossf7: Sorry I mean we should rename to `Config`. `AppConfig` is the old naming convention.

Gopkg.toml
```diff
@@ -35,6 +35,13 @@ required = [
 ]
 
 
+[[constraint]]
+  branch = "master"
+  name = "github.com/giantswarm/operatorkit"
+
+[[constraint]]
+  branch = "catalog_app_deployment"
+  name = "github.com/giantswarm/apiextensions"
```
rossf7: Nitpick but the packages in Gopkg.toml should be in alphabetical order.

PR #7, opened 2019-01-03
new helm chart for app operator

tomahawk28:
Adding helm client for app-operator

Issues: https://github.com/giantswarm/giantswarm/issues/4845


helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,110 @@
...
+  verbs:
+    - get
+    - list
+    - patch
+    - update
+    - watch
+- apiGroups:
+    - core.giantswarm.io
+  resources:
+    - charts/status
```
puja108: shouldn't this be `apps/status` and above be `application.giantswarm.io`?

helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,110 @@
...
+  - "*"
+- apiGroups:
+    - application.giantswarm.io
+  resources:
+    - apps
+  verbs:
+    - get
+    - list
+    - patch
+    - update
```
puja108: I do not think we need patch and update on `apps` as the operator should not be allowed to change it's own CRs, but only the status field below

helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,110 @@
...
+  - update
+  - watch
+- apiGroups:
+  - core.giantswarm.io
+  resources:
+  - charts/status
+  verbs:
+  - create
+  - patch
+  - update
```
puja108: chart/status is set by chart-operator and should not be changed by app-operator, it should only be able to get the status (and maybe list?) right?

also this and above here would be only in case it creates chart CRs in control plane (which we want in future), but we need the same in tenant-clusters too

helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,110 @@
...
+    - watch
+- apiGroups:
+    - core.giantswarm.io
+  resources:
+    - charts/status
+  verbs:
+    - create
+    - patch
+    - update
+- apiGroups:
```
puja108: don't we also need get (and maybe list?) or is that included in the above rules of charts?

helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,110 @@
...
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - pods
+  verbs:
+  - "get"
+  - "list"
+- apiGroups:
```
puja108: why do we need get and list on CMs and pods? afaik we do not have anything to do there in app operator, we might have a CM with the global config to configure it, but that should be a single CM not all, also we need access to appCatalog CRs

helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,110 @@
...
+  - ""
+  resources:
+  - pods/portforward
```
puja108: why?

puja108: I think this is because you came from chart-operator RBAC, right? I think we need to also remove all of the below

helm/app-operator-chart/templates/deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+        searches:
+        - giantswarm.svc.cluster.local
+        - svc.cluster.local
+        - cluster.local
+        options:
+        - name: ndots
+          value: "5"
+      containers:
+      - name: {{ .Values.name }}
+        image: "quay.io/{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
kopiczko:
```suggestion
        image: "{{ .Values.image.registry }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```

helm/app-operator-chart/templates/rbac.yaml
tomahawk28: Oops, should delete that now.

helm/app-operator-chart/templates/rbac.yaml
kopiczko: Problem is `finalizers` and `annotations`.

helm/app-operator-chart/templates/rbac.yaml
kopiczko: Yeah probably copy/paste. `chart-operator` does it fo talk with tiller.

helm/app-operator-chart/templates/rbac.yaml
puja108: we do not have any `finalizers` and `annotations` in form of resource version we have only in chart CRD, so I think it's ok to leave those out as long as we do not use that functionality.

helm/app-operator-chart/templates/rbac.yaml
puja108: btw, maybe we add a note/comment here in the YAML that this is only for creating chart CRs in control plane (in the future) and not for tenant clusters as those RBAC roles are managed inside the tenants

helm/app-operator-chart/templates/rbac.yaml
rossf7: Yes this is chart-operator specific because it needs to access tiller via port forwarding.

helm/app-operator-chart/templates/rbac.yaml
rossf7: We will need this later as app-operator will need to manage configmaps for chart-operator. But we should keep this as the minimal permissions for now and add it later when we need it.

helm/app-operator-chart/templates/rbac.yaml
rossf7: I would actually use '*' here. The operators role is to manage chart CRs. This is what we do in cluster-operator.

```
  - apiGroups:
      - application.giantswarm.io
    resources:
      - charts
    verbs:
      - "*"
```

helm/app-operator-chart/templates/rbac.yaml
tomahawk28: I will delete this permission at this moment

puja108: ah, right, CMs (and maybe secrets at some point) it will need to create for charts at least for charts in Control Plane we would need that

rossf7: Yes for the chart CRs I think we should make sure it works for tenant and control plane clusters.

If the chart CR has a `kubeconfig` spec it will use it to access the tenant cluster. If its empty then it should create the chart CR in the control plane cluster the operator is running in.

tomahawk28:
I'm getting this error from CI, which is not clear to me. 
```
2019/01/03 14:48:20 retrying task 'helm-push' due to error (exit status 2)
usage: cnr [-h]
           {show,inspect,logout,plugins,helm,pull,list,delete-package,version,push,login,config,channel}
           ...
cnr: error: 403 Client Error: FORBIDDEN for url: https://quay.io/cnr/api/v1/packages/giantswarm/app-operator-chart?force=true
```
Anyhing I need to add in yaml?

rossf7:
@tomahawk28 You need to create a CNR repo in Quay for app-operator.

> Create a application repository (for the helm chart)

https://github.com/giantswarm/giantswarm/blob/master/ops-recipes/container-registry.md#setting-up-a-repository

helm/app-operator-chart/templates/rbac.yaml
puja108: hmm, wait, right, it should have a kubeconfig also for control plane cluster, so it does not even need any of the chart related roles in here

helm/app-operator-chart/templates/rbac.yaml
puja108: as per https://github.com/giantswarm/app-operator/pull/7#discussion_r245029875 we do not need any chart CR related rights in the role as they should be bound to the kubeconfig it gets not to the operator itself

helm/app-operator-chart/templates/rbac.yaml
puja108: but shouldn't it also get a kubeconfig for control plane? I thin that would be nicer than splitting it between kubeconfig or SA RBAC

helm/app-operator-chart/templates/rbac.yaml
tomahawk28: @puja108 But don't `app-operator` need to have creation permission for `charts` CR at least?

puja108: it will get it through `kubeconfig` for the cluster it needs to, not from this RBAC role

helm/app-operator-chart/templates/rbac.yaml
puja108: hmm, ok, if you think that's nicer we can go that way, but then let's have notes/comments in thise RBAC yaml to point out which rules are for CP chart functionality, or better even let's split the RBAC role into one for normal app-operator functionality and one for applying chart CRs, that way we can have the latter both bound to the inCluster SA config on control plane but also bound in tenant clusters to the kubeconfig. WDYT? That way if we have changes, we just sync the two roles, for app-operator the role binding can just bind both roles.

rossf7:
Yes I like splitting the roles. Makes it clearer what is going on.

helm/app-operator-chart/templates/rbac.yaml
rossf7: In the rbac.yaml we should have 2 cluster roles, `app-operator` and `app-operator-chart`. The `-chart` role should just have the permissions for managing chart CRs.

The reason for this is when managing chart CRs inside tenant clusters we will use the `kubeconfig` rather than the `-chart` role.

helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,137 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: {{ .Values.chart }}
```
rossf7:
```suggestion
  name: {{ .Values.name }}-chart
```

I would just add the suffix rather than defining another value. `values.chart` isn't a clear name for me. 

Also can you make the app-operator role the first role here. As it is the most important.

helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,137 @@
...
+  resources:
+    - appcatalogs
+  verbs:
+    - get
+    - list
+    - watch
+- apiGroups:
+    - application.giantswarm.io
+  resources:
+    - apps
```
rossf7: This looks strange to me. `apps` is in the role multiple times. Any dupes need to be removed.

tomahawk28:
Do you have any objection to merging? :)

helm/app-operator-chart/values.yaml
```diff
@@ -0,0 +1,27 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+name: app-operator
+namespace: giantswarm
+port: 8000
+
+replicas: 1
+
+clusterDNSIP: 172.31.0.10
```
rossf7: This should be removed.

helm/app-operator-chart/values.yaml
```diff
@@ -0,0 +1,27 @@
...
+resources:
+  requests:
+    cpu: 250m
+    memory: 250Mi
+  limits:
+    cpu: 250m
+    memory: 250Mi
+
+cnr:
+  address: https://quay.io
```
rossf7: So should this as we're not using CNR.

helm/app-operator-chart/templates/configmap.yaml
```diff
@@ -0,0 +1,25 @@
...
+data:
+  config.yaml: |
+    server:
+      enable:
+        debug:
+          server: true
+      listen:
+        address: 'http://0.0.0.0:{{ .Values.port }}'
+    service:
+      cnr:
```
rossf7: We're not using CNR so this should be removed.

helm/app-operator-chart/README.md
```diff
@@ -0,0 +1,14 @@
+# app-operator Helm Chart
```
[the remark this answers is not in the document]
rossf7: I still don't think we need this README.

helm/app-operator-chart/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart for the app-operator
+name: app-operator-chart
+version: 0.1.0
```
indyfree:
```suggestion
version: 0.1.0

```

indyfree: Small nitpick: New line at EOF was missing :)

helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -0,0 +1,129 @@
...
+    namespace: {{ .Values.namespace }}
+roleRef:
+  kind: ClusterRole
+  name: {{ .Values.name }}
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ .Values.name }}
```
puja108:
```suggestion
  name: {{ .Values.name }}-chart
```

the name of the k8s resource (here the binding) must always be unique, otherwise k8s won't accept it or overwrite

helm/app-operator-chart/templates/rbac.yaml
puja108: this is resolved right?

tomahawk28: Yes, it's up here https://github.com/giantswarm/app-operator/blob/790ca67fdaa50439485100352e472671a7c4a3d9/helm/app-operator-chart/templates/rbac.yaml#L96-L97

PR #8, opened 2019-01-04
add appCatalog CRD on running 

tomahawk28:
After running operator, new 2 CRDs will be created inside k8s.
`> kubectl get crd`

```
NAME                                    AGE
appcatalogs.application.giantswarm.io   1m
apps.application.giantswarm.io          1d
```
The pain point in here: 
Currently, all our operator structures only allow you to create one CRD at operator booting time.  
But app-operator itself have to create two new CRDs, which have to make the code a bit dirty to make it works. 

Why dirty?: 
I stored `CRDClient` instance in `app` instance which only supposed to be stored in `controller`,
so I can run it inside `Service.Boot` method.

I'm not sure that this approach is the best practice so please let me know if it's not!

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -1,7 +1,12 @@
 package service
 
 import (
+	"context"
+	"fmt"
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/backoff"
 	"sync"
+	"time"
 
 	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
 	"github.com/giantswarm/microendpoint/service/version"
@@ -134,10 +139,19 @@ func New(config Config) (*Service, error) {
 	return newService, nil
 }
 
+func (s *Service) createAppCatalogCRDResource() {
+	ctx := context.Background()
+	err := s.appController.CRDClient.EnsureCreated(ctx, v1alpha1.NewAppCatalogCRD(), backoff.NewMaxRetries(7, 1*time.Second))
+	if err != nil {
+		panic(fmt.Sprintf("%#v\n", microerror.Maskf(err, "service.createAppCatalogCRDResource")))
+	}
+}
+
 // Boot starts top level service implementation.
 func (s *Service) Boot() {
 	s.bootOnce.Do(func() {
 		// Start the controller.
+		s.createAppCatalogCRDResource()
 		go s.appController.Boot()
 	})
 }
--- a/service/controller/app.go
+++ b/service/controller/app.go
@@ -26,6 +26,7 @@ type Config struct {
 
 type App struct {
 	*controller.Controller
+	*k8scrdclient.CRDClient
 }
 
 func NewApp(config Config) (*App, error) {
@@ -114,6 +115,7 @@ func NewApp(config Config) (*App, error) {
 
 	c := &App{
 		Controller: appController,
+		CRDClient:  crdClient,
 	}
 
 	return c, nil


service/service.go
```diff
@@ -1,7 +1,12 @@
 package service
 
 import (
+	"context"
+	"fmt"
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/backoff"
 	"sync"
+	"time"
```
rossf7: Imports https://github.com/giantswarm/fmt/blob/master/go/imports.md.

service/service.go
```diff
@@ -134,10 +139,19 @@ func New(config Config) (*Service, error) {
 	return newService, nil
 }
 
+func (s *Service) createAppCatalogCRDResource() {
```
rossf7:
```suggestion
func (s *Service) ensureAppCatalogCRDCreated() {
```

service/service.go
```diff
@@ -134,10 +139,19 @@ func New(config Config) (*Service, error) {
...
+	ctx := context.Background()
+	err := s.appController.CRDClient.EnsureCreated(ctx, v1alpha1.NewAppCatalogCRD(), backoff.NewMaxRetries(7, 1*time.Second))
```
rossf7: Why these backoff settings?

I would use `backoff.NewExponential(2*time.Minute, 5*time.Second)` here.

kopiczko: Or ShortMaxWait & ShortMaxInterval: https://github.com/giantswarm/backoff/blob/4763148864dc4815b691a69b4f60b4f4a8f88dad/backoff.go#L10

xh3b4sd:
Can you show more of the logs? This does not really give a good idea about what might be missing.

tomahawk28:
Sure, @xh3b4sd 
```
D 01/07 14:25:23 creating out-cluster config | operatorkit/client/k8srestconfig/k8s_rest_config.go:134
D 01/07 14:25:23 running server at http://127.0.0.1:8000 | microkit/server/server.go:308
D 01/07 14:25:23 ensuring custom resource definition exists | operatorkit/controller/controller.go:322
D 01/07 14:25:24 ensured custom resource definition exists | operatorkit/controller/controller.go:329
D 01/07 14:25:24 booting informer | operatorkit/controller/controller.go:333
D 01/07 14:25:24 booting collector | exporterkit/collector/set.go:48
D 01/07 14:25:24 registering collector | exporterkit/collector/set.go:55
D 01/07 14:25:24 describing metrics | exporterkit/collector/set.go:99
D 01/07 14:25:24 described metrics | exporterkit/collector/set.go:121
D 01/07 14:25:24 registered collector | exporterkit/collector/set.go:63
D 01/07 14:25:24 booted collector | exporterkit/collector/set.go:67
D 01/07 14:25:24 booted informer | operatorkit/controller/controller.go:340
D 01/07 14:25:24 starting list-watch | operatorkit/controller/controller.go:363
W 01/07 14:25:24 retrying backoff in '552.330144ms' due to error | backoff/notifier.go:13
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/informer/informer.go:286:
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/informer/informer.go:275:
	the server could not find the requested resource (get apps.application.giantswarm.io)
ERROR: logging before flag.Parse: E0107 15:25:56.295493    2393 streamwatcher.go:109] Unable to decode an event from the watch stream: net/http: request canceled (Client.Timeout exceeded while reading body)
ERROR: logging before flag.Parse: E0107 15:26:26.301194    2393 streamwatcher.go:109] Unable to decode an event from the watch stream: net/http: request canceled (Client.Timeout exceeded while reading body)
```
Please ignore that `ERROR` messages since it was already happening even before this PR changes.

rossf7:
@tomahawk28 The actual error is `the server could not find the requested resource (get apps.application.giantswarm.io)`. Does the app CRD exist in your cluster?

The backoff should retry until the CRD is created. So some errors may occur but eventually it will succeed. I would look for errors in the CRD creation.

tomahawk28:
Both CRD had been created at the same time. So I guess it's retrying to create it.

tomahawk28:
@rossf7 But we can keep this change now, right? we don't need to create a controller for now I guess.

tomahawk28:
@rossf7 @giantswarm/sig-operator can I go with this change? thanks :)

xh3b4sd:
I am a bit confused here. Usually the `operatorkit` framework does the boot and initialization of the CRD when you boot the controller, so there is no need to do it on your own. I think the approach is a bit off as I understand the code here.

rossf7:
@xh3b4sd Yes I think the current design is off. We are still iterating and improving it. I think what we actually want it an app-catalog-operator. It then creates the appCatalog CRD.

There are 2 issues. 

- Basic one is we need to create the appCatalog CRD which this PR was to address.
- More importantly we need to store the index.yaml for the Helm catalog in a configmap so we can use it in the front end (@oponder is working on that).

@tomahawk28 Please close this PR. Sorry about the confusion.

In terms of communication I think its better if @giantswarm/team-batman handle this internally. Its better to communicate with the SIG once the design is clear.

xh3b4sd:
My last overview was that we agreed on having one operator. This is `app-operator`. That is cool. It can have multiple frameworks. That is fine. When you think this is not right or there are other hurdles, let's maybe hang for five minutes and sort that out.

rossf7:
@xh3b4sd Having a separate operator is just a proposal. If we change the design we'll update the spec and we can discuss in the PR. 

Using a 2nd controller does solve the problem but my priority is completing the `chart` resource in chart-operator and the `app` resource in app-operator. Better to park this and we can hang if necessary.

@tomahawk28 Its better to continue with the `kubeconfig` library which we need for the `app` resource.

rossf7:
@xh3b4sd and I chatted. Adding the 2nd controller will do what we need and both CRDs are closely related.

@tomahawk28 Lets discuss next time we chat. It will be easier to discuss on a call.

kopiczko:
Will the app catalog be watched and reconciled?

PR #11, opened 2019-01-15
Add appcatalog controller 

tomahawk28:


service/controller/appcatalog/v1/resource/appcatalog/resource.go
```diff
@@ -0,0 +1,80 @@
+package appcatalog
```
rossf7: Lets rename this resource to `index`. We want to fetch the index.yaml for the catalog and save it to a configmap. This is so we can use it in our APIs and display in Happa. 
https://github.com/giantswarm/sample-catalog/blob/master/index.yaml 

For now we can just add the empty resource.

cc @puja108 @oponder

service/controller/appcatalog/v1/resource/appcatalog/resource.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange interface{}) error {
+	return nil
+}
+
+func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange interface{}) error {
+	return nil
+}
+
+// New creates a new configured chart resource.
```
rossf7: Comment will need to be updated.

rossf7:
Can you also remove this error check from the service. It looks a bit weird.

https://github.com/giantswarm/app-operator/blob/7d32d69339b0c2e998157bf96314f2d042be03f6/service/service.go#L56-L60

service/controller/appcatalog/v1/resource/appcatalog/resource.go
tomahawk28: I don't get that, in diagram that I saw it before, it doesn't have any CR named `index`.

rossf7: The CRD is appCatalog the resources are named after what the operator needs to do. In this case we need to save the index.yaml so calling it index makes sense.

@giantswarm/team-batman Chime in if you disagree.

service/controller/appcatalog/v1/key/key.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/microerror"
+)
+
+const (
+	versionBundleAnnotation = "giantswarm.io/version-bundle"
+)
+
```
rossf7: Lets add some tests for these. It's good for the `key` package to have coverage for when we need to change the CRDs.

service/controller/appcatalog/v1/key/key.go
```diff
@@ -0,0 +1,37 @@
...
+func AppCatalogTitle(customObject v1alpha1.AppCatalog) string {
+	return customObject.Spec.Title
+}
+
+// ToCustomObject converts value to v1alpha1.App and returns it or error
```
rossf7:
```suggestion
// ToCustomObject converts value to v1alpha1.AppCatalog and returns it or error
```

service/controller/appcatalog/v1/resource/index/resource.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/giantswarm/operatorkit/controller"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// Name is the identifier of the resource.
+	Name = "indexv1"
+)
+
+// Config represents the configuration used to create a new chart resource.
```
rossf7:
```suggestion
// Config represents the configuration used to create a new index resource.
```

rossf7: There are several references to `chart` here. Please make sure they all get updated.

service/controller/appcatalog/v1/resource/index/resource.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange interface{}) error {
+	return nil
+}
+
+func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange interface{}) error {
+	return nil
+}
+
+// New creates a new configured appCatalog resource.
```
rossf7:
```suggestion
// New creates a new configured index resource.
```

service/controller/appcatalog/v1/resource_set.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/giantswarm/operatorkit/controller/resource/metricsresource"
+	"github.com/giantswarm/operatorkit/controller/resource/retryresource"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/giantswarm/app-operator/service/controller/appcatalog/v1/key"
+	"github.com/giantswarm/app-operator/service/controller/appcatalog/v1/resource/index"
+)
+
+// ResourceSetConfig contains necessary dependencies and settings for
+// AppConfig controller ResourceSet configuration.
```
rossf7:
```suggestion
// AppCatalog controller ResourceSet configuration.
```

service/controller/appcatalog/v1/resource_set.go
```diff
@@ -0,0 +1,137 @@
...
+	// Dependencies.
+	K8sClient kubernetes.Interface
+	Logger    micrologger.Logger
+
+	// Settings.
+	HandledVersionBundles []string
+	ProjectName           string
+}
+
+// NewResourceSet returns a configured App controller ResourceSet.
```
rossf7: Here too.

service/controller/appcatalog/v1/resource_set.go
```diff
@@ -0,0 +1,137 @@
...
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
+	}
+
+	// Settings.
+	if config.ProjectName == "" {
+		return nil, microerror.Maskf(invalidConfigError, "%T.ProjectName must not be empty", config)
+	}
+
+	var appCatalogResource controller.Resource
```
rossf7:
```suggestion
	var indexResource controller.Resource
```

service/service.go
```diff
@@ -91,9 +90,27 @@ func New(config Config) (*Service, error) {
...
+			WatchNamespace: config.Viper.GetString(config.Flag.Service.Kubernetes.Watch.Namespace),
+		}
+
+		appController, err = app.NewApp(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var appCatalogController *appcatalog.AppCatalog
```
rossf7: Lets make this controller first. Generally we always try and order code alphabetically so its cleaner easier to navigate. Also in this case we boot the appCatalog controller first.

service/version_bundles.go
```diff
@@ -3,7 +3,7 @@ package service
 import (
 	"github.com/giantswarm/versionbundle"
 
-	"github.com/giantswarm/app-operator/service/controller/v1"
+	"github.com/giantswarm/app-operator/service/controller/app/v1"
```
rossf7: Version bundles for both controllers need to be added here.

service/controller/appcatalog/v1/key/key.go
```diff
@@ -0,0 +1,37 @@
+package key
+
+import (
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/microerror"
+)
+
+const (
+	versionBundleAnnotation = "giantswarm.io/version-bundle"
```
xh3b4sd: In the light of https://github.com/giantswarm/giantswarm/issues/3703 and `epic/voo` which will make version bundles actually obsolete, do we still need this?

service/controller/appcatalog/v1/key/key.go
xh3b4sd: We wanted to start aligning more with upstream. This would then be `ToCustomResource` and `customResource` respectively.

service/controller/appcatalog/v1/resource/index/resource.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/giantswarm/operatorkit/controller"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// Name is the identifier of the resource.
+	Name = "indexv1"
```
xh3b4sd: What is an index resource?

service/controller/appcatalog/v1/key/key.go
tomahawk28: We are using this variable to compare between CRs version. 
Are you suggesting to use another way?

rossf7: VOO spec is to use a label `operator.giantswarm.io/version=v1.2.3`. Here we are using an annotation. I'd like us to align here and use a label too `application.giantswarm.io/version=v1.2.3`.

We went with an annotation to keep the CRs lean but I think a label is better. Its explicit and there are some cases where we would want to use it as a selector.

@puja108 @tomahawk28 WDYT?

Good to discuss but for the context of this PR lets please stick with the annotation.

service/controller/appcatalog/v1/key/key.go
rossf7: That's nice. Lets do that. I'll also change the release resource in chart-operator.

service/controller/appcatalog/v1/resource/index/resource.go
rossf7: Helm catalogs natively support an index.yaml. We want to use this in the front end as the master copy for metadata about App charts. 

e.g. https://github.com/giantswarm/sample-catalog/blob/master/index.yaml

So this resource will save the index to a configmap. This is to decouple the APIs from accessing GitHub directly. In the future its likely we will have either chartmuseum or harbor running in the control plane. But IMO its still better for the APIs to access via a configmap.

https://github.com/helm/chartmuseum
https://goharbor.io/

Note: Harbor still runs chartmuseum under the hood.

service/controller/app/v1/key/key.go
```diff
@@ -21,19 +21,19 @@ func ReleaseName(customObject v1alpha1.App) string {
 	return customObject.Spec.Release
 }
 
-// ToCustomObject converts value to v1alpha1.App and returns it or error
+// ToCustomResource converts value to v1alpha1.App and returns it or error
```
rossf7: <3

service/controller/appcatalog/appcatalog.go
```diff
@@ -0,0 +1,120 @@
...
+			ProjectName: config.ProjectName,
+		}
+
+		resourceSetV1, err = v1.NewResourceSet(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var appController *controller.Controller
```
rossf7:
```suggestion
	var appCatalogController *controller.Controller
```

service/controller/appcatalog/v1/key/key.go
kopiczko: There was an update to the VOO spec. We will use annotations like `app-operator. giantswarm.io/version=v1.2.3`. The same ones we have for `nodeconfigs`. There are two benefits: (1) we already have that defined and (2) multiple operators can reconcile the same CR.

kopiczko: I think `application.giantswarm.io/version=v1.2.3` makes sense. Let's just put that in `fmt`.

service/controller/appcatalog/appcatalog.go
```diff
@@ -0,0 +1,120 @@
...
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var newInformer *informer.Informer
+	{
+		c := informer.Config{
+			Logger:  config.Logger,
+			Watcher: config.G8sClient.ApplicationV1alpha1().Apps(config.WatchNamespace),
```
kopiczko: Does `appcatalog` controller watch the same objects as `app` controller? Does it make sense to have two controllers if so?

tomahawk28: You're right, I almost miss it. 😱

tomahawk28: It should look for `appCatalog`, I will make a change now.

tomahawk28:
I'm not sure how I didn't got the same error, but with below error, I can't run the operator. 
```
Jihyuks-MacBook-Pro-2:app-operator jihyukbok$
Jihyuks-MacBook-Pro-2:app-operator jihyukbok$ go run main.go daemon --service.kubernetes.incluster=false         --service.kubernetes.address="https://$(minikube ip):8443"         --service.kubernetes.tls.cafile="$HOME/.minikube/ca.crt"         --service.kubernetes.tls.crtfile="$HOME/.minikube/client.crt"         --service.kubernetes.tls.keyfile="$HOME/.minikube/client.key" | luigi
D 01/16 08:09:27 creating out-cluster config | operatorkit/client/k8srestconfig/k8s_rest_config.go:134
panic: [{/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/main.go:62: service.New} {/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/service/service.go:141: } {/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/microendpoint/service/version/service.go:62: } {/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/versionbundle/bundles.go:45: name must be the same for all version bundles} {invalid bundles}]


goroutine 1 [running]:
main.mainWithError.func1(0xc000148400, 0xc000148a00, 0xc0003cef00)
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/main.go:62 +0x4a7
github.com/giantswarm/app-operator/vendor/github.com/giantswarm/microkit/command/daemon.(*command).Execute(0xc0001792c0, 0xc00016c780, 0xc0000cee10, 0x0, 0x5)
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/microkit/command/daemon/command.go:104 +0x195
github.com/giantswarm/app-operator/vendor/github.com/giantswarm/microkit/command/daemon.(*command).Execute-fm(0xc00016c780, 0xc0000cee10, 0x0, 0x5)
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/microkit/command/daemon/command.go:56 +0x52
github.com/giantswarm/app-operator/vendor/github.com/spf13/cobra.(*Command).execute(0xc00016c780, 0xc0000cedc0, 0x5, 0x5, 0xc00016c780, 0xc0000cedc0)
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/spf13/cobra/command.go:766 +0x2cc
github.com/giantswarm/app-operator/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc00016cc80, 0x1fafaa6, 0x35, 0xc0003cec80)
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/spf13/cobra/command.go:852 +0x2fd
github.com/giantswarm/app-operator/vendor/github.com/spf13/cobra.(*Command).Execute(0xc00016cc80, 0xc00016cc80, 0x1e)
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/spf13/cobra/command.go:800 +0x2b
main.mainWithError(0x5000104, 0xc000161f88)
```

So I need to change the bundle version in either controller as different slightly.
But I think it's bit weird and don't think it's nice to have different version between controller. 
(it denied to have different bundle version name though)

rossf7:
@tomahawk28 Nice! 🎉 Thanks for getting this sorted. I know it got blocked on lots of issues.

service/controller/appcatalog/v1/key/key.go
kopiczko: So this should be actually be `app-operator.giantswarm.io/version` according to the latest spec.

kopiczko: If that is a subject to change then this is perfectly fine.

PR #13, opened 2019-01-18
Chart GetDesiredState part

tomahawk28:

### How to test

0. First, Run `chart-operator` to create `Chart` CRD
1. Create sample appCatalog by `catalog.yaml` 
```
kubectl create -f https://gist.githubusercontent.com/tomahawk28/af3a71fedff127e9e728ed8391cb0eaf/raw/10ec3c8bdeb8a4ee6f0961548f74df049ddaa14d/catalog.yaml
```
2. Create sample app by `app.yaml`
```
kubectl create -f https://gist.githubusercontent.com/tomahawk28/af3a71fedff127e9e728ed8391cb0eaf/raw/10ec3c8bdeb8a4ee6f0961548f74df049ddaa14d/app.yaml
```

3. Run this operator and you would find out it create `Chart` CR. 
```
Jihyuks-MacBook-Pro-2:app-operator jihyukbok$ kubectl get chart
NAME                    AGE
kubernetes-prometheus   37m
```


service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,66 @@
...
+	customResource, err := key.ToCustomResource(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	catalogName := key.CatalogName(customResource)
+
+	appCatalog, err := r.g8sClient.ApplicationV1alpha1().AppCatalogs("default").Get(catalogName, v1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return nil, microerror.Maskf(notFoundError, "appCatalog '%s' in namespace 'default' not found", catalogName)
```
kopiczko:
```suggestion
		return nil, microerror.Maskf(notFoundError, "appCatalog %#q in namespace %#q", catalogName, "default")
```

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/microerror"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
+	customResource, err := key.ToCustomResource(obj)
```
kopiczko: https://github.com/giantswarm/fmt/blob/master/go/variable_names.md#custom-resource

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,66 @@
...
+			Annotations: customResource.GetObjectMeta().GetAnnotations(),
+		},
+		Spec: v1alpha1.ChartSpec{
+			Name:       customResource.GetObjectMeta().GetName(),
+			Namespace:  customResource.Spec.Namespace,
+			TarballURL: chartURL,
+		},
+	}
+
+	if customResource.Spec.KubeConfig != (v1alpha1.AppSpecKubeConfig{}) {
```
kopiczko: Why is this `if` needed? Also applies to the one below.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,63 @@
...
+		},
+		Spec: v1alpha1.ChartSpec{
+			Name:       cr.GetObjectMeta().GetName(),
+			Namespace:  cr.Spec.Namespace,
+			TarballURL: chartURL,
+		},
+	}
+
+	chartCR.Spec.KubeConfig.Secret.Name = cr.Spec.KubeConfig.Secret.Name
+	chartCR.Spec.KubeConfig.Secret.Namespace = cr.Spec.KubeConfig.Secret.Namespace
```
rossf7: Lets remove the setting of the KubeConfig.

The app CRs have the kubeconfig set because they need to be created in the tenant cluster. The chart CRs are managed by chart-operator in the same cluster. So we don't need to set the kubeconfig.

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -0,0 +1,226 @@
...
+	k8sfake "k8s.io/client-go/kubernetes/fake"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func TestResource_GetDesiredState(t *testing.T) {
+
+	tests := []struct {
+		name          string
+		appObj        *v1alpha1.App
```
rossf7:
```suggestion
		obj        *v1alpha1.App
```

Convention is we use obj here because its the object received from the k8s informer.

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -0,0 +1,226 @@
...
+					},
+					Name: "my-cool-prometheus",
+					KubeConfig: v1alpha1.ChartSpecKubeConfig{
+						Secret: v1alpha1.ChartSpecKubeConfigSecret{
+							Name:      "giantswarm-12345",
+							Namespace: "12345",
+						},
+					},
+					Namespace:  "monitoring",
+					TarballURL: "https://giantswarm.github.com/app-catalog/-kubernetes-prometheus-1.0.0.tgz",
```
rossf7: Why does the path have  `-kubernetes-prometheus`?

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -0,0 +1,226 @@
...
+						"giantswarm.io/managed-by":     "app-operator",
+						"giantswarm.io/version-bundle": "0.1.0",
+					},
+				},
+				Spec: v1alpha1.AppSpec{
+					Catalog: "giantswarm",
+					Release: "1.0.0",
+					Config: v1alpha1.AppSpecConfig{
+						ConfigMap: v1alpha1.AppSpecConfigConfigMap{
+							Name:      "ginat-swarm-config",
```
rossf7:
```suggestion
							Name:      "giant-swarm-config",
```

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -0,0 +1,226 @@
...
+		}}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			objs := make([]runtime.Object, 0, 0)
+			if tc.appCatalog != nil {
+				objs = append(objs, tc.appCatalog)
+			}
+			c := Config{
+				K8sClient: k8sfake.NewSimpleClientset(),
+				G8sClient: fake.NewSimpleClientset(objs...),
```
rossf7: Should be sorted alphabetically.

service/controller/app/v1/resource/chart/resource.go
```diff
@@ -32,19 +39,32 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
 }
 
-func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
-	return nil, nil
-}
-
 func (r *Resource) NewUpdatePatch(ctx context.Context, obj, currentState, desiredState interface{}) (*controller.Patch, error) {
-	return nil, nil
+	patch := controller.NewPatch()
+	patch.SetCreateChange(desiredState)
```
rossf7: Why is this set?

In a later PR `SetCreateChange` can be implemented but this is incorrect as it doesn't consider the current state.

service/controller/app/v1/resource/chart/resource.go
```diff
@@ -32,19 +39,32 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+
+	return patch, nil
 }
 
 func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desiredState interface{}) (*controller.Patch, error) {
 	return nil, nil
 }
 
 func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
+	app, err := key.ToCustomResource(obj)
```
rossf7: Same here. This can be added later this PR should just implement the desired state.

service/controller/app/v1/resource/chart/desired.go
tomahawk28: @rossf7 If that is the case, then we should delete this `kubeConfig` part in `chart` CR specification. https://github.com/giantswarm/giantswarm/blob/b6af8e15df1600aada45567fbd37e053d1403e1f/roadmap/MANAGED-SERVICES-CATALOG.MD#release-process

service/controller/app/v1/resource/chart/desired_test.go
tomahawk28: Oops. I can change it to `https://giantswarm.github.com/app-catalog/kubernetes-prometheus-1.0.0.tgz`
So it should be : `{base_URL}/{app_name}-{release-version}.tgz`

rossf7: Yes.

service/controller/app/v1/resource/chart/resource.go
tomahawk28: I set this method just for checking the creation of `chart` CRD, I can comment it out now rather than delete it, What do you think?

rossf7: I'd prefer it was removed but in this case I guess its OK.

I'm not a fan of lots of commented out code. It's untidy and its easy to forget to update it later.

tomahawk28: In this case I need to fill out these functions eventually :)

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	catalogName := key.CatalogName(cr)
+
+	appCatalog, err := r.g8sClient.ApplicationV1alpha1().AppCatalogs("default").Get(catalogName, v1.GetOptions{})
```
rossf7: "default" shouldn't be hardcoded. We should use the flag for the watch namespace.

https://github.com/giantswarm/app-operator/blob/aafeed93353391ff15f25ce33886f11f7d594371/flag/service/kubernetes/watch/watch.go#L6

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,70 @@
...
+			Annotations: cr.GetObjectMeta().GetAnnotations(),
+		},
+		Spec: v1alpha1.ChartSpec{
+			Name:       cr.GetObjectMeta().GetName(),
+			Namespace:  cr.Spec.Namespace,
+			TarballURL: chartURL,
+		},
+	}
+
+	chartCR.Spec.Config.Secret.Name = cr.Spec.Config.Secret.Name
```
rossf7: Why not set these inline with the rest of the spec? Also we should have key functions for these.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,70 @@
...
+	chartCR.Spec.Config.Secret.Namespace = cr.Spec.Config.Secret.Namespace
+
+	chartCR.Spec.Config.ConfigMap.Name = cr.Spec.Config.ConfigMap.Name
+	chartCR.Spec.Config.ConfigMap.Namespace = cr.Spec.Config.ConfigMap.Namespace
+
+	return chartCR, nil
+}
+
+func generateCatalogURL(baseURL string, appName string, release string) (string, error) {
```
rossf7: What happens if the appName and release are empty? I think we should return a failedExecution error in that case.

service/controller/app/v1/resource/chart/error.go
```diff
@@ -19,3 +19,12 @@ var wrongTypeError = &microerror.Error{
 func IsWrongType(err error) bool {
 	return microerror.Cause(err) == wrongTypeError
 }
+
+var notFoundError = &microerror.Error{
+	Kind: "notFoundError",
+}
```
rossf7: Errors should be ordered alphabetically.

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -0,0 +1,220 @@
...
+					},
+					KubeConfig: v1alpha1.AppSpecKubeConfig{
+						Secret: v1alpha1.AppSpecKubeConfigSecret{
+							Name:      "giantswarm-12345",
+							Namespace: "12345",
+						},
+					},
+					Name:      "kubernetes-prometheus",
+					Namespace: "monitoring",
+				},
```
rossf7: I'd like to see more error cases for when the tarballURL cannot be generated.

service/controller/app/v1/resource/chart/resource.go
```diff
@@ -11,19 +11,24 @@ import (
 
 const (
 	// Name is the identifier of the resource.
-	Name = "chartv1"
+	chartAPIVersion           = "application.giantswarm.io"
```
rossf7: Godoc is wrong when this is added. Lets have `Name` first and a newline to separate the un-exported constants.

service/controller/app/v1/resource/chart/resource.go
rossf7: >  but in this case I guess its OK

Yep :) but still not a fan of this approach.

service/controller/app/v1/resource/chart/desired_test.go
tomahawk28: Should be one enough?

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,77 @@
...
+	catalogName := key.CatalogName(cr)
+
+	appCatalog, err := r.g8sClient.ApplicationV1alpha1().AppCatalogs(r.watchNamespace).Get(catalogName, v1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return nil, microerror.Maskf(notFoundError, "appCatalog %#q in namespace %#q", catalogName, "default")
+	} else if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	chartURL, err := generateCatalogURL(appCatalog.Spec.CatalogStorage.URL, cr.Spec.Name, cr.Spec.Release)
```
rossf7:
```suggestion
	tarballURL, err := generateTarballURL(appCatalog.Spec.CatalogStorage.URL, cr.Spec.Name, cr.Spec.Release)
```

I missed it earlier but I think it makes sense to rename this.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,77 @@
...
+			TarballURL: chartURL,
+		},
+	}
+
+	return chartCR, nil
+}
+
+func generateCatalogURL(baseURL string, appName string, release string) (string, error) {
+	if baseURL == "" || appName == "" || release == "" {
+		return "", microerror.Maskf(failedExecution, "baseURL(%s), appName(%s), release(%s) should not left as blank string", baseURL, appName, release)
```
rossf7:
```suggestion
		return "", microerror.Maskf(failedExecution, "baseURL %#q, appName %#q, release %#q should not be empty", baseURL, appName, release)
```

Our convention is to use %#q which means the strings will be quoted in log messages. So they are easier to read.

service/controller/app/v1/resource/chart/desired_test.go
rossf7: I think that's fine.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,77 @@
...
+	catalogName := key.CatalogName(cr)
+
+	appCatalog, err := r.g8sClient.ApplicationV1alpha1().AppCatalogs(r.watchNamespace).Get(catalogName, v1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return nil, microerror.Maskf(notFoundError, "appCatalog %#q in namespace %#q", catalogName, "default")
+	} else if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	tarballURL, err := generateTarballURL(appCatalog.Spec.CatalogStorage.URL, cr.Spec.Name, cr.Spec.Release)
```
rossf7: Lets have more key functions for these. Reading values from `.Spec` is a sign we're missing these.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	chartCR := &v1alpha1.Chart{
+		TypeMeta: v1.TypeMeta{
+			Kind:       chartKind,
+			APIVersion: chartAPIVersion,
+		},
+		ObjectMeta: v1.ObjectMeta{
+			Name:        cr.Spec.Name,
+			Labels:      cr.GetObjectMeta().GetLabels(),
```
rossf7:
```suggestion
			Labels:      cr.ObjectMeta.Labels,
```

This will also work and is simpler.

service/controller/appcatalog/v1/key/key.go
```diff
@@ -13,6 +13,10 @@ func AppCatalogTitle(customObject v1alpha1.AppCatalog) string {
 	return customObject.Spec.Title
 }
 
+func CatalogStorageURL(customObject v1alpha1.AppCatalog) string {
```
rossf7: Same here we should use `customResource` here to align with the new naming in `fmt`.

service/controller/app/v1/key/key.go
```diff
@@ -13,6 +13,18 @@ func AppName(customObject v1alpha1.App) string {
 	return customObject.Spec.Name
 }
 
+func CatalogName(customObject v1alpha1.App) string {
```
```suggestion
func CatalogName(customResource v1alpha1.App) string {
```

We should use the new naming here.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -0,0 +1,78 @@
+package chart
+
+import (
+	"context"
+	"fmt"
+	appcatalogkey "github.com/giantswarm/app-operator/service/controller/appcatalog/v1/key"
```
rossf7: This should be in the final block.

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -0,0 +1,280 @@
+package chart
+
+import (
+	"context"
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned/fake"
```
rossf7: Should be in 2nd block.

service/controller/app/v1/resource/chart/resource.go
```diff
@@ -2,7 +2,7 @@ package chart
 
 import (
 	"context"
-
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
```
rossf7: Imports need fixing here too.

PR #14, opened 2019-01-21
Add kubeconfig service for connecting to the cluster configured in the app CR

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/4845

app-operator needs to support 2 modes for App Catalog and VOO. For App Catalog we need to create chart CRs, configmaps and secrets in the tenant cluster. We use the new kubeconfig library for this.

https://github.com/giantswarm/kubeconfig

For VOO we just need to create these resources in the control plane cluster that app-operator is running in. This PR adds a kubeconfig service which wraps this functionality. As we'll need it in the chart, configmap and secret resources.


service/controller/app/v1/key/key_test.go
```diff
@@ -5,7 +5,7 @@ import (
 	"testing"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
rossf7: Fixing the import name. 

@tomahawk28 FYI. We use this naming convention to avoid conflicts when we need multiple k8s packages.

service/controller/app/v1/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,107 @@
...
+
+// Config represents the configuration used to create a new kubeconfig service.
+type Config struct {
+	G8sClient versioned.Interface
+	K8sClient kubernetes.Interface
+	Logger    micrologger.Logger
+}
+
+// KubeConfig service provides primitives for connecting to the Kubernetes
+// cluster configured in the kubeconfig section of the app CR.
```
rossf7: Logic is in a service because we'll need it in multiple resources.

service/controller/app/v1/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,107 @@
...
+	return k, nil
+}
+
+// NewG8sClientForApp returns a generated clientset for the cluster configured
+// in the kubeconfig section of the app CR. If this is empty a clientset for
+// the current cluster is returned.
+func (k KubeConfig) NewG8sClientForApp(ctx context.Context, customResource v1alpha1.App) (versioned.Interface, error) {
+	secretName := key.KubeConfigSecretName(customResource)
+
+	// KubeConfig is not configured so connect to current cluster.
```
rossf7: This conditional allows us to support both modes.

service/controller/app/v1/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,107 @@
...
+	if secretName == "" {
+		return k.g8sClient, nil
+	}
+
+	secretNamespace := key.KubeConfigSecretNamespace(customResource)
+	k8sClient, err := k.kubeConfig.NewG8sClientFromSecret(ctx, secretName, secretNamespace)
```
rossf7: We use the library to handle getting the secret and parsing the kubeconfig with the upstream client-go method.

service/controller/app/v1/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	return k8sClient, nil
+}
+
+// NewK8sClientForApp returns a Kubernetes clientset for the cluster configured
+// in the kubeconfig section of the app CR. If this is empty a clientset for
+// the current cluster is returned.
```
rossf7: Same thing for the k8s client.

service/controller/app/v1/resource_set.go
```diff
@@ -52,11 +53,26 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		return nil, microerror.Maskf(invalidConfigError, "%T.WatchNamespace must not be empty", config)
 	}
 
-	var appResource controller.Resource
+	var kubeConfigService *kubeconfig.KubeConfig
```
rossf7: Here is the wiring.

service/controller/app/v1/resource_set.go
```diff
@@ -52,11 +53,26 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
+			Logger:    config.Logger,
+		}
+
+		kubeConfigService, err = kubeconfig.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var chartResource controller.Resource
```
rossf7: Fixing the resource name.

service/controller/app/v1/resource_set.go
```diff
@@ -88,10 +104,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 	}
 
 	{
-		c := metricsresource.WrapConfig{
-			Name: config.ProjectName,
-		}
-
+		c := metricsresource.WrapConfig{}
```
rossf7: Needed for latest operatorkit.

service/controller/app/v1/key/key_test.go
```diff
@@ -93,6 +93,44 @@ func Test_Namespace(t *testing.T) {
...
+				Secret: v1alpha1.AppSpecKubeConfigSecret{
+					Name:      "cluster-12345-kubeconfig",
+					Namespace: "default",
+				},
+			},
+		},
+	}
+
+	if KubeConfigSecretNamespace(obj) != expectedNamespace {
+		t.Fatalf("app namespace %s, want %s", KubeConfigSecretNamespace(obj), expectedNamespace)
```
tomahawk28:
```suggestion
		t.Fatalf("app kubeconfig secret namespace %s, want %s", KubeConfigSecretNamespace(obj), expectedNamespace)
```

service/controller/app/v1/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+// Config represents the configuration used to create a new kubeconfig service.
+type Config struct {
+	G8sClient versioned.Interface
```
tomahawk28: I can assume that these `G8sClient` and `K8sClient` should the same as what `chart` resource contains itself, doesn't it? 
https://github.com/giantswarm/app-operator/blob/a9ce6eb4c31200c73be1cdec814559fec0e403ec/service/controller/app/v1/resource/chart/resource.go#L37-L38
Rather than copy clients to another instance, what about returning nil but error when Secret is not placed in app CR?

rossf7:
@xh3b4sd The approach with the kubeconfig library is more generic. app-operator can manage resources for any cluster it has a kubeconfig for. It doesn't need to be a tenant cluster.

When we migrate the current managed apps kube-state-metrics etc from chartconfig CRs to app CRs we'll stop using tenantcluster in cluster-operator. As the app CRs will be in the control plane cluster.

Maybe we can remove the other uses of tenantcluster. But that might need CRD changes to add the kubeconfig fields. Not sure yet.

service/controller/app/v1/key/key_test.go
rossf7: Thanks good catch. Fixed.

service/controller/app/v1/kubeconfig/kubeconfig.go
rossf7: The problem is this logic will be needed in 3 resources (chart, configmap and secret). Rather than repeat it in all 3 resources the kubeconfig service handles making sure the correct clientset is used.

PR #15, opened 2019-01-22
Add GetCurrentState for app CR

tomahawk28:

Created `GetCurrentState` method for `app` controller.


service/controller/app/v1/resource/chart/create.go
```diff
@@ -0,0 +1,28 @@
+package chart
+
+import (
+	"context"
+	"fmt"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+	"github.com/giantswarm/microerror"
+	"k8s.io/apimachinery/pkg/api/errors"
```
rossf7:
```suggestion
	apierrors "k8s.io/apimachinery/pkg/api/errors"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

These are the usual aliases we use. Also the imports need sorting.

service/controller/app/v1/resource/chart/create.go
```diff
@@ -0,0 +1,28 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
+	customResource, err := key.ToCustomResource(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	name := key.AppName(customResource)
+	chart, err := r.g8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, v1.GetOptions{})
```
rossf7: We can't always use the g8sClient. For App Catalog we need to connect to the tenant cluster. Its merged now so you can use the kubeconfig service to get the right client.

https://github.com/giantswarm/app-operator/blob/873ea9601aa18a42bbe35947cf4e3a953af95b6a/service/controller/app/v1/kubeconfig/kubeconfig.go#L72

service/controller/app/v1/resource/chart/create.go
```diff
@@ -0,0 +1,28 @@
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("chart %#q is not created yet", name))
```
rossf7:
```suggestion
		// Return early as chart is not installed.
```

I don't think we need to log here. It can just be a comment.

service/controller/app/v1/resource/chart/create_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned/fake"
+	"github.com/giantswarm/micrologger/microloggertest"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
```
rossf7:
```suggestion
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

service/controller/app/v1/resource/chart/create_test.go
```diff
@@ -0,0 +1,263 @@
+package chart
```
rossf7: Filename should be current_test.go.

service/controller/app/v1/resource/chart/create_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func TestResource_GetCurrentState(t *testing.T) {
+
+	tests := []struct {
+		name          string
+		obj           *v1alpha1.App
+		returnChart   *v1alpha1.Chart
```
rossf7:
```suggestion
		returnedChart   *v1alpha1.Chart
```

service/controller/app/v1/resource/chart/create_test.go
```diff
@@ -0,0 +1,263 @@
...
+		expectedChart *v1alpha1.Chart
+		errorMatcher  func(error) bool
+	}{
+		{
+			name: "case 0: chart already created",
+			obj: &v1alpha1.App{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      "my-cool-prometheus",
+					Namespace: "default",
+					Labels: map[string]string{
```
rossf7: IMO you could remove the labels and annotations here. As they aren't needed for the test. Just to make it easier to read.

service/controller/app/v1/resource/chart/create.go
```diff
@@ -0,0 +1,28 @@
...
+
+func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
+	customResource, err := key.ToCustomResource(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	name := key.AppName(customResource)
+	chart, err := r.g8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, v1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
```
xh3b4sd: Can you check how we prefer to handle errors here: https://github.com/giantswarm/fmt/blob/master/go/errors.md#matching-errors.

service/controller/app/v1/resource/chart/create_test.go
```diff
@@ -0,0 +1,263 @@
...
+						},
+					},
+					Name:       "my-cool-prometheus",
+					Namespace:  "monitoring",
+					TarballURL: "https://giantswarm.github.com/app-catalog/kubernetes-prometheus-1.0.0.tgz",
+				},
+			},
+		},
+		{
+			name: "case 0: chart not found",
```
rossf7:
```suggestion
			name: "case 1: chart not found",
```

service/controller/app/v1/resource/chart/create.go
xh3b4sd: Please have a look at how we prefer to log information: https://github.com/giantswarm/fmt/blob/master/go/log_messages.md.

tomahawk28:
After using KubeConfig library, it failed in unittest since I didn't mock that library part. 
Moreover `KubeConfig` service generate library instance inside of `New` generator and keep it their own client instances, it need some more efforts to mock it entirely but I'm not sure the whole g8s would like the idea.

kopiczko:
I think testing getting object from k8s api with mocks is a waste of time. I would invest that time in e2e testing. #opinion

service/controller/app/v1/resource/chart/create.go
tomahawk28: If you are talking about returning variables, I deliberately return it as nil since no instance I can find.

service/controller/app/v1/resource/chart/create.go
tomahawk28: this log will be deleted.

service/controller/app/v1/resource/chart/create.go
xh3b4sd: I mean the conditional indentation. We try to keep it straight forward and test one matcher after another in case there are multiple ones. Let's hang if this is not clear.

service/controller/app/v1/kubeconfig/kubeconfig.go
```diff
@@ -23,10 +23,10 @@ type Config struct {
...
 // cluster configured in the kubeconfig section of the app CR.
 type KubeConfig struct {
-	g8sClient  versioned.Interface
-	k8sClient  kubernetes.Interface
-	kubeConfig kubeconfiglib.Interface
-	logger     micrologger.Logger
+	G8sClient  versioned.Interface
+	K8sClient  kubernetes.Interface
+	KubeConfig kubeconfiglib.Interface
+	Logger     micrologger.Logger
```
kopiczko: Why did this become public?

service/controller/app/v1/resource/chart/create.go
xh3b4sd: I think the logging makes sense, just the wording is off and it should align with the rest. So far we have logging before and after an action to have a better understanding of the internals at runtime. We can hang if this is unclear.

service/controller/app/v1/kubeconfig/kubeconfig.go
tomahawk28: Otherwise, I don;t know how to set this up inside unittest, 
https://github.com/giantswarm/app-operator/blob/ec8c0f1b8fb6237048f7cdeb83c1325ed6dbba7f/service/controller/app/v1/resource/chart/current_test.go#L192-L196
any idea?

service/controller/app/v1/resource/chart/create.go
tomahawk28: okay, then it's contradicted with @rossf7 but I also think logging in here make sense.

rossf7:
@tomahawk28 For the error you just need to configure the kubeConfig library.

e.g. https://github.com/giantswarm/app-operator/blob/873ea9601aa18a42bbe35947cf4e3a953af95b6a/service/controller/app/v1/resource/chart/desired_test.go#L246-L258

In a call will address other comments later.

tomahawk28:
@rossf7 Unfortunately `desired_test.go` do not using kubeconfig library, so copying its doesn't help in this test situation.

service/controller/app/v1/kubeconfig/kubeconfig.go
tomahawk28: Instead of making all those public, I have changed to use it `TestKubeconfig` for the testing method, is it better?

service/controller/app/v1/resource/chart/current.go
```diff
@@ -0,0 +1,34 @@
...
+	customResource, err := key.ToCustomResource(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	name := key.AppName(customResource)
+	client, err := r.kubeConfig.NewG8sClientForApp(ctx, customResource)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	chart, err := client.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, metav1.GetOptions{})
```
rossf7: Whitespace is a bit strange here.

service/controller/app/v1/resource/chart/current_test.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/giantswarm/app-operator/service/controller/app/v1/kubeconfig"
+)
+
+func TestResource_GetCurrentState(t *testing.T) {
+
+	tests := []struct {
+		name          string
+		obj           *v1alpha1.App
+		returnedChart *v1alpha1.Chart
+		expectedChart *v1alpha1.Chart
```
rossf7: Why is there returnedChart and expectedChart? Aren't they the same?

service/controller/app/v1/resource/chart/fakekubeconfig.go
```diff
@@ -0,0 +1,20 @@
+package chart
```
rossf7: Can this file be removed?

service/controller/app/v1/resource/chart/resource.go
```diff
@@ -109,6 +104,7 @@ func New(config Config) (*Resource, error) {
 		k8sClient:      config.K8sClient,
 		logger:         config.Logger,
 		watchNamespace: config.WatchNamespace,
+		kubeConfig:     config.KubeConfig,
```
rossf7: Order should be alphabetical.

tomahawk28:
All comment had been addressed, PTAL.

service/controller/app/v1/resource/chart/current.go
```diff
@@ -0,0 +1,33 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
+	customResource, err := key.ToCustomResource(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
```
rossf7: Lets have a newline here.

service/controller/app/v1/resource/chart/current.go
```diff
@@ -0,0 +1,33 @@
...
+	name := key.AppName(customResource)
+	client, err := r.kubeConfig.NewG8sClientForApp(ctx, customResource)
```
rossf7:
```suggestion
	g8sClient, err := r.kubeConfig.NewG8sClientForApp(ctx, customResource)
```

service/controller/app/v1/resource/chart/current.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
```
rossf7: Another newline here pls.

service/controller/app/v1/resource/chart/current.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/giantswarm/microerror"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
+	customResource, err := key.ToCustomResource(obj)
```
kopiczko:
```suggestion
	cr, err := key.ToCustomResource(obj)
```

service/controller/app/v1/resource/chart/current.go
```diff
@@ -0,0 +1,35 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	name := key.AppName(customResource)
```
kopiczko:
```suggestion
	name := key.AppName(cr)
```

service/controller/app/v1/resource/chart/current.go
```diff
@@ -0,0 +1,35 @@
...
+	g8sClient, err := r.kubeConfig.NewG8sClientForApp(ctx, customResource)
```
kopiczko:
```suggestion
	g8sClient, err := r.kubeConfig.NewG8sClientForApp(ctx, cr)
```

service/controller/app/v1/resource/chart/current.go
```diff
@@ -0,0 +1,35 @@
...
+
+	name := key.AppName(cr)
+	g8sClient, err := r.kubeConfig.NewG8sClientForApp(ctx, cr)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	chart, err := g8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart %#q", name))
```
xh3b4sd: The logging here is cool. Can we also have some for before and after the action? The visibility on what is going on is pretty useful, especially when developing new resources. Check our `fmt` docs or other resource implementations.

tomahawk28: it make sense, I will create other msgs before/after action as well.

tomahawk28: @xh3b4sd I need your approval, PTAL :)

PR #18, opened 2019-01-23
 Minor changes to chart resource

rossf7:
Some minor things I noticed during the previous PR. Keeping them separate to have cleaner diffs like we discussed in sig-operator sync.

### DIFF
--- a/service/controller/app/v1/resource/chart/desired_test.go
+++ b/service/controller/app/v1/resource/chart/desired_test.go
@@ -32,7 +32,7 @@ func TestResource_GetDesiredState(t *testing.T) {
 					Name:      "my-cool-prometheus",
 					Namespace: "default",
 					Labels: map[string]string{
-						"app": "prometheus",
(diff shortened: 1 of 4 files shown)


service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -43,8 +43,10 @@ func TestResource_GetDesiredState(t *testing.T) {
 					},
 				},
 				Spec: v1alpha1.AppSpec{
-					Catalog: "giantswarm",
-					Version: "1.0.0",
+					Catalog:   "giantswarm",
+					Name:      "kubernetes-prometheus",
+					Namespace: "monitoring",
+					Version:   "1.0.0",
```
rossf7: I find it easier to read when the top level fields are first.

rossf7:
Sigh `gofmt` integration in `VS Code` seems to be struggling at the moment. Should be good now.

PR #19, opened 2019-01-24
Add processLabels to desired state

rossf7:

Introduces a `processLabels` function that makes sure chart CRs have the correct labels. Any additional labels are passed through.


service/controller/app/v1/resource/chart/desired.go
```diff
@@ -76,3 +77,18 @@ func generateTarballURL(baseURL string, appName string, version string) (string,
 	u.Path = path.Join(u.Path, fmt.Sprintf("%s-%s.tgz", appName, version))
 	return u.String(), nil
 }
+
+func processLabels(projectName string, inputLabels map[string]string) map[string]string {
```
tomahawk28: Even if it's a private function, I would like to know that what this function does, why it was needed the first place. Probably more godoc? `processLabels` are quite unclear to me.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -78,8 +78,12 @@ func generateTarballURL(baseURL string, appName string, version string) (string,
 	return u.String(), nil
 }
 
-// processLabels ensures chart resources have the labels required by
-// chart-operatorbut and any additional labels remain.
+// processLabels ensures the chart-operator.giantswarm.io/version label is
```
tomahawk28: <3

service/controller/app/v1/resource/chart/desired.go
rossf7: There was already godoc but good point that it needed more detail. Is it clearer now?

PR #21, opened 2019-01-24
add ApplyCreateChange 

tomahawk28:
- Add `ApplyCreateChange` function to `app` resource
- Add `addCreateChange` prv function

### DIFF
--- a/service/controller/app/v1/resource/chart/create.go
+++ b/service/controller/app/v1/resource/chart/create.go
@@ -20,10 +20,12 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		return microerror.Mask(err)
 	}
 
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating chart %#q", chart.Name))
 	_, err = r.g8sClient.ApplicationV1alpha1().Charts(app.Namespace).Create(&chart)
 	if err != nil {
 		return microerror.Mask(err)
 	}
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("created %#q", chart.Name))
 	return nil
 }
 
--- a/service/controller/app/v1/resource/chart/create.go
+++ b/service/controller/app/v1/resource/chart/create.go
@@ -21,7 +21,6 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 	}
 
 	_, err = r.g8sClient.ApplicationV1alpha1().Charts(app.Namespace).Create(&chart)
-
 	if err != nil {
 		return microerror.Mask(err)
 	}
--- a/service/controller/app/v1/resource/chart/create.go
+++ b/service/controller/app/v1/resource/chart/create.go
@@ -11,7 +11,7 @@ import (
 )
 
 func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
-	app, err := key.ToCustomResource(obj)
+	cr, err := key.ToCustomResource(obj)
 	if err != nil {
 		return microerror.Mask(err)
 	}
@@ -20,12 +20,12 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		return microerror.Mask(err)
 	}
 
-	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating chart %#q", chart.Name))
-	_, err = r.g8sClient.ApplicationV1alpha1().Charts(app.Namespace).Create(&chart)
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensuring creation of chart %#q", chart.Name))
+	_, err = r.g8sClient.ApplicationV1alpha1().Charts(cr.Namespace).Create(&chart)
 	if err != nil {
 		return microerror.Mask(err)
 	}
-	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("created %#q", chart.Name))
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensured creation of chart %#q", chart.Name))
 	return nil
 }
 
@@ -43,7 +43,7 @@ func (r *Resource) newCreateChange(ctx context.Context, currentResource, desired
 
 	createChart := &v1alpha1.Chart{}
 
-	if isEmpty(currentChart) || currentChart.Name != desiredChart.Name {
+	if isEmpty(currentChart) {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the %#q chart needs to be created", desiredChart.Name))
 		createChart = &desiredChart
 	} else {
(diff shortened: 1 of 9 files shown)


service/controller/app/v1/resource/chart/create.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	chart, err := key.ToChart(createChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	_, err = r.g8sClient.ApplicationV1alpha1().Charts(app.Namespace).Create(&chart)
+
```
xh3b4sd:
```suggestion
```

service/controller/app/v1/resource/chart/create.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
```
xh3b4sd: Would be nice to get some logging in here.

service/controller/app/v1/resource/chart/create_test.go
```diff
@@ -0,0 +1,198 @@
...
+		{
+			name:            "case 0: new chart should be created",
+			currentResource: &v1alpha1.Chart{},
+			desiredResource: &v1alpha1.Chart{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "Chart",
+					APIVersion: "application.giantswarm.io",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "kubernetes-prometheus",
```
rossf7:
```suggestion
					Name: "prometheus",
```

We're dropping this prefix as its not really useful.

service/controller/app/v1/resource/chart/create_test.go
```diff
@@ -0,0 +1,198 @@
...
+					Labels: map[string]string{
+						"app": "prometheus",
+					},
+				},
+				Spec: v1alpha1.ChartSpec{
+					Config: v1alpha1.ChartSpecConfig{
```
rossf7: Can you remove the Config section here? It's not needed for the test and its easier to read the code if its shorter.

service/controller/app/v1/resource/chart/create_test.go
```diff
@@ -0,0 +1,198 @@
...
+						},
+					},
+					Name:       "my-cool-prometheus",
+					Namespace:  "monitoring",
+					TarballURL: "https://giantswarm.github.com/app-catalog/kubernetes-prometheus-1.0.0.tgz",
+				},
+			},
+		},
+		{
+			name: "case 1: chart already existed",
```
rossf7:
```suggestion
			name: "case 1: chart already exists",
```

Just a 🇬🇧 nitpick.

service/controller/app/v1/resource/chart/create_test.go
```diff
@@ -0,0 +1,198 @@
...
+				},
+			},
+			expectedChart: &v1alpha1.Chart{},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			var err error
+			var kc *kubeconfig.KubeConfig
```
rossf7: We should add a `kubeconfigtest` package like we have for `helmclient`.

https://github.com/giantswarm/helmclient/blob/c93791a1e9e60a08f2ee48394d416a726c5e00f2/helmclienttest/helmclient.go#L31

That way we only need `kubeconfigtest.New()` and it keeps the tests shorter.

service/controller/app/v1/resource/chart/resource.go
```diff
@@ -113,3 +89,19 @@ func New(config Config) (*Resource, error) {
...
 }
+
+// equals asseses the equality of ReleaseStates with regards to distinguishing fields.
+func equals(a, b v1alpha1.Chart) bool {
+	if a.Name != b.Name {
+		return false
+	}
+	if !reflect.DeepEqual(a.Spec, b.Spec) {
+		return false
+	}
```
rossf7: This is good but we should also compare ObjectMeta. As we may add or change the metadata in the future.

tomahawk28: But `ObjectMeta` contains several TimeStamp object it could be different in many cases. Can we select some few variables in ObjectMeta, just to compare?

rossf7: Good point. What about if we just consider labels and annotations?

service/controller/app/v1/resource/chart/create_test.go
tomahawk28: I added the new file https://github.com/giantswarm/app-operator/blob/e5cd4788e943a1b24b03705972b439fcc2213d6b/kubeconfigtest/kubeconfig.go

service/controller/app/v1/resource/chart/resource.go
tomahawk28: Sure, I would make that change right away.

service/controller/app/v1/resource/chart/create_test.go
rossf7: @tomahawk28 Thanks for adding the `kubeconfigtest` package but sorry my comment could have been clearer.

This should be created in the kubeconfig repo and then we just import it. That way we can use it wherever we use the library. https://github.com/giantswarm/kubeconfig

It's a cleanup task so if you want to do that in a separate PR that's fine with me.

tomahawk28: Oops, I got it, You're right we should do that on next PR.

service/controller/app/v1/resource/chart/create.go
```diff
@@ -0,0 +1,54 @@
...
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	chart, err := key.ToChart(createChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensuring creation of chart %#q", chart.Name))
```
rossf7: Sorry nitpicking on the whitespace but can you add a newline before and after the Create API call.

service/controller/app/v1/resource/chart/create_test.go
```diff
@@ -0,0 +1,143 @@
...
+				KubeConfig:     kc,
+				Logger:         microloggertest.New(),
+				WatchNamespace: "default",
+			}
+			r, err := New(c)
+			if err != nil {
+				t.Fatalf("error == %#v, want nil", err)
+			}
+
+			got, err := r.newCreateChange(context.TODO(), tt.currentResource, tt.desiredResource)
```
kopiczko:
```suggestion
			got, err := r.newCreateChange(context.Background(), tt.currentResource, tt.desiredResource)
```

PR #22, opened 2019-01-25
Add update function for app CR

tomahawk28:
- Add `ApplyUpdateChange` for app CR


service/controller/app/v1/resource/chart/create_test.go
```diff
@@ -6,26 +6,15 @@ import (
...
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
-	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
 	"github.com/giantswarm/apiextensions/pkg/clientset/versioned/fake"
-	"github.com/giantswarm/micrologger"
 	"github.com/giantswarm/micrologger/microloggertest"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
 
 	"github.com/giantswarm/app-operator/kubeconfigtest"
```
rossf7: As discussed this package should be `github.com/giantswarm/kubeconfig/kubeconfigtest`. Let's get this fixed having it in the wrong repo is messy.

service/controller/app/v1/resource/chart/update.go
```diff
@@ -2,21 +2,81 @@ package chart
...
 func (r *Resource) NewUpdatePatch(ctx context.Context, obj, currentChart, desiredChart interface{}) (*controller.Patch, error) {
 	create, err := r.newCreateChange(ctx, currentChart, desiredChart)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
+
+	update, err := r.newUpdateChange(ctx, currentChart, desiredChart)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
```
rossf7: Newline please.

service/controller/app/v1/resource/chart/update.go
```diff
@@ -2,21 +2,81 @@ package chart
...
 	patch := controller.NewPatch()
 	patch.SetCreateChange(create)
+	patch.SetUpdateChange(update)
```
rossf7: Also here.

service/controller/app/v1/resource/chart/update.go
```diff
@@ -2,21 +2,81 @@ package chart
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	chart, err := key.ToChart(updateChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if chart.Name != "" {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensuring updating of chart %#q", chart.Name))
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensuring update of chart %#q", chart.Name))
```

service/controller/app/v1/resource/chart/update.go
```diff
@@ -2,21 +2,81 @@ package chart
...
+
+		_, err = r.g8sClient.ApplicationV1alpha1().Charts(cr.Namespace).Update(&chart)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensured updating of chart %#q", chart.Name))
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensured update of chart %#q", chart.Name))
```

service/controller/app/v1/resource/chart/update.go
```diff
@@ -2,21 +2,81 @@ package chart
...
+	} else {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("no need to update charts"))
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("no need to update chart"))
```

service/controller/app/v1/resource/chart/update.go
```diff
@@ -2,21 +2,81 @@ package chart
...
+	}
+
+	desiredChart, err := key.ToChart(desiredResource)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding out if the chart has to be updated")
+
+	updateChart := &v1alpha1.Chart{}
```
rossf7: Why is `updateChart` needed?

I think setting the resource version of the desiredChart with the version from the current chart should be enough.

service/controller/app/v1/resource/chart/update_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned/fake"
+	"github.com/giantswarm/micrologger/microloggertest"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sfake "k8s.io/client-go/kubernetes/fake"
+
+	"github.com/giantswarm/app-operator/kubeconfigtest"
+)
+
+func TestResource_newUpdateChange(t *testing.T) {
```
rossf7:
```suggestion
func Test_Resource_newUpdateChange(t *testing.T) {
```

service/controller/app/v1/resource/chart/update_test.go
```diff
@@ -0,0 +1,148 @@
...
+			var err error
+
+			kc, err := kubeconfigtest.New(nil)
+			if err != nil {
+				t.Fatalf("error == %#v, want nil", err)
+			}
+			c := Config{
+				G8sClient:      fake.NewSimpleClientset(),
+				K8sClient:      k8sfake.NewSimpleClientset(),
+				KubeConfig:     kc,
```
rossf7:
```suggestion
				KubeConfig:     kubeconfigtest.New(kubeconfigtest.Config{}),
```

As per previous comment this needs to move to the `kubeconfig` repo. You can use this signature which keeps the tests simpler.

See https://github.com/giantswarm/helmclient/blob/c93791a1e9e60a08f2ee48394d416a726c5e00f2/helmclienttest/helmclient.go for an example.

service/controller/app/v1/resource/chart/update_test.go
```diff
@@ -0,0 +1,148 @@
...
+				WatchNamespace: "default",
+			}
+			r, err := New(c)
+			if err != nil {
+				t.Fatalf("error == %#v, want nil", err)
+			}
+
+			got, err := r.newUpdateChange(context.Background(), tt.currentResource, tt.desiredResource)
+			if err != nil {
+				t.Errorf("Resource.newCreateChange() error = %v", err)
```
rossf7: Why is this `Errorf`? Also the text is wrong.

I think you can just use `t.Fatalf("error == %#v, want nil", err)` here.

service/controller/app/v1/resource/chart/update_test.go
```diff
@@ -0,0 +1,148 @@
...
+				return
+			}
+			if !reflect.DeepEqual(got, tt.expectedChart) {
+				t.Errorf("Resource.newCreateChange() = %v, want %v", got, tt.expectedChart)
```
rossf7: Same here.

service/controller/app/v1/resource/chart/create_test.go
tomahawk28: I'm not sure about moving it to `kubeconfig` since then it will use app-operator repo as well, 
https://github.com/giantswarm/app-operator/blob/93d450bfd4aaccfc2df80d06b9c52ec76b4c8824/kubeconfigtest/kubeconfig.go#L9

should we move this `kubeconfig.Config` struct too?

rossf7: Ack. This is a problem with the kubeconfig service I introduced :(

The problem is this file is in the wrong place and needs to be removed.

https://github.com/giantswarm/app-operator/blob/master/kubeconfigtest/kubeconfig.go

Let's just use the current kubeconfig service with fake clientsets. 

```
			var kc kubeconfig.KubeConfig
			{
				c := kubeconfig.Config{
					G8sClient: fake.NewSimpleClientset(),
					K8sClient: k8sfake.NewSimpleClientset(),
					Logger:    microloggertest.New(),
				}

				kc, err := kubeconfig.New(nil)
				if err != nil {
					t.Fatalf("error == %#v, want nil", err)
				}
			}

			c := Config{
				G8sClient:      fake.NewSimpleClientset(),
				K8sClient:      k8sfake.NewSimpleClientset(),
				KubeConfig:     kc,
				Logger:         microloggertest.New(),
				WatchNamespace: "default",
			}
			r, err := New(c)
			if err != nil {
				t.Fatalf("error == %#v, want nil", err)
			}
```

It makes the test longer but its the best option for now. Next week I'll take another look at the kubeconfig service and see if this can be improved.

service/controller/app/v1/resource/chart/update_test.go
tomahawk28: We decided that we revert this change to use it previous method.

service/controller/app/v1/resource/chart/update.go
rossf7: OK that makes sense. Thx.

PR #23, opened 2019-01-28
Add delete function

tomahawk28:
- Add deletion function for app resource.

### DIFF
--- a/service/controller/app/v1/resource/chart/delete.go
+++ b/service/controller/app/v1/resource/chart/delete.go
@@ -1,0 +1,78 @@
+package chart
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/controller"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desiredState interface{}) (*controller.Patch, error) {
+	delete, err := r.newDeleteChange(ctx, obj, currentState, desiredState)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	patch := controller.NewPatch()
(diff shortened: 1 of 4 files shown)


service/controller/app/v1/resource/chart/delete.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	patch := controller.NewPatch()
+	patch.SetDeleteChange(delete)
+
+	return patch, nil
+
+}
+
+func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange interface{}) error {
```
xh3b4sd: Moving that method up would be grand. We try to order by the alphabet where no other order makes more sense.

service/controller/app/v1/resource/chart/delete.go
```diff
@@ -0,0 +1,78 @@
...
+	delete, err := r.newDeleteChange(ctx, obj, currentState, desiredState)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	patch := controller.NewPatch()
+	patch.SetDeleteChange(delete)
+
+	return patch, nil
+
```
xh3b4sd: I think `gofmt` should do that. Don't know why that got not implemented by anybody. What kind of use could such a line have at the end of a function? The tooling could really support us here. 

```suggestion
```

service/controller/app/v1/resource/chart/delete.go
```diff
@@ -0,0 +1,77 @@
...
+		err := r.g8sClient.ApplicationV1alpha1().Charts(cr.Namespace).Delete(chart.Name, &metav1.DeleteOptions{})
+		if apierrors.IsNotFound(err) {
+			// fall through
+		} else if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleted the %#q chart", chart.Name))
+	} else {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "the chart do not need to be deleted")
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "the chart does not need to be deleted")
```

PR #26, opened 2019-01-29
Chart status 

tomahawk28:
- Adding status to `app` CR

- Problem 
1. I added `app` CR and and add status to `chart` CR manually. 
- `k create -f catalog.yml` 
- `k create -f app.yml`
- See the link[https://gist.github.com/tomahawk28/af3a71fedff127e9e728ed8391cb0eaf] for the attachment.
2. But it's end up in errors as below, 
```
E 01/29 07:58:16 /apis/application.giantswarm.io/v1alpha1/namespaces/default/apps/my-cool-prometheus chartstatusv1 stop reconciliation due to error | operatorkit/controller/controller.go:363 | controller=app-operator | event=update | loop=0 | version=8172
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/controller/controller.go:518:
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/controller/resource/metricsresource/resource_wrapper.go:39:
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/controller/resource/retryresource/resource_wrapper.go:79:
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/controller/resource/retryresource/resource_wrapper.go:67:
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/service/controller/app/v1/resource/chartstatus/create.go:45:
	the server could not find the requested resource (put apps.application.giantswarm.io my-cool-prometheus)
```


service/controller/app/v1/key/key.go
```diff
@@ -70,8 +74,8 @@ func ToChart(v interface{}) (v1alpha1.Chart, error) {
 	return *customResource, nil
 }
 
-func Version(customResource v1alpha1.App) string {
-	return customResource.Spec.Version
+func Release(customResource v1alpha1.App) string {
+	return customResource.Spec.Release
```
rossf7: Something has gone wrong here. We renamed this field from Release to Version because the name was clearer. Can you run `dep ensure -update` again?

https://github.com/giantswarm/apiextensions/blob/1f1e998db676f6bfc5698286fd697f5b4c0a55e5/pkg/apis/application/v1alpha1/app_types.go#L121

service/controller/app/v1/key/key.go
```diff
@@ -43,6 +43,10 @@ func SecretNamespace(customResource v1alpha1.App) string {
 	return customResource.Spec.Config.Secret.Namespace
 }
 
+func ReleaseStatus(customObject v1alpha1.App) string {
```
```suggestion
func AppStatus(customObject v1alpha1.App) string {
```

service/controller/app/v1/resource/chartstatus/create_test.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned/fake"
+	"github.com/giantswarm/micrologger/microloggertest"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	k8sfake "k8s.io/client-go/kubernetes/fake"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/kubeconfig"
+)
+
+func TestResource_EnsureCreated(t *testing.T) {
```
rossf7:
```suggestion
func Test_Resource_EnsureCreated(t *testing.T) {
```

service/controller/app/v1/resource/chartstatus/create.go
```diff
@@ -0,0 +1,49 @@
...
+	chart, err := g8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		return microerror.Maskf(notFoundError, "chart %#q in namespace %#q", name, r.watchNamespace)
+	}
+
+	if chart.Status.Status != "" && key.ReleaseStatus(customResource) != chart.Status.Status {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting chart %#q status as %#q", name, chart.Status.Status))
+		customResourceCopy := customResource.DeepCopy()
+		customResourceCopy.Status.LastDeployed = *chart.Status.LastDeployed.DeepCopy()
+		customResourceCopy.Status.Status = chart.Status.Status
```
rossf7: This looks a bit strange to me. customResourceCopy is an app CR and chart is a chart CR. I think it would be better to set the status fields individually. That also makes it clear what data is stored in the app CR status.

service/controller/app/v1/key/key.go
```diff
@@ -43,6 +43,10 @@ func SecretNamespace(customResource v1alpha1.App) string {
...
+	return customObject.Status.Status
```
```suggestion
	return customResource.Status.Status
```

service/controller/app/v1/key/key_test.go
```diff
@@ -289,7 +289,7 @@ func Test_VersionLabel(t *testing.T) {
 			result := VersionLabel(tc.obj)
 
 			if !reflect.DeepEqual(result, tc.expectedVersion) {
-				t.Fatalf("Version label == %#v, want %#v", result, tc.expectedVersion)
+				t.Fatalf("Release label == %#v, want %#v", result, tc.expectedVersion)
```
rossf7: This looks wrong.

service/controller/app/v1/resource/chartstatus/create.go
```diff
@@ -0,0 +1,49 @@
+package chartstatus
```
rossf7:
```suggestion
package status
```

Lets just call this `status` like in chart-operator. It sets the status for the app CRs.

service/controller/app/v1/resource/chartstatus/create.go
```diff
@@ -0,0 +1,49 @@
...
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	customResource, err := key.ToCustomResource(obj)
```
rossf7:
```suggestion
	cr, err := key.ToCustomResource(obj)
```

service/controller/app/v1/resource/chartstatus/create.go
rossf7: All 4 fields should be copied not just these 2.

service/controller/app/v1/resource/chartstatus/create.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("getting status for chart %#q", name))
+	chart, err := g8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		return microerror.Maskf(notFoundError, "chart %#q in namespace %#q", name, r.watchNamespace)
+	}
+
+	if chart.Status.Status != "" && key.ReleaseStatus(customResource) != chart.Status.Status {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting chart %#q status as %#q", name, chart.Status.Status))
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting app %#q status as %#q", name, chart.Status.Status))
```

This is setting the app status by reading the status of the chart CR. So the logging isn't accurate.

Also please add a newline to make this easier to read.

service/controller/app/v1/resource/chartstatus/create_test.go
```diff
@@ -0,0 +1,243 @@
...
+	tests := []struct {
+		name         string
+		obj          *v1alpha1.App
+		chart        *v1alpha1.Chart
```
rossf7: It would be good to add the App status to the tests struct. So you can check that the app status is set correctly.

service/controller/app/v1/resource/chartstatus/create.go
tomahawk28: Sure, I will copy other 2 variables as well.

tomahawk28:
What's going on with Circle CI? 💭 
```repo 'LICENSE' file does not contain required text 'Copyright 2016 - 2019 Giant Swarm GmbH'```
But LICENSE file is intact

**Comment by rossf7** [comment]
See https://gigantic.slack.com/archives/C04TGHDEF/p1548774785003300.

File: service/controller/app/v1/resource/status/create.go [hunk 11]
```diff
@@ -0,0 +1,53 @@
...
+
+	if chart.Status.Status != "" && key.AppStatus(cr) != chart.Status.Status {
+		r.logger.LogCtx(ctx, "level", "debug", "message",
+			fmt.Sprintf("setting app %#q status as %#q", name, chart.Status.Status))
+
+		customResourceCopy := cr.DeepCopy()
+		customResourceCopy.Status.LastDeployed = *chart.Status.LastDeployed.DeepCopy()
+		customResourceCopy.Status.Status = chart.Status.Status
+		customResourceCopy.Status.Version = chart.Status.Version
+		customResourceCopy.Status.AppVersion = chart.Status.AppVersion
```
rossf7: Alpha order pls.

service/controller/app/v1/resource/status/create.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	name := key.AppName(cr)
+
+	g8sClient, err := r.kubeConfig.NewG8sClientForApp(ctx, cr)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("getting status for chart %#q", name))
```
rossf7:
```suggestion
	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("getting status for app %#q", name))
```

This logging isn't clear to me. We need to get the status the `chart` CR status and use it to set the status for the `app` CR.

Maybe we should also adding logging for `finding status for chart %#q"` and `found status for chart %#q"`
https://github.com/giantswarm/fmt/blob/master/go/log_messages.md#fact-gathering WDYT?

service/controller/app/v1/resource/status/create.go
```diff
@@ -0,0 +1,53 @@
...
+		customResourceCopy.Status.Status = chart.Status.Status
+		customResourceCopy.Status.Version = chart.Status.Version
+		customResourceCopy.Status.AppVersion = chart.Status.AppVersion
+
+		_, err = r.g8sClient.ApplicationV1alpha1().Apps(cr.Namespace).UpdateStatus(customResourceCopy)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("status set for chart %#q", name))
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("status set for app %#q", name))
```

Log line below also needs to be updated.

service/controller/app/v1/resource/status/create.go
rossf7: These should be in alpha order.

service/controller/app/v1/resource/status/create.go
tomahawk28: I'll change this log `finding status for chart %#q` and add one more log as `found status for chart %#`, in that way we are more clear I guess.

service/controller/app/v1/key/key.go
```diff
@@ -35,6 +35,10 @@ func SecretNamespace(customResource v1alpha1.App) string {
 	return customResource.Spec.Config.Secret.Namespace
 }
 
+func AppStatus(customResource v1alpha1.App) string {
+	return customResource.Status.Status
```
xh3b4sd: `Status.Status` is not cool. Can we improve that naming? What kind of status is this about? What are the possible values?

service/controller/app/v1/resource/status/resource.go
```diff
@@ -0,0 +1,74 @@
...
+		k8sClient:  config.K8sClient,
+		logger:     config.Logger,
+
+		watchNamespace: config.WatchNamespace,
+	}
+
+	return r, nil
+}
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
```
xh3b4sd: There is already a `create.go`. Can we please have a `delete.go`?

service/controller/app/v1/resource_set.go
```diff
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"context"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/resource/status"
```
xh3b4sd: Let's move that down.

service/controller/app/v1/resource/status/create.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	name := key.AppName(cr)
+
+	g8sClient, err := r.kubeConfig.NewG8sClientForApp(ctx, cr)
```
xh3b4sd: That resource is now pretty expensive and takes a rather long time to reconcile constantly. Would it make sense to not compute the client and fetch the chart status in case the status was already set to what we want it to be?

service/controller/app/v1/resource/status/create.go
```diff
@@ -0,0 +1,54 @@
...
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding status for chart %#q", name))
+	chart, err := g8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		return microerror.Maskf(notFoundError, "chart %#q in namespace %#q", name, r.watchNamespace)
+	}
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found status for chart %#q", name))
+
+	if chart.Status.Status != "" && key.AppStatus(cr) != chart.Status.Status {
+		r.logger.LogCtx(ctx, "level", "debug", "message",
+			fmt.Sprintf("setting app %#q status as %#q", name, chart.Status.Status))
```
xh3b4sd: Let's please not do that. It increases the code complexity without providing real value. Having one line should be just fine. These things are way harder to read personally.

service/controller/app/v1/resource/status/create.go
tomahawk28: Good catch, we don't need to create instance even though it's not needed

service/controller/app/v1/key/key.go
xh3b4sd: Can we then have `Release.Status`? I think the structure makes total sense. Like it a lot in the VOO specs.

service/controller/app/v1/resource/status/create.go
kopiczko: We can do the same trick we do for BYOC AWS clients. We can store it in the context by one resource and then fetch in other.

tomahawk28: Sorry to back and forth, but we need this g8sclient to know that the chart status in tenant cluster, otherwise we would never know which status we want to set to `app` status at the first place.

xh3b4sd: Then let's do what Pawel suggested with the context. In case the app status is used in multiple places, we should only fetch it once and put it into the controller context. Then we do not have to fetch it multiple times. The context filling could either be done in a separate resource depending on complexity, or in the controller init function.

rossf7: > We can store it in the context by one resource and then fetch in other. ❤️ 

I like that a lot. We do need to check the tenant cluster for App Catalog but this would reduce the amount of processing needed.

https://github.com/giantswarm/aws-operator/blob/master/service/controller/v22/controllercontext/context.go

@tomahawk28 Here is an example from aws-operator. I'm not deep into the approach used on BYOC (Bring Your Own Cloud) but you can check with Pawel and Theo.

service/controller/app/v1/key/key.go
rossf7: Yes I think that works.

tomahawk28: Should I wait until it merges into apiextention? or we can have it later?

rossf7: @tomahawk28 We can deal with that separately IMO. I'll take it as I want to keep the chart and app CR statuses aligned. I've just been busy with other things.

But I don't see any changes related to storing the g8s client in the context. Have you looked into that?

service/controller/app/v1/resource/status/create.go
tomahawk28: I can do add `g8sClient` inside of context, rather than `status` itself, is that right? 
Since we need to check the status regularly, caching it doesn't really help I guess.

kopiczko: Yes you add g8sClient to the context.

tomahawk28: There are few constraints still. 
1. What if `app` CR update their `secret` name? or content itself? 
Then we should invalidate g8sClient inside of running operator, but it's far more complicated then this PR :)

2. Different `app` object meta name and ns
We should have different g8sclient for each of app CR I guess, but is it okay?

rossf7:
I'm not deep into the context caching but IMO we should cache it once per
reconciliation loop. That way we only fetch the secret once per loop.

If the CR spec changes it will be a new loop. If the secret changes we will
detect it after the next resync period (usually 5 mins).

service/controller/app/v1/resource/status/create.go
xh3b4sd: The client is only added to the context which is about the scope of the reconciliation loop. I think the secret name will not change while a loop is executed. When the secret name changes the operator code changes and thus will be redeployed, causing further loops to be executed in the aligned way. 

Adding the client in general is a nice best practise we should maybe also document for SIG Operator. Also consider doing the same for the status itself in case this is used in multiple places. This would then be another level of optimization.

service/controller/app/v1/resource/chart/update.go
```diff
@@ -42,12 +43,16 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
...
-		g8sClient, err := r.kubeConfig.NewG8sClientForApp(ctx, cr)
+		ctlctx, err := controllercontext.FromContext(ctx)
 		if err != nil {
 			return microerror.Mask(err)
 		}
 
-		_, err = g8sClient.ApplicationV1alpha1().Charts(cr.Namespace).Update(&chart)
+		if err != nil {
+			return microerror.Mask(err)
+		}
```
rossf7: This looks like a dupe?

service/controller/app/v1/resource_set.go
```diff
@@ -114,6 +134,21 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
+
+		g8sClient, err := kubeConfig.NewG8sClientForApp(ctx, cr)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		c := controllercontext.Context{
+			G8sClient: g8sClient,
+		}
+		ctx = controllercontext.NewContext(ctx, c)
```
rossf7: <3

service/controller/app/v1/resource/status/create.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	name := key.AppName(cr)
+
+	ctlctx, err := controllercontext.FromContext(ctx)
```
kopiczko:
```suggestion
	ctlCtx, err := controllercontext.FromContext(ctx)
```

service/controller/app/v1/resource/status/create.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding status for chart %#q", name))
+
+	chart, err := ctlctx.G8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, metav1.GetOptions{})
```
kopiczko:
```suggestion
	chart, err := ctlCtx.G8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, metav1.GetOptions{})
```

service/controller/app/v1/resource/status/create.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return microerror.Maskf(notFoundError, "chart %#q in namespace %#q", name, r.watchNamespace)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found status for chart %#q", name))
+
+	if chart.Status.Status != "" && key.AppStatus(cr) != chart.Status.Status {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting app %#q status as %#q", name, chart.Status.Status))
+
+		customResourceCopy := cr.DeepCopy()
```
kopiczko: Nit: I think after we agreed to name `customResource` a `cr` we can name it `crCopy`.

service/controller/app/v1/resource/chart/delete.go
```diff
@@ -25,12 +26,12 @@ func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange inte
 	if chart.Name != "" {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting the %#q chart", chart.Name))
 
-		g8sClient, err := r.kubeConfig.NewG8sClientForApp(ctx, cr)
+		ctlctx, err := controllercontext.FromContext(ctx)
```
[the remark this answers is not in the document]
kopiczko: cc is a CloudConfig to me :D I prefer `ctlCtx`.

service/controller/app/v1/resource/status/create.go
kopiczko: We talked. This is already implemented.

service/controller/app/v1/key/key.go
rossf7: Spec PR is up that includes the status changes. Please discuss there and I'll take care once we're happy.

https://github.com/giantswarm/giantswarm/pull/5240

rossf7: The CRD changes are separate but AppStatus should be ordered in the file.

LICENSE
```diff
@@ -178,15 +178,15 @@
    APPENDIX: How to apply the Apache License to your work.
 
       To apply the Apache License to your work, attach the following
-      boilerplate notice, with the fields enclosed by brackets "[]"
+      boilerplate notice, with the fields enclosed by brackets "{}"
```
xh3b4sd: Why does this actually change?

service/controller/app/v1/controllercontext/context.go
```diff
@@ -0,0 +1,29 @@
...
+
+type contextKey string
+
+const controllerKey contextKey = "controller"
+
+type Context struct {
+	G8sClient versioned.Interface
+}
+
+func NewContext(ctx context.Context, c Context) context.Context {
```
xh3b4sd: IMO when `FromContext` returns a pointer, `NewContext` should take one.

service/controller/app/v1/resource/chart/create.go
```diff
@@ -23,12 +24,12 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 	if chart.Name != "" {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensuring creation of chart %#q", chart.Name))
 
-		g8sClient, err := r.kubeConfig.NewG8sClientForApp(ctx, cr)
+		ctlctx, err := controllercontext.FromContext(ctx)
```
xh3b4sd: The naming so far was `cc` or `ctlCtx`.

service/controller/app/v1/resource/chart/current_test.go
```diff
@@ -153,7 +154,12 @@ func Test_Resource_GetCurrentState(t *testing.T) {
 				t.Fatalf("error == %#v, want nil", err)
 			}
 
-			result, err := r.GetCurrentState(context.TODO(), tc.obj)
+			ctlConfig := controllercontext.Context{
+				G8sClient: g8sClient,
+			}
+			ctx := controllercontext.NewContext(context.TODO(), ctlConfig)
```
xh3b4sd: Mr Pawel usually has a stronger opinion on not using `TODO` in tests.

service/controller/app/v1/resource/status/create.go
```diff
@@ -0,0 +1,56 @@
...
+	ctlCtx, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding status for chart %#q", name))
+
+	chart, err := ctlCtx.G8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		return microerror.Maskf(notFoundError, "chart %#q in namespace %#q", name, r.watchNamespace)
```
xh3b4sd: Well, `Get` can return all kinds of errors. Let's please check for the specific error we want to handle and have the catch all as last resort only. What we would do now is to blindly return a very specific error even if the actual error is because of whatever reason.

service/controller/app/v1/resource/status/create.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found status for chart %#q", name))
+
+	if chart.Status.Status != "" && key.AppStatus(cr) != chart.Status.Status {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting app %#q status as %#q", name, chart.Status.Status))
+
+		crCopy := cr.DeepCopy()
```
xh3b4sd: The meta data resource version is not set. When we do not set it there will be errors because the CR got updated by someone else meanwhile and this code will fail.

service/controller/app/v1/resource/status/create.go
tomahawk28: Such as `not found` error you mean?

LICENSE
tomahawk28: I copy & paste from other repo, will revert it then.

service/controller/app/v1/resource/status/create.go
tomahawk28: I'm not sure how to set meta data setup since it's not introduced from other repos (chart-operator, cluster-operator), would you tell me which part needs to be done?

service/controller/app/v1/resource/status/create.go
xh3b4sd: Yes, we should only return `notFoundError` in case it is really a not found error. It does not make sense to return `notFoundError` when there is actually a broken pipe or 500 or whatnot. That only confuses the shit out of people who have to debug something in the mids of a oncall shit storm.

service/controller/app/v1/controllercontext/context.go
tomahawk28: we could do it, but I don't want to take apart from its original source, such as `aws-operator`

tomahawk28:
@xh3b4sd PTAL when you have time. thanks

PR #27, opened 2019-01-31
Structure changes for appcatalog and app CRDs

rossf7:
Follow up to https://github.com/giantswarm/apiextensions/pull/196. I also aligned the status resource with chart-operator.


service/controller/app/v1/resource/chart/desired.go
```diff
@@ -31,7 +31,7 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
-	tarballURL, err := generateTarballURL(appcatalogkey.CatalogStorageURL(*appCatalog), key.AppName(cr), key.Version(cr))
+	tarballURL, err := generateTarballURL(appcatalogkey.AppCatalogStorageURL(*appCatalog), key.AppName(cr), key.Version(cr))
```
rossf7: Having the App prefix makes the func name clearer for me.

service/controller/app/v1/resource/status/create.go
```diff
@@ -36,15 +37,22 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting app %#q status as %#q", name, chart.Status.Status))
+	chartStatus := key.ChartStatus(*chart)
+	desiredStatus := v1alpha1.AppStatus{
+		AppVersion: chartStatus.AppVersion,
+		Release: v1alpha1.AppStatusRelease{
+			LastDeployed: *chartStatus.Release.LastDeployed.DeepCopy(),
+			Status:       chartStatus.Release.Status,
+		},
+		Version: chartStatus.Version,
+	}
```
rossf7: Approach is updated a bit so we calculate the desired state and then use the `equals` func for comparison.  This means all fields are compared.

service/controller/app/v1/resource/status/create.go
```diff
@@ -36,15 +37,22 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+
+	if !equals(desiredStatus, key.AppStatus(cr)) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting status for app %#q", name))
```
rossf7: I simplified the logging a bit.

service/controller/app/v1/resource/status/resource.go
```diff
@@ -66,3 +67,22 @@ func New(config Config) (*Resource, error) {
...
 	return Name
 }
+
+// equals asseses the equality of AppStatuses with regards to distinguishing
+// fields.
+func equals(a, b v1alpha1.AppStatus) bool {
+	if a.AppVersion != b.AppVersion {
+		return false
+	}
+	if a.Release.LastDeployed != b.Release.LastDeployed {
```
tomahawk28: Should we compare `LastDeployed` as well? I can imagine that a situation that only this timestamp would have changed inside chart CR but anything else.

rossf7: The `LastDeployed` is the value we get from `helmclient` and that is also shown by `helm ls`. I don't see a downside to updating it if it changes.

PR #28, opened 2019-02-01
e2e testing boilerplate

tomahawk28:
After 20+ commits, it passed.


.circleci/config.yml
```diff
@@ -19,3 +47,32 @@ jobs:
...
+              only: master
+          requires:
+            - e2eTestChartInstall
+
+      - publish_to_stable:
+          filters:
+            branches:
+              only: master
+          requires:
+            - deploy
```
MarcelMue: Check your editor settings

integration/setup/setup.go
```diff
@@ -0,0 +1,79 @@
...
+
+func resources(ctx context.Context, h *framework.Host, crdClient *k8scrdclient.CRDClient, l micrologger.Logger) error {
+	var err error
+
+	err = crdClient.EnsureCreated(ctx, v1alpha1.NewChartCRD(), backoff.NewExponential(backoff.ShortMaxWait, backoff.ShortMaxInterval))
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	err = h.InstallStableOperator("app-operator", "app", "")
```
MarcelMue: `InstallStableOperator` is the wrong method here as it will install the `master` branch of your operator.
Where did you take this boilerplate from?

The goto way to do this should be like in `aws-operator`: https://github.com/giantswarm/aws-operator/blob/master/integration/setup/setup.go#L118 
This uses the release package which you can find here:
https://github.com/giantswarm/e2e-harness/blob/master/pkg/release/release.go#L286

The `release` package uses `helmClient` inside to make creation more graceful and give you better error handling.

integration/test/chart/main_test.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+var (
+	h          *framework.Host
+	helmClient *helmclient.Client
+	l          micrologger.Logger
+	crdClient  *k8scrdclient.CRDClient
+)
+
+func init() {
```
MarcelMue: There are a few clients missing in your setup here, this piece of code should give you all the necessary clients:
https://github.com/giantswarm/aws-operator/blob/master/integration/setup/config.go#L49-L134

integration/setup/setup.go
MarcelMue: See especially the `release.NewVersion(env.CircleSHA())` to obtain the correct version for your branch operator.

tomahawk28:
Should we use `EnsureTenantClusterCreated` & `EnsureTenantClusterDeleted` in this case? @rossf7 ?

rossf7:
@tomahawk28 No we don't need that. No tenant cluster is created. That is just for projects like aws-operator where we do need to create tenant clusters.

.circleci/config.yml
```diff
@@ -19,3 +47,32 @@ jobs:
           fi
 
 
+  e2eTestChartInstall:
+    environment:
+      TEST_DIR: "integration/test/chart"
+      TESTED_VERSION: "current"
```
rossf7: Can this be removed? As we don't need the version bundle support.

helm/app-operator-chart/templates/deployment.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apps/v1
+apiVersion: extensions/v1beta1
```
rossf7: The version has regressed. We should use `apps/v1`.

helm/app-operator-chart/templates/deployment.yaml
```diff
@@ -27,7 +27,6 @@ spec:
           - key: config.yaml
             path: config.yaml
       serviceAccountName: {{ .Values.name }}
-      dnsPolicy: None
```
rossf7: <3

integration/env/common.go
```diff
@@ -0,0 +1,53 @@
...
+const (
+	// EnvVarCircleCI is the process environment variable representing the
+	// CIRCLECI env var.
+	EnvVarCircleCI = "CIRCLECI"
+	// EnvVarCircleSHA is the process environment variable representing the
+	// CIRCLE_SHA1 env var.
+	EnvVarCircleSHA = "CIRCLE_SHA1"
+	// EnvVarGithubBotToken is the process environment variable representing
+	// the GITHUB_BOT_TOKEN env var.
+	EnvVarGithubBotToken = "GITHUB_BOT_TOKEN"
```
rossf7: This constant isn't needed.

integration/env/common.go
```diff
@@ -0,0 +1,53 @@
...
+	// EnvVarKeepResources is the process environment variable representing the
+	// KEEP_RESOURCES env var.
+	EnvVarKeepResources = "KEEP_RESOURCES"
+	// EnvVarTestedVersion is the process environment variable representing the
+	// TESTED_VERSION env var.
+	EnvVarTestedVersion = "TESTED_VERSION"
```
rossf7: Same here.

integration/setup/config.go
```diff
@@ -0,0 +1,134 @@
...
+	Host    *framework.Host
+	K8s     *k8s.Setup
+	Release *release.Release
+	Logger  micrologger.Logger
+
+	// UseDefaultTenant defines whether the standard test setup should ensure the
+	// default tenant cluster. This is enabled by default. Most tests simply use
+	// the standard test tenant cluster. One exception is IPAM. There we launch
+	// multiple tenant clusters and do not want to make use of the default one.
+	UseDefaultTenant bool
```
rossf7: Lets remove this. 

Its not needed now and we can add it later when we add tests that use a tenant cluster.

integration/setup/setup.go
```diff
@@ -0,0 +1,80 @@
...
+
+func Setup(m *testing.M, config Config) {
+	ctx := context.Background()
+
+	var v int
+	var err error
+
+	err = installResources(ctx, config)
+	if err != nil {
+		config.Logger.LogCtx(ctx, "level", "error", "message", "failed to install AWS resources", "stack", fmt.Sprintf("%#v", err))
```
rossf7: AWS is wrong here.

integration/setup/teardown.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+
+	"github.com/giantswarm/e2e-harness/pkg/framework"
+	"github.com/giantswarm/microerror"
+)
+
+func teardown(ctx context.Context, config Config) error {
+	// clean host cluster components
```
rossf7:
```suggestion
	// clean control plane components
```

integration/setup/teardown.go
```diff
@@ -0,0 +1,29 @@
...
+	err := framework.HelmCmd("delete --purge giantswarm-app-operator")
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// clean guest cluster components
```
rossf7:
```suggestion
	// clean tenant cluster components
```

Is this needed? I'm not sure this chart is actually installed.

helm/app-operator-chart/templates/deployment.yaml
tomahawk28: I'm not sure why `apps/v1` gave an error, ```{"caller":"github.com/giantswarm/app-operator/vendor/github.com/giantswarm/backoff/notifier.go:13","level":"warning","message":"retrying backoff in '582.280027ms' due to error","stack":"[{/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:488: } {rpc error: code = Unknown desc = release giantswarm-app-operator failed: Deployment.apps \"app-operator\" is invalid: [spec.selector: Required value, spec.template.metadata.labels: Invalid value: map[string]string{\"giantswarm.io/service-type\":\"managed\", \"app\":\"app-operator\"}: `selector` does not match template `labels`]}]","time":"2019-02-05T12:21:35.917182+00:00"}``` 
That's why I change this to `extnsions/v1beta`, any idea?

rossf7: `extensions/v1beta1` is an old API version for the deployment resource. We should move to `apps/v1` wherever possible and especially for new components.

The validation for `apps/v1` is a bit stricter so you need to add the YAML below. For the old API version this selector was optional.

```
  selector:
    matchLabels:
      app: {{ .Values.name }}
```

PR #30, opened 2019-02-04
Add create and update support for configmap resource

rossf7:

### DIFF
--- a/service/controller/app/v1/resource/configmap/create.go
+++ b/service/controller/app/v1/resource/configmap/create.go
@@ -1,0 +1,61 @@
+package configmap
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+)
+
+func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
+	configMap, err := toConfigMap(createChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if configMap.Name != "" {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensuring creation of configmap %#q", configMap.Name))
+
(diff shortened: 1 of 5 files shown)


service/controller/app/v1/resource/configmap/create.go
```diff
@@ -0,0 +1,61 @@
...
+		_, err = ctlCtx.K8sClient.CoreV1().ConfigMaps(configMap.Namespace).Create(configMap)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensured creation of configmap %#q", configMap.Name))
+	} else {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("no need to create configmap"))
+	}
+	return nil
```
xh3b4sd:
```suggestion

	return nil
```

service/controller/app/v1/resource/configmap/create_test.go
```diff
@@ -0,0 +1,124 @@
...
+					Namespace: "default",
+				},
+			},
+			expectedConfigMap: &corev1.ConfigMap{},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			c := Config{
```
tomahawk28: I think we can move `c` out of this loop, right?

rossf7:
> We could maybe come up with a general one.

I like the idea but there is some app-operator specific stuff. We need to use k8s clients from the context because the operator needs to connect to control plane and tenant clusters.

My concern is more with the desired state. We need merge logic that is quite complex and basically the same for configmaps and secrets. We also need to merge the configmap and secret values in chart-operator. 

I'm going to add a `values` package in app-operator but that could become a library.

service/controller/app/v1/resource/configmap/create_test.go
rossf7: Thanks good point. Done.

rossf7:
Going with this. I'll keep in mind if we can do something with more generic k8s resources.

PR #33, opened 2019-02-07
Add chartcrd resource

tomahawk28:
- Upon booting app-operator, we need to access to tenant cluster and create chart CRD beforehand.

### DIFF
--- a/service/controller/app/v1/resource/chartcrd/create.go
+++ b/service/controller/app/v1/resource/chartcrd/create.go
@@ -1,0 +1,32 @@
+package chartcrd
+
+import (
+	"context"
+	"time"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/backoff"
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	var err error
+	ctlCtx, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring chart crd creation on tenant cluster")
+
+	crdBackoff := backoff.NewMaxRetries(3, 1*time.Second)
+	err = ctlCtx.CRDClient.EnsureCreated(ctx, v1alpha1.NewChartCRD(), crdBackoff)
+	if err != nil {
+		r.logger.LogCtx(ctx, "level", "debug", "error", "failed to ensured chart crd creation on tenant cluster")
+		return microerror.Mask(err)
+	}
+	r.logger.LogCtx(ctx, "level", "debug", "message", "ensured chart crd creation on tenant cluster")
+
+	return nil
+}
--- a/service/controller/app/v1/resource/chartcrd/create.go
+++ b/service/controller/app/v1/resource/chartcrd/create.go
@@ -30,10 +30,6 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			return microerror.Mask(err)
 		}
 
-		if err != nil {
-			return microerror.Mask(err)
-		}
-
 		k8sExtClient, err := apiextensionsclient.NewForConfig(ctlCtx.RESTConfig)
 		if err != nil {
 			return microerror.Mask(err)
@@ -48,8 +44,10 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		crdBackoff := backoff.NewMaxRetries(3, 1*time.Second)
 
 		err = crdClient.EnsureCreated(ctx, v1alpha1.NewChartCRD(), crdBackoff)
-		if err != nil {
-			r.logger.LogCtx(ctx, "level", "debug", "error", "failed to ensured chart crd creation on tenant cluster")
+		if IsNotEstablished(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "chart crd in creation at the moment")
+			return nil
+		} else if err != nil {
 			return microerror.Mask(err)
 		}
 		r.logger.LogCtx(ctx, "level", "debug", "message", "ensured chart crd creation on tenant cluster")
--- a/service/controller/app/v1/resource/chartcrd/create.go
+++ b/service/controller/app/v1/resource/chartcrd/create.go
@@ -5,28 +5,57 @@ import (
 	"time"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
 	"github.com/giantswarm/backoff"
 	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/client/k8scrdclient"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 
 	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
 )
 
 func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
-	var err error
-	ctlCtx, err := controllercontext.FromContext(ctx)
+	cr, err := key.ToCustomResource(obj)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
-	r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring chart crd creation on tenant cluster")
+	secretName := key.SecretName(cr)
 
-	crdBackoff := backoff.NewMaxRetries(3, 1*time.Second)
-	err = ctlCtx.CRDClient.EnsureCreated(ctx, v1alpha1.NewChartCRD(), crdBackoff)
-	if err != nil {
-		r.logger.LogCtx(ctx, "level", "debug", "error", "failed to ensured chart crd creation on tenant cluster")
-		return microerror.Mask(err)
+	if secretName != "" {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring chart crd creation on tenant cluster")
+
+		ctlCtx, err := controllercontext.FromContext(ctx)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		k8sExtClient, err := apiextensionsclient.NewForConfig(ctlCtx.RESTConfig)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		c := k8scrdclient.Config{
+			K8sExtClient: k8sExtClient,
+			Logger:       r.logger,
+		}
+
+		crdClient, err := k8scrdclient.New(c)
+		crdBackoff := backoff.NewMaxRetries(3, 1*time.Second)
+
+		err = crdClient.EnsureCreated(ctx, v1alpha1.NewChartCRD(), crdBackoff)
+		if err != nil {
+			r.logger.LogCtx(ctx, "level", "debug", "error", "failed to ensured chart crd creation on tenant cluster")
+			return microerror.Mask(err)
+		}
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensured chart crd creation on tenant cluster")
+	} else {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "do not need to create chart crd on tenant cluster")
 	}
-	r.logger.LogCtx(ctx, "level", "debug", "message", "ensured chart crd creation on tenant cluster")
 
 	return nil
 }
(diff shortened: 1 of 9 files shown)


service/controller/app/v1/resource/chartcrd/create.go
```diff
@@ -0,0 +1,61 @@
...
+			K8sExtClient: k8sExtClient,
+			Logger:       r.logger,
+		}
+
+		crdClient, err := k8scrdclient.New(c)
+		crdBackoff := backoff.NewMaxRetries(3, 1*time.Second)
+
+		err = crdClient.EnsureCreated(ctx, v1alpha1.NewChartCRD(), crdBackoff)
+		if err != nil {
+			r.logger.LogCtx(ctx, "level", "debug", "error", "failed to ensured chart crd creation on tenant cluster")
```
xh3b4sd: I think logging and returning an error at the same time is an anti-pattern. The returned error will be logged at the end anyway and we do not need it twice. I would like to drop that line.

service/controller/app/v1/resource/chartcrd/create.go
```diff
@@ -0,0 +1,61 @@
...
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring chart crd creation on tenant cluster")
+
+		ctlCtx, err := controllercontext.FromContext(ctx)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		if err != nil {
+			return microerror.Mask(err)
+		}
```
xh3b4sd: I think that error handling here is not necessary.

service/controller/app/v1/resource/chartcrd/resource.go
```diff
@@ -0,0 +1,53 @@
...
+type Config struct {
+	G8sClient  versioned.Interface
+	K8sClient  kubernetes.Interface
+	KubeConfig kubeconfig.Interface
+	Logger     micrologger.Logger
+
+	WatchNamespace string
+}
+
+// Resource implements the chartstatus resource.
```
xh3b4sd: I am wondering right now why the resource is installing the CRD and where. We could maybe have some more useful godoc and describe in which part of the infrastructure this resource acts in and why it does so. Just like drawing a little picture. What do you think about that?

service/controller/app/v1/resource_set.go
```diff
@@ -159,19 +175,34 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
+		secretName := key.SecretName(cr)
 
-		k8sClient, err := kubeConfig.NewK8sClientForApp(ctx, cr)
-		if err != nil {
-			return nil, microerror.Mask(err)
+		var restConfig *rest.Config
+		var g8sClient versioned.Interface
+		var k8sClient kubernetes.Interface
+
+		if secretName != "" {
```
xh3b4sd: It is a bit odd that the secret name decides when to initialize the clients but the interface of initializing the clients does not seem to be concerned with the secret name. That smells a bit.

service/controller/app/v1/resource_set.go
```diff
@@ -159,19 +175,34 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
+			if err != nil {
+				return nil, microerror.Mask(err)
+			}
+			k8sClient, err = kubernetes.NewForConfig(restConfig)
+			if err != nil {
+				return nil, microerror.Mask(err)
+			}
+		} else {
+			g8sClient = config.G8sClient
+			k8sClient = config.K8sClient
```
xh3b4sd: Why are we setting the clients to something else than a tenant specific client? These are usually clients accessing the control plane. Why should we set clients differently here? I miss some background information I guess.

service/controller/app/v1/resource/chartcrd/create.go
```diff
@@ -0,0 +1,61 @@
...
+
+		c := k8scrdclient.Config{
+			K8sExtClient: k8sExtClient,
+			Logger:       r.logger,
+		}
+
+		crdClient, err := k8scrdclient.New(c)
+		crdBackoff := backoff.NewMaxRetries(3, 1*time.Second)
+
+		err = crdClient.EnsureCreated(ctx, v1alpha1.NewChartCRD(), crdBackoff)
```
xh3b4sd: When `EnsureCreated` would return an error indicating that the creation got initialized but is not yet completed, we would not need the backoff and could cancel the resource without returning an error. A resource returning an error causes the `retryresource` to execute the resource here again and we start over again, ending up in a wait lock for some time and only throwing errors. I would suggest to use some more specific error handling and cancel the resource gracefully when we notice the CRD installation still takes a bit more time.

service/controller/app/v1/resource/chartcrd/create.go
tomahawk28: Thanks, I will delete this part.

service/controller/app/v1/resource_set.go
tomahawk28: In this specific case, we don't have `secret` information in `app` CR which is essential to access to tenant cluster. Therefore we regard it intended as install inside control plane.

service/controller/app/v1/resource_set.go
tomahawk28: As explained in here, https://github.com/giantswarm/app-operator/pull/33#discussion_r254638721, `secret` information is necessary to access to tenant cluster.

service/controller/app/v1/resource/chartcrd/create.go
```diff
@@ -0,0 +1,59 @@
+package chartcrd
```
kopiczko: Why does it need to create this CRD? Isn't chart-operator running in the tenant cluster?

service/controller/app/v1/resource_set.go
xh3b4sd: I see. That implementation detail can lead to pretty serious consequences. I think we should add some logging to express what is going on and why we use which client. I suppose it is a feature to work against control planes and tenant clusters in the same way. I am just wondering if it can more easily happen to miss the secret even though we intent to work against a tenant cluster.

service/controller/app/v1/resource_set.go
xh3b4sd: I understood this, but the decision making the the interface diverge so there are different parts doing the same thing, which is just not obvious in the first place.

service/controller/app/v1/resource/chartcrd/create.go
xh3b4sd: I had the same question in https://github.com/giantswarm/app-operator/pull/33#discussion_r254598559. I would like to understand this better as well.

tomahawk28: In the case of chart-operator was not running in tenant cluster beforehand, app-operator could not create `chart` CR. That's what I got from @rossf7, and he pointed out that another place that same practice is ongoing, https://github.com/giantswarm/cluster-operator/pull/377/files#diff-6ec70deab7af8061be1a5eb8b75ed5e9R20
WDYT?

service/controller/app/v1/resource/chartcrd/resource.go
tomahawk28: Sorry, we can keep go in https://github.com/giantswarm/app-operator/pull/33#discussion_r254644232

service/controller/app/v1/resource/chartcrd/create.go
xh3b4sd: It would be good to have a diagram to visualize who is supposed to do what. I do not understand the circumstances in which who should install what.

tomahawk28: I would suggest to make it another PR in here, https://github.com/giantswarm/sig-operator, 
Just a min.

xh3b4sd: Not sure. This here is pretty specific and not a general thing as I understood it. Maybe just try to explain how this here is supposed to look like, as it does not seem to be clear for several people now.

kopiczko: Does chart-operator run in the tenant cluster? I think yes, because how otherwise it would watch CRs from multiple clusters. If so why app-operator can't cancel for as long as chart CRD isn't created by the chart-operator?

puja108:
There seem to be a few questions here, I cannot solve all of them as some are related to things that @rossf7 has thought of. Still I feel there's a few things we need to be careful about if I understood above discussions right:

1. A differentiation between CP and TCs being made on there being a secret or not seems risky indeed. The secret in question here is part of the kubeConfig section of app CRD. I was thinking we could just not differentiate between CP and TC and just use kubeConfigs for both instead of InCluster for the CP. Otherwise, we might need to make it more explicit. e.g. as we use namespaces for the app CRs, a user creating an app CR in their cluster namespace should never be escalated to be able to install sth even by accident on the control plane, that should only work in `giantswarm` namespace maybe. But again it smells doing sth based on a specific ns, so I would prefer being more explicit there.

2. I also feel a bit of an issue when app-operator tries to install chart CRDs. I feel it should rely on chart-operator and if it's not there, it might just need to retry later. I'm not sure of the specific case @rossf7 might have had in mind there, but in general it does smell a bit.

tomahawk28:
1. There are statements about how we should use `app` CR in the case of missing `KubeConfig` 
 ```If no kubeconfig is specified in the spec then app-operator connects to the local cluster (i.e. the Control Plane it's running in) using in-cluster config.```
https://github.com/giantswarm/giantswarm/blob/master/roadmap/MANAGED-SERVICES-CATALOG.MD#kubeconfig
If we agree on that point to make it explicit, then I guess we need to update this as well. 
2. It's also unclear for me now, since the reference from Ross also has not merged it yet. 
Is this new case? I mean creating CRD for tenant cluster from another operator?

xh3b4sd:
Always having a secret and only acting upon it makes sense to me, in case I got that part right. Who creates the app CRs and would need to do it for the control plane as well?

tomahawk28:
`app` CR is created by cluster-operator, Should we update the spec as well?

xh3b4sd:
Updating the spec makes sense, then more people can maybe comment and agree on.

puja108:
App CRs are also created by the user.

I'd suggest we wait for next week and talk with @rossf7 to find the right solution. I do not want to revert it without hearing his side

tomahawk28:
Sure, I will stop making progress on this issue.

service/controller/app/v1/resource_set.go
kopiczko: To me it's pretty clear. It was described in the spec. When no Secret is specified in the App CR spec then app-operator creates chart CRs using in-cluster credentials, otherwise with credentials from referenced Secret.

xh3b4sd:
I think I understand the matter a bit better after talking with Ji and Marcel on Friday. I think installing CRDs should be done by the operator who watches its corresponding CRs. Having a resource for this in `app-operator` does not make sense to me. E2E tests can be written how we do already. The provider operators need `node-operator` to be present because of the `DrainerConfig` CR. They do not install it themselves and should not. We simply install `node-operator` in the e2e tests, like we do in production. Same should apply here. I would be curious what alternatives we have for improving the performance of the overall system.

rossf7:
In general I think operators should install their CRDs. This is a specific timing problem with installing the chart CRD in tenant clusters.

When app-operator needs to bootstrap a tenant cluster it can only proceed as far as installing chart-operator. Creating chart CRs fails because the CRD does not exist. So we need to wait up to 5 mins for the next resync period.

I'm working on the same issue for a PM in cluster-operator. By creating the CRD the CRs can also be created. As soon as chart-operator boots it can start installing components.

The current 5 mins delay is a poor UX and app-operator will have the same problem. This goes away once we stop using Tiller and stop running chart-operator in tenant clusters.

puja108:
I also feel this is a bit ugly. I might have mentioned this before, but I think we could add CRD creation to the cluster bootstrap (i.e. in ignition just like RBAC bootstrap) as we know which CRDs are needed in TC and CP respectively. We would still have the "ensuring" by the operators just to be sure (also in case someone manually removed or changed the CRD), but if the CRDs are there when we bootstrap the cluster already, this would not only speed up the process but it would also open up new opportunities for us as it avoids race conditions between CRs and operators in bootstrap situations.

xh3b4sd:
There should be one source of truth for the CRD and its specification. When there are multiple authorities managing it it will just be painful and complicated to change it when we have to. At worst we deadlock the process and have to rewrite or disable the management in one place or the other. The design is bad. 

How long does it take to install `chart-operator` and have the CRD being installed? 1 second? 10 seconds? 100 seconds? One option would be to block the resource for 10 seconds. We do not want to do this but this is way better than doing something worse. Another, or maybe even additional option would be to lower the resync period. Another option would be to extend `operatorkit` to enable us to schedule another resync after a predefined time or condition. This could also be as dirty as having a goroutine updating something in the CR to trigger the event.

rossf7:
I can check how long chart-operator takes to install. Blocking might work and I guess is cleaner than having a resource for it.

I'm working on this for my cluster-operator PM in https://github.com/giantswarm/giantswarm/issues/5218. Let's get a solution for the problem there and then we can apply it in app-operator too.

@tomahawk28 Is that OK from your side?

tomahawk28:
@rossf7 Sure :)

rossf7:
@tomahawk28 This can be closed. We create the CRD by installing the operator. Sorry about the back and forth on this.

tomahawk28:
I'll close this PR since we agree to install operator instead.

PR #35, opened 2019-02-15
Use cc not ctlCtx

rossf7:
See https://github.com/giantswarm/fmt/pull/42. 

Democracy in action. You can't argue with PR #42.


rossf7:
@xh3b4sd I looked but I don't see `sc` anywhere. What am I missing?

xh3b4sd:
I have tomatoes on my eyes and thought this is the `aws-operator`.

rossf7:
> I have tomatoes on my eyes 

❤️ 

OK cool going with this! We're cooking with water now. 😉

xh3b4sd:
It goes more like, we all only cook with water after all. But yeah, definitely. :D

PR #37, opened 2019-02-15
Remove unused circle config

rossf7:
Currently CI is red on merge to master. Looks like some copy pasta from chart-operator.

Simplifying to just have what we use now. We still need to have app-operator installed in control planes with draughtsman. I plan to set that up next week.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,17 +61,3 @@ workflows:
       - e2eTestChartInstall:
           requires:
             - build
-
-      - deploy:
-          filters:
-            branches:
-              only: master
-          requires:
-            - e2eTestChartInstall
-
-      - publish_to_stable:
-          filters:
-            branches:
-              only: master
-          requires:
-            - deploy


.circleci/config.yml
```diff
@@ -61,17 +61,3 @@ workflows:
...
-              only: master
-          requires:
-            - e2eTestChartInstall
-
-      - publish_to_stable:
-          filters:
-            branches:
-              only: master
-          requires:
-            - deploy
```
rossf7: Publish to stable is for the release channels stuff in chart-operator. Not needed for app-operator.

.circleci/config.yml
```diff
@@ -61,17 +61,3 @@ workflows:
       - e2eTestChartInstall:
           requires:
             - build
-
-      - deploy:
-          filters:
-            branches:
-              only: master
-          requires:
-            - e2eTestChartInstall
```
rossf7: `deploy` currently errors because this is the workflow but there is no deploy job.

PR #38, opened 2019-02-19
Add appCatalog CR to controller context

rossf7:
This is prep for adding the desired state for the `configmap` resource. 

We need the appCatalog CR to calculate the desired state in multiple resources.  This PR adds an `appcatalog` package that gets it. Its used to add it to the `app` controller context in the `initCtxFunc`. 

This is safe because in the unlikely event the catalog name changes we will just process the update event.


service/controller/app/v1/appcatalog/appcatalog.go
```diff
@@ -0,0 +1,76 @@
+package appcatalog
```
rossf7: The appcatalog package is used to fetch the CR because its needed by chart, configmap and secret resources. It will be added to the context so we only call the K8s API once.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -22,16 +22,12 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
 	}
 
-	catalogName := key.CatalogName(cr)
-
-	appCatalog, err := r.g8sClient.ApplicationV1alpha1().AppCatalogs(r.watchNamespace).Get(catalogName, metav1.GetOptions{})
-	if apierrors.IsNotFound(err) {
-		return nil, microerror.Maskf(notFoundError, "appCatalog %#q in namespace %#q", catalogName, "default")
-	} else if err != nil {
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
```
rossf7: Now we just get the appCatalog CR from the context.

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -109,7 +109,7 @@ func Test_Resource_GetDesiredState(t *testing.T) {
 			},
 		},
 		{
-			name: "case 1: appcatalog not found",
+			name: "case 1: generating catalog url failed",
```
rossf7: Test gets simplified.

service/controller/app/v1/resource_set.go
```diff
@@ -158,6 +174,11 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 			return nil, microerror.Mask(err)
 		}
 
+		catalogCR, err := appCatalog.GetCatalogForApp(ctx, cr)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
```
rossf7: We get the catalog CR in the `initCtxFunc` and add it to the context.

rossf7:
@tomahawk28 A fresh controller context is created at the state of each reconciliation loop. So we will always get the latest appCatalog CR.

Its unlikely we will change the appCatalog for an app CR. But if we do then an update event will be generated and there will be a new loop to process.

service/controller/app/v1/appcatalog/appcatalog.go
xh3b4sd: That is nice, but when the catalog is a CR, why don't we watch the CR?

tomahawk28: I'm working on that another `index` resource which listening to appCatalog changes, maybe it would answer for yours?

rossf7: Yes like @tomahawk28 says the operator has 2 controllers with different purposes.

The appCatalog is cluster scoped and acts as a "reference CR" storing catalog level configuration. So to calculate the desired state we need both CRs.

xh3b4sd: I see. Thank you.

PR #40, opened 2019-02-20
Index.yaml: GetDesiredState

tomahawk28:
GetDesiredState for appCatalog/index resource: 
------
- It will fetch `index.yaml` from `appCatalog/Storage/URL` website, and store it to `configMap` which share same object name, namespace.
- yaml validation is out of scope.


pkg/label/label.go
```diff
@@ -13,3 +13,29 @@ const (
...
 	ManagedBy = "giantswarm.io/managed-by"
 )
+
+// ProcessLabels ensures the chart-operator.giantswarm.io/version label is
+// present and the app-operator.giantswarm.io/version label is removed. It
+// also ensures the giantswarm.io/managed-by label is accurate.
+//
+// Any other labels added to the app custom resource are passed on to the chart
+// custom resource.
+func ProcessLabels(projectName, chartCustomResourceVersion string, inputLabels map[string]string) map[string]string {
```
rossf7:
```suggestion
func ProcessLabels(inputLabels, requiredLabels, excludedLabels map[string]string) map[string]string {
```
This is cool but I'd like to see it made more generic so we can reuse in all resources. WDYT to using this method signature?

For example for the chart CRs we need to replace the app-operator version label with the chart-operator version. But we don't need to do that for the configmaps.

We should also have unit tests for this.

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -221,30 +222,31 @@ func Test_processLabels(t *testing.T) {
...
+			chartVersion: "1.0.0",
 			inputLabels: map[string]string{
-				"app":                                "prometheus",
-				"app-operator.giantswarm.io/version": "1.0.0",
-				"giantswarm.io/cluster":              "5xchu",
-				"giantswarm.io/managed-by":           "cluster-operator",
-				"giantswarm.io/organization":         "giantswarm",
+				"app":                        "prometheus",
+				"giantswarm.io/cluster":      "5xchu",
+				"giantswarm.io/managed-by":   "cluster-operator",
```
rossf7:
```suggestion
				"giantswarm.io/managed-by":   "app-operator",
```

service/controller/appcatalog/v1/resource/index/desired.go
```diff
@@ -0,0 +1,57 @@
+package index
+
+import (
```
rossf7: Imports need ordering.

service/controller/appcatalog/v1/resource/index/desired.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+	"io/ioutil"
+	"k8s.io/api/core/v1"
```
rossf7:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

service/controller/appcatalog/v1/resource/index/desired.go
```diff
@@ -0,0 +1,57 @@
...
+	"net/http"
+	"net/url"
+	"path"
+
+	"github.com/giantswarm/microerror"
+	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
rossf7:
```suggestion
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

service/controller/appcatalog/v1/resource/index/error.go
```diff
@@ -19,3 +19,12 @@ var wrongTypeError = &microerror.Error{
 func IsWrongType(err error) bool {
 	return microerror.Cause(err) == wrongTypeError
 }
+
+var indexNotFound = &microerror.Error{
```
rossf7:
```suggestion
var notFound = &microerror.Error{
```

We can use a more generic error here. This helps minimize the number of error types we need to deal with.

service/controller/appcatalog/v1/resource/index/resource.go
```diff
@@ -6,6 +6,7 @@ import (
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	"github.com/giantswarm/operatorkit/controller"
+	"k8s.io/api/core/v1"
```
rossf7:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

service/controller/appcatalog/v1/resource/index/resource.go
```diff
@@ -17,22 +18,22 @@ const (
 // Config represents the configuration used to create a new index resource.
 type Config struct {
 	// Dependencies.
-	K8sClient kubernetes.Interface
-	Logger    micrologger.Logger
+	K8sClient   kubernetes.Interface
+	Logger      micrologger.Logger
+	ProjectName string
```
rossf7: The settings comment is missing here.

service/controller/appcatalog/v1/resource/index/resource.go
```diff
@@ -65,11 +66,15 @@ func New(config Config) (*Resource, error) {
 	if config.Logger == nil {
 		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
 	}
+	if config.ProjectName == "" {
```
rossf7:
```suggestion

	if config.ProjectName == "" {
```

service/controller/appcatalog/v1/resource/index/resource.go
```diff
@@ -65,11 +66,15 @@ func New(config Config) (*Resource, error) {
...
+		return nil, microerror.Maskf(invalidConfigError, "%T.ProjectName must not be empty", config)
+	}
 
 	r := &Resource{
 		// Dependencies.
-		k8sClient: config.K8sClient,
-		logger:    config.Logger,
+		k8sClient:   config.K8sClient,
+		logger:      config.Logger,
+		projectName: config.ProjectName,
```
rossf7: Settings comment needed here too.

service/controller/appcatalog/v1/resource_set.go
```diff
@@ -46,8 +46,9 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 	var indexResource controller.Resource
 	{
 		c := index.Config{
-			K8sClient: config.K8sClient,
-			Logger:    config.Logger,
+			K8sClient:   config.K8sClient,
+			Logger:      config.Logger,
+			ProjectName: config.ProjectName,
```
rossf7:
```suggestion

			ProjectName: config.ProjectName,
```

service/controller/appcatalog/v1/resource/index/desired.go
```diff
@@ -0,0 +1,57 @@
...
+
+	defer response.Body.Close()
+	content, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	configMap := &v1.ConfigMap{
+		ObjectMeta: v12.ObjectMeta{
+			Name:        cr.Name,
```
rossf7: I'm not sure just using the CR name makes sense. What about having a `-index` suffix?

service/controller/appcatalog/v1/resource/index/desired.go
```diff
@@ -0,0 +1,57 @@
...
+			Namespace:   cr.Namespace,
```
rossf7: I'm not sure this is going to work! As the appCatalog CR is cluster scoped not namespace scoped. 

I think it makes sense to store it in the `giantswarm` namespace because that is where our API runs. @puja108 WDYT?

service/controller/appcatalog/v1/resource/index/desired_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"reflect"
+	"testing"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/micrologger/microloggertest"
+	"github.com/google/go-cmp/cmp"
+	"k8s.io/api/core/v1"
```
rossf7:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

Another alias to set. Since we use the k8s packages so often using the naming convention is important.

service/controller/appcatalog/v1/resource/index/desired.go
```diff
@@ -0,0 +1,57 @@
...
+			Labels:      label.ProcessLabels(r.projectName, "", cr.ObjectMeta.Labels),
```
puja108: do we really replicate all labels of the CR to the CM? Could there be some that are not right here, like managed-by or sth?

service/controller/appcatalog/v1/resource/index/desired.go
```diff
@@ -0,0 +1,57 @@
...
+			Annotations: cr.ObjectMeta.Annotations,
```
puja108: same here, do we really want to duplicate everything from CR?

service/controller/appcatalog/v1/resource/index/desired.go
rossf7: The purpose of `ProcessLabels` is to get this right. There are some like the `managed-by` label that we need to update.

There are others like `giantswarm.io/cluster` and `giantswarm.io/organization` that are useful to pass on. Then we can just set them once in cluster-operator or wherever. 

But yes we should try and keep this simple.

tomahawk28:
@rossf7 I think it's nice to merge both `key` package in the same place. Shall I continue to do that?

pkg/label/label.go
tomahawk28: Good idea.

pkg/label/label.go
```diff
@@ -13,3 +13,29 @@ const (
 	// ManagedBy is set for Kubernetes resources managed by the operator.
 	ManagedBy = "giantswarm.io/managed-by"
 )
+
+// ProcessLabels ensures the chart-operator.giantswarm.io/version label is
+// present and the app-operator.giantswarm.io/version label is removed. It
+// also ensures the giantswarm.io/managed-by label is accurate.
```
xh3b4sd: This comment is pretty nice, and it shows that the method does too much IMO. The primitives here are defaulting and validation. The code below is simple in this very use case but we should try to separate and isolate implementations so that they are maintainable and testable. When I test two conceptually different functionalities and change one implementation, breaking two tests shows that the design is wrong. A little bit of docs and implementations can be found in various places. 

* https://github.com/giantswarm/giantswarm/blob/master/sigs/operator/command_line_tools.md
* https://github.com/giantswarm/api/tree/master/service/cluster/creator/defaulting
* https://github.com/giantswarm/api/tree/master/service/cluster/creator/validation
* https://github.com/giantswarm/api/tree/master/service/cluster/updater/defaulting
* https://github.com/giantswarm/api/tree/master/service/cluster/updater/validation

service/controller/appcatalog/v1/resource/index/error.go
xh3b4sd: Note that `Kind` and variable name should be equal. The matcher name should also be aligned. Otherwise it will be a complete mess.

rossf7:
@tomahawk28 Lets do the moving around after we've completed the `index` and `configmap` resources. Otherwise it can cause some disruption, conflicts etc.

We should do the moving as a separate PR so its easy to review.

pkg/label/label.go
rossf7: The current comment is very detailed and only makes for the chart resource. I agree the logic should be generic and basing it on our existing implementations is a good approach.

tomahawk28: I agree I'll make it smaller then.

service/controller/app/v1/resource/chart/desired_test.go
tomahawk28: I'm moving it to new file `lable_test.go`

service/controller/appcatalog/v1/resource/index/desired.go
tomahawk28: I'm not sure either, where we should keep it?

rossf7: We should use the `giantswarm` namespace but lets add a flag for it like with the watch namespace.

`f.Service.AppCatalog.IndexNamespace` should do the job.

tomahawk28: We already have `f.Service.Kubernetes.Watch.Namespace` which is already using inside `app` controller, make it sense to use it here too?

rossf7: I don't think that makes sense because they have different purposes.

The watch namespace is where we check for app CRs. This namespace is where we save the index so it can be used by the API.

service/controller/appcatalog/v1/resource/index/desired.go
puja108: I'm also currently thinking if the whole label stuff would not be a case for OPA policies, but ok for now

rossf7:
@tomahawk28 @puja108 OK taking a step back on this. 

This PR is about the desired state for the index configmap. This is just for our cache of the index.yaml. So we don't need to pass on the labels on like we do for the chart and configmap resources.

I checked our standard labels and we should just set the `giantswarm.io/managed-by` label with no annotations and we're done.

https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md

It would still be good to have a generic `ProcessLabels` method for the chart and configmap resource but lets do that in a separate PR.

Sorry I missed that in my earlier review because I was thinking about the configmap resource. WDYT?

puja108:
agree, as for the `ProcessLabels` method, maybe we create an issue and take it to SIG operator to talk about it. I'd also like to evaluate if we actually need to do so much defaulting and validation in operators or services and api, or if we could move that all to a more declarative way with OPA like I am testing with giantswarm/OPA-PoC

rossf7:
Yes lets get this PR sorted and decide what to do about `ProcessLabels` separately.

I have a concern about the OPA approach. I think its ideal for customers creating app CRs.

For operators I have a worry that something else is updating the metadata but we're using those labels to control the versioning. That could get painful when debugging.

tomahawk28:
Sure, But I don't want to git revert it now :( That's a pain

rossf7:
> Sure, But I don't want to git revert it now

Why? Just delete the code that isn't needed and you can get it back from the git history later.

It is a pain but sometimes this happens during the review process.

tomahawk28:
It's not just deleting, but some further changes after that also stack onto it, make it more difficult to disintegrate. Moreover, I remember that we have a decision to move out of this function from the beginning. Anyway I will remove it for now

flag/service/service.go
```diff
@@ -4,7 +4,12 @@ import (
 	"github.com/giantswarm/app-operator/flag/service/kubernetes"
 )
 
+type AppCatalog struct {
```
rossf7: We should have godoc here.

flag/service/service.go
```diff
@@ -4,7 +4,12 @@ import (
...
+	IndexNamespace string
+}
+
 // Service is an intermediate data structure for command line configuration flags.
 type Service struct {
 	Kubernetes kubernetes.Kubernetes
+	AppCatalog AppCatalog
```
rossf7: Fields should be ordered.

main.go
```diff
@@ -111,6 +111,7 @@ func mainWithError() (err error) {
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.CrtFile, "", "Certificate file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.KeyFile, "", "Key file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.Watch.Namespace, "default", "The namespace where appcatalog and app CRs are located.")
+	daemonCommand.PersistentFlags().String(f.Service.AppCatalog.IndexNamespace, "giantswarm", "The namespace where operator keep index file as configMap.")
```
rossf7:
```suggestion
	daemonCommand.PersistentFlags().String(f.Service.AppCatalog.IndexNamespace, "giantswarm", "The namespace where the catalog index.yaml is stored in a configmap.")
```

rossf7: Flags also need ordering.

service/controller/appcatalog/v1/resource/index/desired.go
rossf7: Lets removing setting the annotations that isn't needed.

The `label.ManagedBy` should be on a new line.

service/controller/appcatalog/v1/resource/index/desired_test.go
```diff
@@ -0,0 +1,137 @@
...
+		expectedConfigMap *corev1.ConfigMap
+		h                 func(w http.ResponseWriter, r *http.Request)
+		errorMatcher      func(error) bool
+	}{
+		{
+			name: "case 0: flawless flow",
+			obj: &v1alpha1.AppCatalog{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "giantswarm",
+					Namespace: "default",
```
rossf7: Lets remove the namespace and labels from the test. The CR is cluster scoped so setting a namespace doesn't make sense.

service/controller/appcatalog/v1/resource/index/error.go
```diff
@@ -19,3 +19,12 @@ var wrongTypeError = &microerror.Error{
 func IsWrongType(err error) bool {
 	return microerror.Cause(err) == wrongTypeError
 }
+
+var notFound = &microerror.Error{
```
rossf7:
```suggestion
var notFoundError = &microerror.Error{
```

service/controller/appcatalog/v1/resource_set.go
```diff
@@ -23,6 +23,7 @@ type ResourceSetConfig struct {
 
 	// Settings.
 	HandledVersionBundles []string
+	IndexNamespace        string
```
rossf7: Not related to this PR but the `HandledVersionBundles` above should be deleted.

tomahawk28:
@rossf7 PTAL :)

service/controller/appcatalog/v1/resource/index/desired_test.go
```diff
@@ -0,0 +1,133 @@
...
+			defer ts.Close()
+
+			tc.obj.Spec.Storage.URL = ts.URL
+
+			c := Config{
+				K8sClient: k8sfake.NewSimpleClientset(),
+				Logger:    microloggertest.New(),
+
+				ProjectName:    "app-operator",
+				IndexNamespace: "giantswarm",
```
rossf7: Minor but would be good to order settings here too.

service/controller/appcatalog/v1/resource/index/desired_test.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+
+			if err == nil && tc.errorMatcher == nil {
+				configMap, err := toConfigMap(result)
+				if err != nil {
+					t.Fatalf("error == %#v, want nil", err)
+				}
+
+				if !reflect.DeepEqual(configMap.ObjectMeta, tc.expectedConfigMap.ObjectMeta) {
+					t.Fatalf("want matching objectmeta \n %s", cmp.Diff(configMap.ObjectMeta, tc.expectedConfigMap.ObjectMeta))
```
rossf7: Nice using `cmp` here is great.

service/controller/appcatalog/v1/resource/index/resource.go
```diff
@@ -19,20 +20,24 @@ type Config struct {
 	// Dependencies.
 	K8sClient kubernetes.Interface
 	Logger    micrologger.Logger
+
+	// Settings.
```
rossf7: The settings need to be ordered in the resource. Its nitpicky but the code gets untidy quickly when stuff isn't ordered.

tomahawk28: Sure, will keep in mind.

flag/service/service.go
```diff
@@ -4,7 +4,13 @@ import (
 	"github.com/giantswarm/app-operator/flag/service/kubernetes"
 )
 
+// AppCatalog is a data structure to hold AppCatalog specific command line configuration flags.
+type AppCatalog struct {
```
xh3b4sd: When you look at how the flags and the folders to define them are structured, you notice that the structure here should be somewhere else. This is actually pretty easy to understand as there is already the `Kubernetes` structure as well. Let's not reinvent the wheel here.

tomahawk28: Sure, I just move it out to another file.

service/controller/appcatalog/v1/resource/index/resource.go
```diff
@@ -19,20 +21,24 @@ type Config struct {
...
 type Resource struct {
 	// Dependencies.
 	k8sClient kubernetes.Interface
 	logger    micrologger.Logger
-}
 
-func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
-	return nil, nil
+	// Settings.
+	projectName    string
```
rossf7: @tomahawk28 The ordering still isn't right in multiple places like here. Please can you do a follow up PR so this gets sorted.

PR #42, opened 2019-02-22
Index GetCurrentState

tomahawk28:
- GetCurrentState for index resource

### DIFF
--- a/service/controller/appcatalog/v1/key/key.go
+++ b/service/controller/appcatalog/v1/key/key.go
@@ -2,9 +2,9 @@ package key
 
 import (
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
-	"github.com/giantswarm/microerror"
-
 	"github.com/giantswarm/app-operator/pkg/label"
+	"github.com/giantswarm/microerror"
+	corev1 "k8s.io/api/core/v1"
 )
 
 func AppCatalogTitle(customResource v1alpha1.AppCatalog) string {
@@ -33,6 +33,21 @@ func ToCustomResource(v interface{}) (v1alpha1.AppCatalog, error) {
 
 	if customResource == nil {
 		return v1alpha1.AppCatalog{}, microerror.Maskf(emptyValueError, "empty value cannot be converted to CustomObject")
+	}
+
+	return *customResource, nil
+}
+
+// ToConfigMap converts value to corev1.ConfigMap and returns it or error
+// if type does not match.
+func ToConfigMap(v interface{}) (corev1.ConfigMap, error) {
+	customResource, ok := v.(*corev1.ConfigMap)
+	if !ok {
+		return corev1.ConfigMap{}, microerror.Maskf(wrongTypeError, "expected '%T', got '%T'", &corev1.ConfigMap{}, v)
+	}
+
+	if customResource == nil {
+		return corev1.ConfigMap{}, microerror.Maskf(emptyValueError, "empty value cannot be converted to CustomObject")
 	}
 
 	return *customResource, nil
--- a/service/controller/appcatalog/v1/key/key.go
+++ b/service/controller/appcatalog/v1/key/key.go
@@ -4,7 +4,6 @@ import (
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/app-operator/pkg/label"
 	"github.com/giantswarm/microerror"
-	corev1 "k8s.io/api/core/v1"
 )
 
 func AppCatalogTitle(customResource v1alpha1.AppCatalog) string {
@@ -33,21 +32,6 @@ func ToCustomResource(v interface{}) (v1alpha1.AppCatalog, error) {
 
 	if customResource == nil {
 		return v1alpha1.AppCatalog{}, microerror.Maskf(emptyValueError, "empty value cannot be converted to CustomObject")
-	}
-
-	return *customResource, nil
-}
-
-// ToConfigMap converts value to corev1.ConfigMap and returns it or error
-// if type does not match.
-func ToConfigMap(v interface{}) (corev1.ConfigMap, error) {
-	customResource, ok := v.(*corev1.ConfigMap)
-	if !ok {
-		return corev1.ConfigMap{}, microerror.Maskf(wrongTypeError, "expected '%T', got '%T'", &corev1.ConfigMap{}, v)
-	}
-
-	if customResource == nil {
-		return corev1.ConfigMap{}, microerror.Maskf(emptyValueError, "empty value cannot be converted to CustomObject")
 	}
 
 	return *customResource, nil
--- a/service/controller/appcatalog/v1/key/key.go
+++ b/service/controller/appcatalog/v1/key/key.go
@@ -2,8 +2,9 @@ package key
 
 import (
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
-	"github.com/giantswarm/app-operator/pkg/label"
 	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/app-operator/pkg/label"
 )
 
 func AppCatalogTitle(customResource v1alpha1.AppCatalog) string {
--- a/service/controller/appcatalog/v1/resource/index/current.go
+++ b/service/controller/appcatalog/v1/resource/index/current.go
@@ -1,0 +1,33 @@
+package index
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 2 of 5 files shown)


service/controller/appcatalog/v1/key/key.go
```diff
@@ -2,9 +2,8 @@ package key
 
 import (
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
-	"github.com/giantswarm/microerror"
-
 	"github.com/giantswarm/app-operator/pkg/label"
+	"github.com/giantswarm/microerror"
```
TheoBrigitte: The way imports where split was right before.

service/service.go
```diff
@@ -98,9 +98,9 @@ func New(config Config) (*Service, error) {
 			K8sClient:    k8sClient,
 			K8sExtClient: k8sExtClient,
 
+			IndexNamespace: config.Viper.GetString(config.Flag.Service.AppCatalog.IndexNamespace),
 			ProjectName:    config.ProjectName,
 			WatchNamespace: config.Viper.GetString(config.Flag.Service.Kubernetes.Watch.Namespace),
-			IndexNamespace: config.Viper.GetString(config.Flag.Service.AppCatalog.IndexNamespace),
```
TheoBrigitte: Alphabetical sorting is nice, and it was right before.

service/controller/appcatalog/v1/resource/index/current.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	name := configMapName(cr.Name)
+
```
TheoBrigitte: Missing a log line here
```suggestion
r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding index configMap %#q", name))
```

service/controller/appcatalog/v1/resource/index/current_test.go
```diff
@@ -0,0 +1,125 @@
...
+			switch {
+			case err != nil && tc.errorMatcher == nil:
+				t.Fatalf("error == %#v, want nil", err)
+			case err == nil && tc.errorMatcher != nil:
+				t.Fatalf("error == nil, want non-nil")
+			case err != nil && !tc.errorMatcher(err):
+				t.Fatalf("error == %#v, want matching", err)
+			}
+
+			if err == nil && tc.errorMatcher == nil {
```
TheoBrigitte: IMHO I think this condition is not needed, it is obvious that we run into this case if the test didn't failed at this point.

service/controller/appcatalog/v1/resource/index/current_test.go
```diff
@@ -0,0 +1,125 @@
...
+				if result != nil {
```
TheoBrigitte: What happen in case result is nil ? Isn't this an error ?

service/controller/appcatalog/v1/resource/index/current_test.go
```diff
@@ -0,0 +1,125 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func Test_Resource_GetCurrentState(t *testing.T) {
+	tests := []struct {
+		name              string
+		obj               *v1alpha1.AppCatalog
```
TheoBrigitte: It would be good to have `obj` as `interface{}` so we can also test for cases like `obj` is not an AppCatalog.

service/service.go
tomahawk28: I think the current status is correct since it follows the order of `Config`?

service/controller/appcatalog/v1/resource/index/current.go
tomahawk28: good catch

service/controller/appcatalog/v1/resource/index/current_test.go
tomahawk28: No, In this case, we regard it like that we don't have related `configMap`.

service/controller/appcatalog/v1/resource/index/current_test.go
```diff
@@ -0,0 +1,123 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func Test_Resource_GetCurrentState(t *testing.T) {
+	tests := []struct {
+		name              string
+		obj               interface{}
+		returnedConfigMap *corev1.ConfigMap
```
rossf7:
```suggestion
		expectedConfigMap *corev1.ConfigMap
```

I'd prefer using expected as this is the expected state we're testing.

service/controller/appcatalog/v1/resource/index/current_test.go
```diff
@@ -0,0 +1,123 @@
...
+			switch {
+			case err != nil && tc.errorMatcher == nil:
+				t.Fatalf("error == %#v, want nil", err)
+			case err == nil && tc.errorMatcher != nil:
+				t.Fatalf("error == nil, want non-nil")
+			case err != nil && !tc.errorMatcher(err):
+				t.Fatalf("error == %#v, want matching", err)
+			}
+
+			if result != nil {
```
rossf7: I think if nil is expected we should put nil in the test case.  The test then just needs to check if the result matches the test case.

service/controller/appcatalog/v1/resource/index/resource.go
```diff
@@ -34,12 +34,8 @@ type Resource struct {
 	logger    micrologger.Logger
 
 	// Settings.
-	projectName    string
 	indexNamespace string
-}
-
-func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
-	return nil, nil
+	projectName    string
```
rossf7: <3

service/service.go
rossf7: Now the ordering looks OK.

TheoBrigitte: All good

service/controller/appcatalog/v1/resource/index/current_test.go
TheoBrigitte: Looks good now.

service/controller/appcatalog/v1/resource/index/current_test.go
TheoBrigitte: It would be nice to add a test case where obj is an invalid value.

service/controller/appcatalog/v1/resource/index/current_test.go
```diff
@@ -0,0 +1,127 @@
...
+			switch {
+			case err != nil && tc.errorMatcher == nil:
+				t.Fatalf("error == %#v, want nil", err)
+			case err == nil && tc.errorMatcher != nil:
+				t.Fatalf("error == nil, want non-nil")
+			case err != nil && !tc.errorMatcher(err):
+				t.Fatalf("error == %#v, want matching", err)
+			}
+
+			if tc.expectedConfigMap == nil {
```
rossf7: Nitpick but I think the conditional logic could be simplified.

```go
if tc.expectedConfigMap == nil && result != nil {
   t.Fatalf("expected nil ConfigMap got %#v", result)
}
```

tomahawk28: 👍 Thanks

tomahawk28: When both `result`, `expectedConfigMap` is nil, it will run all the code after these, then it would have null pointer reference problem eventually. So I think `else` statement like this is needed.

rossf7: I would just add another test case for the reverse. Style wise I prefer it if the nil handling is checked first and then the expected state is checked. IMO the code is harder to read with the else statements.


```
if tc.expectedConfigMap != nil && result == nil {
   t.Fatal("expected non-nil ConfigMap got nil")
}
```

tomahawk28: It's not just reverse checking is needed, if `result` is `nil` then we should not convert it to `ConfigMap`. I'm not sure how to avoid it just using if statement like above.

rossf7: The test is using `t.Fatalf` rather than `t.Errorf`. So the test will fail and stop executing if the result is nil when it shouldn't be.

So you just need to call `toConfigMap` after the nil check.

tomahawk28: So we decide to move on as it is, we can fix that on later with more debugging.

PR #43, opened 2019-02-22
Updated unit test

rossf7:
Follow up from https://github.com/giantswarm/app-operator/pull/42.

### DIFF
--- a/service/controller/appcatalog/v1/resource/index/current_test.go
+++ b/service/controller/appcatalog/v1/resource/index/current_test.go
@@ -115,19 +115,20 @@ func Test_Resource_GetCurrentState(t *testing.T) {
 				t.Fatalf("error == %#v, want matching", err)
 			}
 
-			if tc.expectedConfigMap == nil {
-				if result != nil {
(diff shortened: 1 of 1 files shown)


service/controller/appcatalog/v1/resource/index/current_test.go
```diff
@@ -115,19 +115,20 @@ func Test_Resource_GetCurrentState(t *testing.T) {
...
+
+			cm, err := toConfigMap(result)
+			if err != nil {
+				t.Fatalf("error == %#v, want nil", err)
+			}
 
-				if !reflect.DeepEqual(*cm, *tc.expectedConfigMap) {
-					t.Fatalf("ConfigMap == %#v, want %#v", cm, tc.expectedConfigMap)
-				}
+			if tc.expectedConfigMap != nil && !reflect.DeepEqual(cm, tc.expectedConfigMap) {
```
rossf7: The problem was the comparison for case 1 the expected CM is nil. Also reflect.DeepEqual will compare pointers so you don't need the `*`.

PR #44, opened 2019-02-22
Create, update function in index resource

tomahawk28:
- Most of them are copy from `app/configmap` resource. 
- We will have a talk about how to keep it as generic resource duplicate parts next week. (25/Feb~)

### DIFF
--- a/service/controller/appcatalog/v1/resource/index/create.go
+++ b/service/controller/appcatalog/v1/resource/index/create.go
@@ -42,7 +42,7 @@ func (r *Resource) newCreateChange(ctx context.Context, currentResource, desired
 
 	createConfigMap := &corev1.ConfigMap{}
 
-	if isEmpty(currentConfigMap) {
+	if currentConfigMap == nil || isEmpty(currentConfigMap) {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the %#q configmap needs to be created", desiredConfigMap.Name))
 		createConfigMap = desiredConfigMap
 	} else {
--- a/service/controller/appcatalog/v1/resource/index/create.go
+++ b/service/controller/appcatalog/v1/resource/index/create.go
@@ -9,20 +9,20 @@ import (
 )
 
 func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
-	cm, err := toConfigMap(createChange)
+	configMap, err := toConfigMap(createChange)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
-	if cm.Name != "" {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensuring creation of index configMap %#q", cm.Name))
+	if configMap.Name != "" {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensuring creation of index configMap %#q", configMap.Name))
 
-		_, err = r.k8sClient.CoreV1().ConfigMaps(r.indexNamespace).Create(cm)
+		_, err = r.k8sClient.CoreV1().ConfigMaps(r.indexNamespace).Create(configMap)
 		if err != nil {
 			return microerror.Mask(err)
 		}
 
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensured creation of index configMap %#q", cm.Name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensured creation of index configMap %#q", configMap.Name))
 	} else {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("no need to create index configmap"))
 	}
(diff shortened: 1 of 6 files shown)


service/controller/appcatalog/v1/resource/index/create.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/giantswarm/microerror"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
+	cm, err := toConfigMap(createChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	if cm.Name != "" {
```
xh3b4sd: Newlines are free. Let's use them to make reading code a bit easier. 

```suggestion

	if cm.Name != "" {
```

service/controller/appcatalog/v1/resource/index/create.go
```diff
@@ -0,0 +1,53 @@
...
+		_, err = r.k8sClient.CoreV1().ConfigMaps(r.indexNamespace).Create(cm)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensured creation of index configMap %#q", cm.Name))
+	} else {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("no need to create index configmap"))
+	}
+	return nil
```
xh3b4sd:
```suggestion

	return nil
```

service/controller/appcatalog/v1/resource/index/update.go
```diff
@@ -0,0 +1,79 @@
...
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/controller"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange interface{}) error {
+	configMap, err := toConfigMap(updateChange)
```
xh3b4sd: You called this `cm` in other places. Let's stick to one approach.

service/controller/appcatalog/v1/resource/index/update.go
```diff
@@ -0,0 +1,79 @@
...
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding out if the index configmap has to be updated")
+
+	updateConfigMap := &corev1.ConfigMap{}
+	isModified := !isEmpty(currentConfigMap) && !equals(currentConfigMap, desiredConfigMap)
+	if isModified {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "the index configmap has to be updated")
+
+		updateConfigMap = desiredConfigMap.DeepCopy()
+		updateConfigMap.ObjectMeta.ResourceVersion = currentConfigMap.ObjectMeta.ResourceVersion
```
xh3b4sd: I think we should fetch the fresh object here again to have the latest resource version. When the update fails because the object got updated from someone else meanwhile, this here will fail and the retry resource will retry for nothing because the same outdated resource version will still be used. Fetching the resource version here is not really cool as we tried to keep `newCreateChange` and `newUpdateChange` free from network dependencies, but we need to fetch the resource version here to be safe. I just fixed an issue we had in the `kvm-operator`. See https://github.com/giantswarm/kvm-operator/pull/653.

service/controller/appcatalog/v1/resource/index/update.go
tomahawk28: Thanks, fixed it.

service/controller/appcatalog/v1/resource/index/create.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
+	cm, err := toConfigMap(createChange)
```
rossf7:
```suggestion
	configMap, err := toConfigMap(createChange)
```

I'd prefer it if we used configMap rather than cm. This then aligns with using currentConfigMap etc in newCreateChange.

service/controller/appcatalog/v1/resource/index/update.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/controller"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange interface{}) error {
+	cm, err := toConfigMap(updateChange)
```
rossf7: Same here.

tomahawk28:
@xh3b4sd PTAL :)

tomahawk28:
Looks like Tim is afk today; will merge now

xh3b4sd:
I am not AFK today. Hello! 👋

tomahawk28:
Oo oh, sorry

PR #46, opened 2019-02-25
add delete function

tomahawk28:
- Add delete function inside index resource

### DIFF
--- a/service/controller/appcatalog/v1/resource/index/delete.go
+++ b/service/controller/appcatalog/v1/resource/index/delete.go
@@ -59,11 +59,11 @@ func (r *Resource) newDeleteChange(ctx context.Context, obj, currentState, desir
 
 	isModified := !isEmpty(currentConfigMap) && equals(currentConfigMap, desiredConfigMap)
 	if isModified {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the %#q index configMap needs to be deleted", desiredConfigMap.Name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the %#q configMap needs to be deleted", desiredConfigMap.Name))
 
 		return desiredConfigMap, nil
 	} else {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the %#q index configMap does not need to be deleted", desiredConfigMap.Name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the %#q configMap does not need to be deleted", desiredConfigMap.Name))
 	}
 
 	return nil, nil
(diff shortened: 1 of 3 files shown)


service/controller/appcatalog/v1/resource/index/delete.go
```diff
@@ -0,0 +1,71 @@
...
+		return nil, microerror.Mask(err)
+	}
+
+	patch := controller.NewPatch()
+	patch.SetDeleteChange(delete)
+
+	return patch, nil
+}
+
+func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange interface{}) error {
```
xh3b4sd: We try to order the methods alphabetically, where public comes before private methods.

service/controller/appcatalog/v1/resource/index/resource.go
```diff
@@ -39,14 +37,6 @@ type Resource struct {
...
-func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desiredState interface{}) (*controller.Patch, error) {
-	return nil, nil
-}
-
-func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange interface{}) error {
-	return nil
-}
-
 // New creates a new configured index resource.
 func New(config Config) (*Resource, error) {
```
xh3b4sd: I was thinking to improve the docs here a bit. Do you think it makes sense to have descriptions and pictures about the purpose and functionality of the resource implementations? Imagine you just join the company and check the operators. It would be cool to provide everyone with more information on **how** our software works.

tomahawk28: Definitely yes, pictures are much needed to understand how operatorkit is working. maybe I could create related issue?

rossf7: Yes the docs situation needs to improve. I was thinking we could have architecture docs for each operator in https://github.com/giantswarm/architecture.

We can then use the resource descriptions from the doc in the operator resources as godoc. WDYT? 

@tomahawk28 Could you start creating an app-operator doc?

I'd like to use the same structure as for cluster-operator. See https://github.com/giantswarm/giantswarm/pull/5405.

service/controller/appcatalog/v1/resource/index/delete.go
```diff
@@ -0,0 +1,70 @@
...
+	currentConfigMap, err := toConfigMap(currentState)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	desiredConfigMap, err := toConfigMap(desiredState)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding out if the %#q configMap has to be deleted", desiredConfigMap.Name))
```
rossf7: The log messages are not consistent. The other 2 mention index configmap this doesn't. I think just using configmap is fine as the resource is called index.

rossf7:
```suggestion
	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding out if the %#q configmap has to be deleted", desiredConfigMap.Name))
```

I missed it before but lets use lowercase for configmap and then all the log messages are aligned.

PR #47, opened 2019-02-25
Use in-cluster

tomahawk28:
- Use inCluser boolean variable from `app` CR


service/controller/app/v1/resource_set.go
```diff
@@ -159,7 +159,10 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		var k8sClient kubernetes.Interface
 		var g8sClient versioned.Interface
 
-		if cr.Spec.KubeConfig.Secret.Name != "" {
+		if cr.Spec.KubeConfig.InCluster {
```
rossf7: Lets have a key function and test rather than accessing the spec directly.

service/controller/app/v1/key/key.go
```diff
@@ -31,6 +31,10 @@ func ConfigMapNamespace(customResource v1alpha1.App) string {
 	return customResource.Spec.Config.ConfigMap.Namespace
 }
 
+func InCluser(customResource v1alpha1.App) bool {
```
```suggestion
func InCluster(customResource v1alpha1.App) bool {
```

PR #51, opened 2019-02-28
Utilise Resource configmap

tomahawk28:
- `app` CR controller is out of scope, it should create `configmap` inside tenant cluster directly, so we should careful


service/controller/appcatalog/v1/resource_set.go
```diff
@@ -56,7 +57,20 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
+		stateGetter, err := index.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		configOps := configmap.Config{
+			K8sClient: config.K8sClient,
+			Logger:    config.Logger,
+
+			Name:        config.ProjectName,
```
kopiczko: The name here should be "configmap". This is what `resource.Name()` returns.

service/controller/app/v1/resource_set.go
```diff
@@ -155,11 +155,13 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 			return nil, microerror.Mask(err)
 		}
 
-		// TODO: Remove if statement after in-cluster flag is implemented.
 		var k8sClient kubernetes.Interface
 		var g8sClient versioned.Interface
 
-		if cr.Spec.KubeConfig.Secret.Name != "" {
+		if key.InCluster(cr) {
```
kopiczko: Can `InCluster` related changes land in a separate PR?

service/controller/appcatalog/v1/resource_set.go
tomahawk28: I will make it constant

service/controller/app/v1/resource_set.go
tomahawk28: No, Since it's already in master after your initial review, it's gone now.

service/controller/appcatalog/v1/resource_set.go
kopiczko: Actually you can do that:

```suggestion
			Name:        index.Name,
```

service/controller/app/v1/resource_set.go
kopiczko: Sorry I don't follow. You mean it will disappear after merging master branch?

tomahawk28: My bad; I just rebase it with master and it's gone :)

service/controller/appcatalog/v1/resource_set.go
tomahawk28: is it okay to have `indexv1`?

service/controller/appcatalog/v1/resource/index/current_test.go
```diff
@@ -122,13 +122,15 @@ func Test_Resource_GetCurrentState(t *testing.T) {
 				t.Fatal("expected non-nil ConfigMap got nil")
 			}
 
-			cm, err := toConfigMap(result)
-			if err != nil {
-				t.Fatalf("error == %#v, want nil", err)
-			}
+			if result != nil {
```
tomahawk28: if not check this condition, then we are falling into memory error.

kopiczko:
Nice. LGTM. Let's test that once more, merge operatorkit changes, vendor it again and merge. Can you open that PR to public?

service/controller/appcatalog/v1/resource_set.go
kopiczko: Yes it's fine.

service/controller/appcatalog/v1/resource/index/current_test.go
kopiczko: I think we should check if `len(result) == 1`.

tomahawk28: is it okay when `result` is nil?

kopiczko: Yes.

kopiczko: https://play.golang.org/p/X9R3OmCAxcm

PR #52, opened 2019-03-05
Add desired state for configmap resource

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/4848

Completes the configmap resource by adding the desired state.

### DIFF
--- a/service/controller/app/v1/resource/configmap/create.go
+++ b/service/controller/app/v1/resource/configmap/create.go
@@ -16,7 +16,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		return microerror.Mask(err)
 	}
 
-	if configMap.Name != "" {
+	if configMap != nil {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("ensuring creation of configmap %#q", configMap.Name))
 
 		cc, err := controllercontext.FromContext(ctx)
--- a/service/controller/app/v1/resource/configmap/create.go
+++ b/service/controller/app/v1/resource/configmap/create.go
@@ -47,6 +47,10 @@ func (r *Resource) newCreateChange(ctx context.Context, currentResource, desired
 		return nil, microerror.Mask(err)
 	}
 
+	if desiredConfigMap == nil || desiredConfigMap.Name == "" {
+		return nil, nil
+	}
+
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding out if the %#q configmap has to be created", desiredConfigMap.Name))
 
 	createConfigMap := &corev1.ConfigMap{}
--- a/service/controller/app/v1/resource/configmap/create.go
+++ b/service/controller/app/v1/resource/configmap/create.go
@@ -47,19 +47,15 @@ func (r *Resource) newCreateChange(ctx context.Context, currentResource, desired
 		return nil, microerror.Mask(err)
 	}
 
-	if desiredConfigMap == nil || desiredConfigMap.Name == "" {
-		return nil, nil
-	}
-
-	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding out if the %#q configmap has to be created", desiredConfigMap.Name))
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding out if the configmap has to be created")
 
 	createConfigMap := &corev1.ConfigMap{}
 
 	if isEmpty(currentConfigMap) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the %#q configmap needs to be created", desiredConfigMap.Name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", "the configmap needs to be created")
 		createConfigMap = desiredConfigMap
 	} else {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the %#q configmap does not need to be created", desiredConfigMap.Name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", "the configmap does not need to be created")
 	}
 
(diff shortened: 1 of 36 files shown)


service/controller/app/v1/resource/configmap/create.go
```diff
@@ -16,7 +16,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		return microerror.Mask(err)
 	}
 
-	if configMap.Name != "" {
+	if !isEmpty(configMap) {
```
rossf7: isEmpty handles if the configmap is nil or empty.

service/controller/app/v1/resource/configmap/create.go
```diff
@@ -47,15 +47,15 @@ func (r *Resource) newCreateChange(ctx context.Context, currentResource, desired
 		return nil, microerror.Mask(err)
 	}
 
-	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding out if the %#q configmap has to be created", desiredConfigMap.Name))
+	r.logger.LogCtx(ctx, "level", "debug", "message", "finding out if the configmap has to be created")
```
rossf7: Simplifies the logging. The configmap will be nil if no config was specified in the app CR.

service/controller/app/v1/resource/configmap/desired.go
```diff
@@ -0,0 +1,106 @@
...
+	catalogConfigMapName := appcatalogkey.ConfigMapName(cc.AppCatalog)
+
+	if appConfigMapName == "" && catalogConfigMapName == "" {
+		// Return early as there is no config.
+		return nil, nil
+	}
+
+	if appConfigMapName != "" && catalogConfigMapName != "" {
+		return nil, microerror.Maskf(executionFailedError, "merging app and catalog configmaps is not yet supported")
+	}
```
rossf7: Merge support will be added later as we don't need it for VOO.

service/controller/app/v1/resource/configmap/desired.go
```diff
@@ -0,0 +1,106 @@
...
+
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	appConfigMapName := key.AppConfigMapName(cr)
+	catalogConfigMapName := appcatalogkey.ConfigMapName(cc.AppCatalog)
+
+	if appConfigMapName != "" && catalogConfigMapName == "" {
```
tomahawk28: I would suggest creating another function name `getConfigMapDataFromCatalog` and `getConfigMapDataFromApp` rather than checking condition as above.

rossf7: Yes that's nicer thanks. I'll get that implemented.

rossf7: Done. @tomahawk28 PTAL

kopiczko:
For this we could use the resource from operatorkit. Note the nuance in desired state if you are going to do so.

tomahawk28:
I remember that we talk about it before, this specific configMap resource need to install CRs inside tenant cluster, using kubeconfig library, doesn't it?

kopiczko:
> I remember that we talk about it before, this specific configMap resource need to install CRs inside tenant cluster, using kubeconfig library, doesn't it?

If `CRs` are `ConfigMaps` here then it's just a matter of configuring right `K8sClient`.

tomahawk28:
Yes, I thought that as well but @rossf7 said that it's because tenant cluster so we have to think about whether it's right to use the same logic as the control plane.

kopiczko:
The desired/current state logic is different but creating/updating/deleting CM should stay the same. Unless I don't understand some detail.

rossf7:
The resource needs to use the kubeconfig library so we connect to the correct cluster. So we can't configure the kubeconfig in advance. Code is below.

With the app-operator design its the only resource that needs it. So I think it makes sense to have a separate resource.

If we were only connecting to local clusters then we could use the generic resource.

```go
cc, err := controllercontext.FromContext(ctx)
if err != nil {
	return microerror.Mask(err)
}

_, err = cc.K8sClient.CoreV1().ConfigMaps(configMap.Namespace).Create(configMap)
```

kopiczko:
Interesting problem. Still you can work around that. You can setup the underlying resource on every loop and just wrap it's methods. E.g. 

```go
	err := r.underlying.ApplyCreateChange(ctx, obj)
	if err != nil {
		return microerror.Mask(err)
	}

	return nil
}
```

For the long run maybe we should add `K8sClient()` to the `StateGetter` interface?

rossf7:
I thought about this a lot. For the configmap and secret resources in the app controller I think having full resources in the operator is the right approach.

The extra logic for connecting to tenants is only needed here. We also needed to make lots of fixes in cluster-operator for connecting to tenants and cancel the reconciliation when they are unavailable. We don't need to complicate the generic resources with that.

With the app-operator design cluster-operator gets simpler because it only manages resources in the control plane. There it makes total sense to use the generic resources and we'll do that.

cc @puja108 @tomahawk28

kopiczko:
> We also needed to make lots of fixes in cluster-operator for connecting to tenants and cancel the reconciliation when they are unavailable. We don't need to complicate the generic resources with that.

Cancelling reconciliation should be okay. If you do it in desired/current methods the rest of the resource won't be executed.

tomahawk28:
I think it's best to move this topic to sig-operator channel, WDYT, I'll organize the topic.

rossf7:
@tomahawk28 I don't think discussing in the channel is a good idea. We've already discussed this multiple times.

I feel strongly on this and I want to continue with the current approach.

If you want to discuss more lets do it in the sync on Wednesday and make a decision then.

kopiczko:
Okay. I added a card to discuss that. I'll be possibly AFK during this sync. If I'm not there please let's keep the card till next sync.

PR #53, opened 2019-03-05
Add test for service creation and updated catalog index flag

rossf7:
Adds a test for service creation to make sure all necessary flags are set.

I'd missed the k8s client from the configmap resource which caused the operator to crash when deployed to a test install.

Also renames `f.Service.AppCatalog.IndexNamespace` to `f.Service.AppCatalog.Index.Namespace` to align with `f.Service.Kubernetes.Watch.Namespace`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -105,7 +105,7 @@ func mainWithError() (err error) {
 
 	daemonCommand := newCommand.DaemonCommand().CobraCommand()
 
-	daemonCommand.PersistentFlags().String(f.Service.AppCatalog.IndexNamespace, "giantswarm", "The namespace where operator keep index file as configMap.")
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -105,7 +105,7 @@ func mainWithError() (err error) {
 
 	daemonCommand := newCommand.DaemonCommand().CobraCommand()
 
-	daemonCommand.PersistentFlags().String(f.Service.AppCatalog.IndexNamespace, "giantswarm", "The namespace where operator keep index file as configMap.")
+	daemonCommand.PersistentFlags().String(f.Service.AppCatalog.Index.Namespace, "giantswarm", "The namespace where operator keep index file as configMap.")
```
xh3b4sd: <3

PR #54, opened 2019-03-05
Updates to chart resource

rossf7:
This is prep for enabling the configmap resource.

### DIFF
--- a/service/controller/app/v1/key/key.go
+++ b/service/controller/app/v1/key/key.go
@@ -1,11 +1,25 @@
 package key
 
 import (
+	"fmt"
+
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/microerror"
 
 	"github.com/giantswarm/app-operator/pkg/label"
 )
+
+// AppConfigMapName returns the name of the configmap that stores app level
+// config for the provided app CR.
+func AppConfigMapName(customResource v1alpha1.App) string {
+	return customResource.Spec.Config.ConfigMap.Name
+}
+
+// AppConfigMapNamespace returns the namespace of the configmap that stores app
+// level config for the provided app CR.
+func AppConfigMapNamespace(customResource v1alpha1.App) string {
+	return customResource.Spec.Config.ConfigMap.Namespace
+}
 
 func AppName(customResource v1alpha1.App) string {
 	return customResource.Spec.Name
@@ -23,12 +37,10 @@ func ChartStatus(customResource v1alpha1.Chart) v1alpha1.ChartStatus {
 	return customResource.Status
 }
 
-func ConfigMapName(customResource v1alpha1.App) string {
-	return customResource.Spec.Config.ConfigMap.Name
-}
-
-func ConfigMapNamespace(customResource v1alpha1.App) string {
-	return customResource.Spec.Config.ConfigMap.Namespace
+// ChartConfigMapName returns the name of the configmap that stores config for
+// the chart CR that is generated for the provided app CR.
+func ChartConfigMapName(customResource v1alpha1.App) string {
+	return fmt.Sprintf("%s-chart-values", customResource.ObjectMeta.Name)
 }
 
 func InCluster(customResource v1alpha1.App) bool {
--- a/service/controller/app/v1/key/key.go
+++ b/service/controller/app/v1/key/key.go
@@ -40,7 +40,7 @@ func ChartStatus(customResource v1alpha1.Chart) v1alpha1.ChartStatus {
 // ChartConfigMapName returns the name of the configmap that stores config for
 // the chart CR that is generated for the provided app CR.
 func ChartConfigMapName(customResource v1alpha1.App) string {
-	return fmt.Sprintf("%s-chart-values", customResource.ObjectMeta.Name)
+	return fmt.Sprintf("%s-chart-values", customResource.GetName())
 }
 
 func InCluster(customResource v1alpha1.App) bool {
--- a/service/controller/app/v1/resource/chart/current.go
+++ b/service/controller/app/v1/resource/chart/current.go
@@ -27,7 +27,7 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding chart %#q", name))
 
-	chart, err := cc.G8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, metav1.GetOptions{})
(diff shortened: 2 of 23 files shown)


service/controller/app/v1/key/key.go
```diff
@@ -1,12 +1,26 @@
...
+	"fmt"
+
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/microerror"
 
 	"github.com/giantswarm/app-operator/pkg/label"
 )
 
+// AppConfigMapName returns the name of the configmap that stores app level
+// config for the provided app CR.
```
rossf7: Previous `ConfigMapName` function wasn't clear. This gets the name of the configmap that stores app level config for this CR.

service/controller/app/v1/key/key.go
```diff
@@ -23,12 +37,10 @@ func ChartStatus(customResource v1alpha1.Chart) v1alpha1.ChartStatus {
...
 }
 
-func ConfigMapName(customResource v1alpha1.App) string {
-	return customResource.Spec.Config.ConfigMap.Name
-}
-
-func ConfigMapNamespace(customResource v1alpha1.App) string {
-	return customResource.Spec.Config.ConfigMap.Namespace
+// ChartConfigMapName returns the name of the configmap that stores config for
+// the chart CR that is generated for the provided app CR.
```
rossf7: This returns the name of the configmap that is generated by the configmap resource. It is used by chart-operator when calling tiller to install or update the chart.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -38,34 +38,35 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 			APIVersion: chartAPIVersion,
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        cr.Spec.Name,
+			Name:        cr.GetName(),
```
rossf7: This was wrong. cr.Spec.Name is the name of the app being installed e.g. prometheus. We now use the same name as the app CR.

It is up to the component creating app CRs to ensure the name is unique.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -38,34 +38,35 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
 		},
 	}
 
+	if key.AppConfigMapName(cr) != "" || appcatalogkey.ConfigMapName(cc.AppCatalog) != "" {
+		config := v1alpha1.ChartSpecConfig{
+			ConfigMap: v1alpha1.ChartSpecConfigConfigMap{
+				Name:      key.ChartConfigMapName(cr),
+				Namespace: key.Namespace(cr),
+			},
+		}
```
rossf7: We should only set the config section if there is config to apply. Secret support will be added later when the secret resource is added.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -38,34 +38,35 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
+		chartCR.Spec.Config = config
+	}
+
 	return chartCR, nil
 }
 
 func generateTarballURL(baseURL string, appName string, version string) (string, error) {
 	if baseURL == "" || appName == "" || version == "" {
-		return "", microerror.Maskf(failedExecution, "baseURL %#q, appName %#q, release %#q should not be empty", baseURL, appName, version)
+		return "", microerror.Maskf(failedExecutionError, "baseURL %#q, appName %#q, release %#q should not be empty", baseURL, appName, version)
```
rossf7: Sneaking this in to correct the type name.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -38,34 +38,35 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
-					Namespace: key.SecretNamespace(cr),
-				},
-			},
+			Name:       cr.ObjectMeta.Name,
+			Namespace:  key.Namespace(cr),
 			TarballURL: tarballURL,
 		},
 	}
 
+	if key.AppConfigMapName(cr) != "" || appcatalogkey.ConfigMapName(cc.AppCatalog) != "" {
```
tomahawk28: doesn't it have to be `&&`?

rossf7: No an OR is fine here. The app CR has 3 levels of config (catalog, app and user). But all 3 levels are optional.

The user support will come later because its not needed for VOO.

service/controller/app/v1/resource/chart/current.go
```diff
@@ -18,7 +18,7 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
-	name := key.AppName(cr)
+	name := cr.ObjectMeta.Name
```
xh3b4sd:
```suggestion
	name := cr.GetName()
```

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -38,34 +38,35 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
-				ConfigMap: v1alpha1.ChartSpecConfigConfigMap{
-					Name:      key.ConfigMapName(cr),
-					Namespace: key.ConfigMapNamespace(cr),
-				},
-				Secret: v1alpha1.ChartSpecConfigSecret{
-					Name:      key.SecretName(cr),
-					Namespace: key.SecretNamespace(cr),
-				},
-			},
+			Name:       cr.ObjectMeta.Name,
```
xh3b4sd:
```suggestion
			Name:       cr.GetName(),
```

service/controller/app/v1/resource/chart/error.go
```diff
@@ -2,13 +2,13 @@ package chart
 
 import "github.com/giantswarm/microerror"
 
-var failedExecution = &microerror.Error{
+var failedExecutionError = &microerror.Error{
```
xh3b4sd: We call this `executionFailedError` usually.

service/controller/app/v1/resource/chart/current.go
rossf7: Good catch I've updated to use the functions whenever accessing the ObjectMeta struct.

service/controller/app/v1/resource/chart/error.go
rossf7: Damn missed that. Its renamed now.

PR #55, opened 2019-03-06
Updated namespace handling in chart resource

rossf7:
Adds new flag for the namespace chart CRs should be created in. Currently the namespace in the app CR is used but this is incorrect.

e.g. App namespace for coredns will be kube-system but the chart CR should be created in the giantswarm namespace.

### DIFF
--- a/flag/service/chart/chart.go
+++ b/flag/service/chart/chart.go
@@ -1,0 +1,7 @@
+package chart
+
+// Chart is a data structure to hold Chart custom resource specific
+// configuration.
+type Chart struct {
+	Namespace string
+}
--- a/service/controller/app/app.go
+++ b/service/controller/app/app.go
@@ -11,7 +11,7 @@ import (
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/client-go/kubernetes"
 
-	"github.com/giantswarm/app-operator/service/controller/app/v1"
+	v1 "github.com/giantswarm/app-operator/service/controller/app/v1"
 )
 
 type Config struct {
@@ -20,6 +20,7 @@ type Config struct {
 	K8sExtClient apiextensionsclient.Interface
 	Logger       micrologger.Logger
 
+	ChartNamespace string
 	ProjectName    string
 	WatchNamespace string
 }
@@ -80,6 +81,7 @@ func NewApp(config Config) (*App, error) {
 	var resourceSetV1 *controller.ResourceSet
 	{
 		c := v1.ResourceSetConfig{
+			ChartNamespace: config.ChartNamespace,
 			G8sClient:      config.G8sClient,
 			K8sClient:      config.K8sClient,
 			Logger:         config.Logger,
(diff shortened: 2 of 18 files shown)


flag/service/chart/chart.go
```diff
@@ -0,0 +1,7 @@
+package chart
+
+// Chart is a data structure to hold Chart custom resource specific
+// configuration.
+type Chart struct {
+	Namespace string
```
rossf7: This is the new flag.

service/controller/app/v1/resource/chart/create.go
```diff
@@ -29,7 +29,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 			return microerror.Mask(err)
 		}
 
-		_, err = cc.G8sClient.ApplicationV1alpha1().Charts(cr.Namespace).Create(&chart)
+		_, err = cc.G8sClient.ApplicationV1alpha1().Charts(cr.ObjectMeta.Namespace).Create(&chart)
```
rossf7: The namespace will now be set in the desired state. So we use that here.

service/controller/app/v1/resource/chart/current.go
```diff
@@ -27,7 +27,7 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding chart %#q", name))
 
-	chart, err := cc.G8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, metav1.GetOptions{})
+	chart, err := cc.G8sClient.ApplicationV1alpha1().Charts(r.chartNamespace).Get(name, metav1.GetOptions{})
```
rossf7: Flag is used when getting the current state.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -39,6 +39,7 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        cr.Spec.Name,
+			Namespace:   r.chartNamespace,
```
rossf7: Setting the namespace.

service/controller/app/app.go
```diff
@@ -11,7 +11,7 @@ import (
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/client-go/kubernetes"
 
-	"github.com/giantswarm/app-operator/service/controller/app/v1"
+	v1 "github.com/giantswarm/app-operator/service/controller/app/v1"
```
xh3b4sd: I would say either `appv1` or no alias. This one is kind of pointless.

service/controller/app/v1/resource/chart/create.go
xh3b4sd: I think you can also do `cr.GetNamespace()`.

service/controller/app/app.go
rossf7: Good catch. This was just my editor misbehaving. Fixed.

service/controller/app/v1/resource/chart/create.go
rossf7: Good shout I went with that. The CR also has a Namespace field in the spec so I think calling GetNamespace is the clearest option.

PR #56, opened 2019-03-06
Order update methods and remove unused setting

rossf7:
Couple of tidying up things I want to get in. So diffs are cleaner in following PRs.

### DIFF
--- a/service/controller/app/v1/resource/chart/update.go
+++ b/service/controller/app/v1/resource/chart/update.go
@@ -3,6 +3,7 @@ package chart
 import (
 	"context"
 	"fmt"
+
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 
 	"github.com/giantswarm/microerror"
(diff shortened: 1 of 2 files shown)


service/controller/app/v1/resource/chart/update.go
```diff
@@ -12,24 +13,6 @@ import (
 	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
 )
 
-func (r *Resource) NewUpdatePatch(ctx context.Context, obj, currentChart, desiredChart interface{}) (*controller.Patch, error) {
```
rossf7: Just fixing the ordering.

service/controller/app/v1/resource_set.go
```diff
@@ -28,9 +28,8 @@ type ResourceSetConfig struct {
 	Logger    micrologger.Logger
 
 	// Settings.
-	HandledVersionBundles []string
```
rossf7: This isn't needed.

PR #59, opened 2019-03-07
Add secret resource with current and desired state

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/4845

This starts adding the secret resource. This generates the secret that will be consumed by chart-operator. Its basically the same as the configmap resource. 

I'm not using a generic operatorkit resource because it needs to connect to control plane and tenant clusters. So it needs kubeconfig support and more error handling for the tenant cluster not being available.

### DIFF
--- a/service/controller/app/v1/resource/secret/current.go
+++ b/service/controller/app/v1/resource/secret/current.go
@@ -1,0 +1,43 @@
+package secret
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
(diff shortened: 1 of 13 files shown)


service/controller/app/v1/resource/secret/desired.go
```diff
@@ -0,0 +1,114 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/app-operator/pkg/label"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+	appcatalogkey "github.com/giantswarm/app-operator/service/controller/appcatalog/v1/key"
+)
+
+func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
```
rossf7: Same deal for the desired state.

service/controller/app/v1/resource/secret/current.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/giantswarm/microerror"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
```
rossf7: The current state. This is the same as for configmap just for the different k8s resource.

PR #62, opened 2019-03-07
Add create and update support to secret resource

rossf7:

Same deal. This is the basically the same as the configmap resource.

### DIFF
--- a/service/controller/app/v1/resource/secret/create.go
+++ b/service/controller/app/v1/resource/secret/create.go
@@ -12,26 +12,26 @@ import (
 )
 
 func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
-	configMap, err := toSecret(createChange)
+	secret, err := toSecret(createChange)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
-	if !isEmpty(configMap) {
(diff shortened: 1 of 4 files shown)


tomahawk28:
Have you thought about using Pawel's change from Hakathon?

kopiczko:
> Have you thought about using Pawel's change from Hakathon?

There is no generic `Secret` resource yet.

rossf7:
@tomahawk28 See https://github.com/giantswarm/app-operator/pull/59#issue-259152732.

We will need extra logic for connecting to control plane clusters that is only needed in app-operator.

Everywhere else like in the `index` resources  you added and the upcoming `clusterconfigmap` resource in cluster-operator we can use the generic resource.

rossf7:
> There is no generic Secret resource yet.

@kopiczko Yes we will likely need a generic secret resource at some point.

At the moment we don't need to generate any secrets in cluster-operator for the managed services but I think we will eventually.

kopiczko:
> > There is no generic Secret resource yet.
> 
> @kopiczko Yes we will likely need a generic secret resource at some point.
> 
> At the moment we don't need to generate any secrets in cluster-operator for the managed services but I think we will eventually.

I'll try to generate one for `Secret` and `Namespace` this week. I think those two are the most common.

PR #63, opened 2019-03-07
Delete and wiring

rossf7:

Adds delete support to complete the resource and wires it up.


helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -67,6 +67,7 @@ rules:
   - ""
   resources:
   - configmaps
+  - secrets
```
rossf7: The permissions here are the same as for confgmaps but they are wide. But I don't think we can restrict by namespace because we need to manage secrets in multiple namespaces. WDYT?

cc @puja108

PR #64, opened 2019-03-08
Set desired state for the config block in the chart resource

rossf7:
Sets the config block for the chart CR with the configmap and or secret. Depending on the config that has been specified in the app CR.

### DIFF
--- a/service/controller/app/v1/resource/chart/desired.go
+++ b/service/controller/app/v1/resource/chart/desired.go
@@ -27,6 +27,7 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
+	config := generateConfig(cr, cc.AppCatalog)
 	tarballURL, err := generateTarballURL(appcatalogkey.AppCatalogStorageURL(cc.AppCatalog), key.AppName(cr), key.Version(cr))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


rossf7:
Test

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -44,24 +45,40 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
-	if key.AppConfigMapName(cr) != "" || appcatalogkey.ConfigMapName(cc.AppCatalog) != "" {
-		config := v1alpha1.ChartSpecConfig{
-			ConfigMap: v1alpha1.ChartSpecConfigConfigMap{
+	return chartCR, nil
+}
+
+func generateConfig(cr v1alpha1.App, appCatalog v1alpha1.AppCatalog) v1alpha1.ChartSpecConfig {
+	config := v1alpha1.ChartSpecConfig{}
+
+	if hasConfigMap(cr, appCatalog) || hasSecret(cr, appCatalog) {
```
xh3b4sd: That condition is not necessary. The specific conditions are below and do what they should.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -76,6 +93,22 @@ func generateTarballURL(baseURL string, appName string, version string) (string,
...
+func hasConfigMap(cr v1alpha1.App, appCatalog v1alpha1.AppCatalog) bool {
+	if key.AppConfigMapName(cr) != "" || appcatalogkey.ConfigMapName(appCatalog) != "" {
+		return true
+	}
+
+	return false
+}
+
+func hasSecret(cr v1alpha1.App, appCatalog v1alpha1.AppCatalog) bool {
+	if key.AppSecretName(cr) != "" || appcatalogkey.SecretName(appCatalog) != "" {
```
xh3b4sd: Where is the difference between app and appcatalog secret name? Why do check for both?

rossf7: The app CR supports 3 levels of config app, catalog and user. Any of the 3 can be specified. But if none are specified we won't create a secret for chart-operator.

The user config support will be added later because it isn't needed for VOO.

service/controller/app/v1/resource/chart/desired.go
rossf7: Thanks. I've removed it.

service/controller/app/v1/resource/chart/desired.go
xh3b4sd: When there are 3, why do we only check for two?

xh3b4sd: Oh, because user config support comes later. With you.

PR #65, opened 2019-03-12
Cancel status resource when chart does not exist

rossf7:
Fixes a couple of issues with the status resource found during testing.

### DIFF
--- a/service/controller/app/v1/resource/status/create.go
+++ b/service/controller/app/v1/resource/status/create.go
@@ -19,7 +19,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	name := key.AppName(cr)
+	name := cr.GetName()
 
 	cc, err := controllercontext.FromContext(ctx)
 	if err != nil {
--- a/service/controller/app/v1/resource/status/create.go
+++ b/service/controller/app/v1/resource/status/create.go
@@ -26,7 +26,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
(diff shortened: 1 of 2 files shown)


service/controller/app/v1/resource/status/create.go
```diff
@@ -19,7 +19,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	name := key.AppName(cr)
+	name := cr.GetName()
```
rossf7: AppName is the name of the app e.g. `prometheus`. We use `cr.GetName()` so we can have multiple instances of an app in a cluster e.g. `my-cool-prometheus`.

See https://github.com/giantswarm/app-operator/blob/ed5cfae10c0035db0763eded7dfd998124c46d02/service/controller/app/v1/resource/chart/desired.go#L42 for where the name is set.

service/controller/app/v1/resource/status/create.go
```diff
@@ -30,7 +30,10 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 	chart, err := cc.G8sClient.ApplicationV1alpha1().Charts(r.watchNamespace).Get(name, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
-		return microerror.Maskf(notFoundError, "chart %#q in namespace %#q", name, r.watchNamespace)
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart %#q in namespace %#q", name, r.watchNamespace))
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+		return nil
```
rossf7: During testing I hit timing issues because the chart CR had not been created yet. We should just cancel the resource and try again later when the chart is not found.

PR #67, opened 2019-03-12
Add support for kubeconfig flag

rossf7:
Adds kubeconfig support for use with `kind`. See https://github.com/giantswarm/operatorkit/pull/289 for the related operatorkit change.

- Creating a cluster.

```bash
go get -u sigs.k8s.io/kind && kind create cluster 
```

- Running locally.

```bash
kubeconfig=`cat $(kind get kubeconfig-path --name="kind")` && \
go run main.go daemon --service.kubernetes.kubeconfig="$kubeconfig"
```

https://github.com/kubernetes-sigs/kind


Gopkg.toml
```diff
@@ -65,7 +65,7 @@ required = [
   name = "github.com/giantswarm/micrologger"
 
 [[constraint]]
-  branch = "master"
+  branch = "kubeconfig"
```

flag/service/kubernetes/kubernetes.go
```diff
@@ -8,8 +8,9 @@ import (
...
 type Kubernetes struct {
-	Address   string
-	InCluster string
-	TLS       tls.TLS
-	Watch     watch.Watch
+	Address    string
+	InCluster  string
+	KubeConfig string
+	TLS        tls.TLS
+	Watch      watch.Watch
```
xh3b4sd: What do you think about having the common flags in `operatorkit` and using the structures in the operators' flag package by importing? We could then be consistent via vendor updates.

rossf7: Yes it makes sense to move the flags to operatorkit. I'd like to add the kubeconfig support to all operators so I can use the vendored flags then.

rossf7: Done. PTAL.

PR #68, opened 2019-03-12
Test kubeconfig

tomahawk28:
- Create new simple e2etest using [giantswarm/kubeconfig](https://github.com/giantswarm/kubeconfig/) library
- This test is using `kubeconfig` secret resource to access new namespace. 
- We could create a real tenant cluster but this take too much time right now.


integration/ensure/error.go
```diff
@@ -0,0 +1,9 @@
+// +build k8srequired
+
+package ensure
+
+import "github.com/giantswarm/microerror"
+
+var testError = &microerror.Error{
+	Kind: "testError",
+}
```
xh3b4sd: The matcher is missing. See `fmt`.

tomahawk28:
@xh3b4sd I will make it readable both of tests

tomahawk28:
@xh3b4sd I tried to make it simpler than before, https://github.com/giantswarm/app-operator/pull/68/commits/5658cfd8b75d0e5aa29cd86d9a3e748f74544aae
Before I changed the same things on `kubeconfig_test.go`, I want to check with you. 
Is that okay from your point of view?

integration/test/app/basic/simple_test.go
```diff
@@ -69,34 +61,56 @@ func TestAppLifecycle(t *testing.T) {
...
-		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating chart %#q", customResourceReleaseName))
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating chart value for release: %#q", customResourceReleaseName))
 
-		chartValues, err := chartvalues.NewAPIExtensionsAppE2E(sampleChart)
+		chartValues, err = chartvalues.NewAPIExtensionsAppE2E(sampleChart)
 		if err != nil {
 			t.Fatalf("expected %#v got %#v", nil, err)
 		}
 
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("created chart value for release: %#q", customResourceReleaseName))
```
xh3b4sd: Let'S drop the colons everywhere. 

```suggestion
		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("created chart value for release %#q", customResourceReleaseName))
```

integration/test/app/basic/simple_test.go
```diff
@@ -69,34 +61,56 @@ func TestAppLifecycle(t *testing.T) {
...
+	{
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("waiting for release deployed: %#q", customResourceReleaseName))
+
 		err = config.Release.WaitForStatus(ctx, fmt.Sprintf("%s-%s", namespace, customResourceReleaseName), "DEPLOYED")
 		if err != nil {
 			t.Fatalf("expected %#v got %#v", nil, err)
 		}
 
-		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("created chart %#q", customResourceReleaseName))
+		config.Logger.LogCtx(ctx, "level", "debug", "message", "waited for release deployed")
```
xh3b4sd: The opening and closing logs should be equal despite the tense.

tomahawk28:
@rossf7 I'm already using `config.Release.WaitForStatus` even before this change, do you mean another method perhaps? like `config.Release.WaitForChartInfo`?
After installing release, I need to check `chart` CR have been created and not sure how I can do that /w `config.Release` package.

rossf7:
> After installing release, I need to check chart CR have been created

The chart CR gets created and is configured to install a helm release called `test-app`. So you can use `config.Release.WaitForStatus` to wait until `test-app` has status `DEPLOYED`.

At that point you can be sure that the chart CR was processed successfully and chart-operator processed it. WDYT would that work?

tomahawk28:
@rossf7 I didn't set up chart-operator as you did in chart-operator e2etest, so I'm afraid that I missed some components (e.g. CNR, DNS setup ) that I need to install `chart`. https://github.com/giantswarm/app-operator/blob/8720e3267653129b9fa0dc9429bdab8a9b0d2728/integration/setup/setup.go#L66-L71

rossf7:
>  I missed some components (e.g. CNR, DNS setup ) that I need to install chart.

Those are not needed. CNR is only used by chartconfig CRs. The chart CR just pulls the tarball from the sample catalog repo.

The DNS settings are for installing CoreDNS in tenant clusters. They are also not needed.

tomahawk28:
@xh3b4sd @rossf7 PTAL :)

PR #73, opened 2019-04-04
remove namespace constraint

tomahawk28:
- Remove `kubernetes.watch.namespace` constraint to watch resources in all namespaces.


service/controller/app/v1/appcatalog/appcatalog.go
```diff
@@ -63,7 +59,7 @@ func (a *AppCatalog) GetCatalogForApp(ctx context.Context, customResource v1alph
 
 	a.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("looking for appCatalog %#q in namespace %#q", catalogName, a.watchNamespace))
 
-	appCatalog, err := a.g8sClient.ApplicationV1alpha1().AppCatalogs(a.watchNamespace).Get(catalogName, metav1.GetOptions{})
+	appCatalog, err := a.g8sClient.ApplicationV1alpha1().AppCatalogs("default").Get(catalogName, metav1.GetOptions{})
```
tomahawk28: If we do not set it as `default`, it failed as below. 
```
E 04/04 13:34:13 /apis/application.giantswarm.io/v1alpha1/namespaces/giantswarm/apps/test-app stop reconciliation due to error | operatorkit/controller/controller.go:347 | controller=app-operator | event=update | loop=1 | version=4254
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/controller/resource_set.go:72:
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/service/controller/app/v1/resource_set.go:204:
	/Users/jihyukbok/go/src/github.com/giantswarm/app-operator/service/controller/app/v1/appcatalog/appcatalog.go:66:
	an empty namespace may not be set when a resource name is provided
```
Anyway `appcatalog` is a cluster-based scope resource, so it doesn't matter.

tomahawk28:
I changed arguments as below
```
      - args:
        - daemon
        - --config.dirs=/var/run/app-operator/configmap/
        - --config.files=config
        - --service.kubernetes.watch.namespace=""
```
I saw that it successfully deployed into `ginger` but it still look for the `app` CR inside `giantswarm` namespace.

helm/app-operator-chart/templates/configmap.yaml
```diff
@@ -21,4 +21,4 @@ data:
       kubernetes:
         incluster: true
         watch:
-          namespace: '{{ .Values.namespace }}'
```
oponder: I feel like this templated value could still be kept, and instead we change it in the values.yaml to be `""` ?

https://github.com/giantswarm/app-operator/blob/master/helm/app-operator-chart/values.yaml

Or otherwise we should remove it from values.yaml as well, so it's not some old code.

oponder:
@kopiczko: What do you want us to see there? I'm trying to understand why that link is relevant here.

tomahawk28:
SIGSEGV during deletion of chart: ```{"caller":"github.com/giantswarm/app-operator/service/controller/app/v1/resource/chart/delete.go:70","controller":"app-operator","event":"delete","function":"NewDeletePatch","level":"debug","loop":"0","message":"finding out if the `kubernetes-test-app-chart` chart has to be deleted","object":"/apis/application.giantswarm.io/v1alpha1/namespaces/giantswarm/apps/kubernetes-test-app-chart","resource":"chartv1","time":"2019-04-05T07:19:34.76653+00:00","version":"561"}
{"caller":"github.com/giantswarm/app-operator/service/controller/app/v1/resource/chart/delete.go:78","controller":"app-operator","event":"delete","function":"NewDeletePatch","level":"debug","loop":"0","message":"the `kubernetes-test-app-chart` chart does not need to be deleted","object":"/apis/application.giantswarm.io/v1alpha1/namespaces/giantswarm/apps/kubernetes-test-app-chart","resource":"chartv1","time":"2019-04-05T07:19:34.766565+00:00","version":"561"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x117fd89]

goroutine 37 [running]:
github.com/giantswarm/app-operator/service/controller/app/v1/resource/chart.(*Resource).ApplyDeleteChange(0xc000432280, 0x1605ea0, 0xc00045f230, 0x142e040, 0xc0001a0000, 0x0, 0x0, 0x112cc6c, 0xc0003f9420)
	/go/src/github.com/giantswarm/app-operator/service/controller/app/v1/resource/chart/delete.go:27 +0x129
github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/controller/resource/retryresource.(*crudResourceWrapperOps).ApplyDeleteChange.func1(0xc000409c80, 0xc0000a3920)
	/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/controller/resource/retryresource/crud_resource_ops_wrapper.go:182 +0x6d
github.com/giantswarm/app-operator/vendor/github.com/cenkalti/backoff.RetryNotify(0xc000038640, 0x7fe2b8414700, 0xc000409c80, 0xc0000a3900, 0x0, 0x0)
	/go/src/github.com/giantswarm/app-operator/vendor/github.com/cenkalti/backoff/retry.go:37 +0xbb
github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/controller/resource/retryresource.(*crudResourceWrapperOps).ApplyDeleteChange(0xc0004309c0, 0x1605ea0, 0xc00045f230, 0x142e040, 0xc0001a0000, 0x0, 0x0, 0xc0003743f0, 0xc0004916d8)```

kopiczko:
> @kopiczko: What do you want us to see there? I'm trying to understand why that link is relevant here.

The types should be generated with nonNamespaced option see this comment: https://github.com/giantswarm/app-operator/pull/73#discussion_r272185022

The linked PR shows how to do that.

helm/app-operator-chart/templates/configmap.yaml
tomahawk28: But that would impact other part as well, just like `deployment.yaml` for example. That should have `giantswarm` namespace.

tomahawk28:
@kopiczko PTAL

service/controller/app/v1/resource_set.go
```diff
@@ -161,7 +158,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 			G8sClient: config.G8sClient,
 			Logger:    config.Logger,
 
-			WatchNamespace: config.WatchNamespace,
```
tomahawk28: I found it should look into chartNamespace, not for watchNamespace.

helm/app-operator-chart/templates/configmap.yaml
kopiczko: I don't like that as well. We should keep that. Isn't that used by the app controller?
It should be ignored in the appcatalog controller.

puja108: App controller also needs to watch multiple namespaces, as we want app CRs to not be in giantswarm namespace

puja108: But I do agree that this should stay templated, just need to be careful that the deployment namespace and the watch namespace are not the same but two different template variables, having them as the same like before was wrong

puja108: And then we need condition als if namespace is set or not

service/controller/app/v1/appcatalog/appcatalog.go
puja108: I think this might be an issue in our library there as we should not require namespaces for non-namespaced resources. Your trick here is a quick workaround but we should fix this in general, maybe sth to talk with @giantswarm/sig-operator about

puja108: I think thatmight be what @kopiczko linked to below

puja108: So no bug

puja108:
@tomahawk28 the SIGSEGV error looks like a PM @rossf7 just closed

service/controller/app/v1/appcatalog/appcatalog.go
kopiczko: This is outdated & fixed already.

helm/app-operator-chart/templates/configmap.yaml
tomahawk28: I just pushed the changed waht @puja108 suggest me.

helm/app-operator-chart/templates/configmap.yaml
```diff
@@ -21,4 +21,5 @@ data:
...
+          {{ if .Values.watchingNamespace }} namespace: '{{ .Values.watchingNamespace }}' {{ else }} namespace: ''
```
puja108:
```suggestion
          {{ if .Values.watchNamespace }} namespace: '{{ .Values.watchNamespace }}' {{ else }} namespace: ''
```

for consistency

tomahawk28: Done, Thanks

helm/app-operator-chart/templates/configmap.yaml
kopiczko: Why do we need an `if` below? Empty string is fine here. Also I think it brakes formatting because there is a single space before `namespace:`.

service/controller/app/v1/resource/chart/create.go
```diff
@@ -31,7 +25,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 			return microerror.Mask(err)
 		}
 
-		_, err = cc.G8sClient.ApplicationV1alpha1().Charts(cr.GetNamespace()).Create(chart)
+		_, err = cc.G8sClient.ApplicationV1alpha1().Charts(r.chartNamespace).Create(chart)
```
kopiczko:
```suggestion
		_, err = cc.G8sClient.ApplicationV1alpha1().Charts(chart.Namespace).Create(chart)
```

PR #75, opened 2019-04-11
Do not copy annotation from app

tomahawk28:
We are copying `annotation` in `app` CR into `chart` CR. 
We only need to copy `label`, not `annotation`.

### DIFF
--- a/service/controller/app/v1/resource/chart/delete.go
+++ b/service/controller/app/v1/resource/chart/delete.go
@@ -69,8 +69,8 @@ func (r *Resource) newDeleteChange(ctx context.Context, obj, currentState, desir
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding out if the %#q chart has to be deleted", desiredChart.Name))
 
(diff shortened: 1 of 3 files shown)


tomahawk28:
@rossf7 Can you help me on here to understand it? 
I really don't understand why `chart` is still remaining after delete. 
it pass this condition, https://github.com/giantswarm/app-operator/blob/1a12a1ccb24e6f0428c4ea72ec6cdb59158f0dce/service/controller/app/v1/resource/chart/delete.go#L75

rossf7:
@tomahawk28 I'm not 100% but I think its because equals compares annotations.

https://github.com/giantswarm/app-operator/blob/1a12a1ccb24e6f0428c4ea72ec6cdb59158f0dce/service/controller/app/v1/resource/chart/resource.go#L88-L90

We should remove that too and just let k8s manage the annotations.

tomahawk28:
Not only annotation but TypeMeta.Kind, APIVersion in `currentChart` are empty string but `desiredChart` had filled that information. Let me look at that pass as well.

tomahawk28:
So it works! I will delete debug msgs and merged it :)

PR #77, opened 2019-04-26
Initial collector structure

tomahawk28:
- Initial collector structure for app-operator

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/giantswarm/microerror"
@@ -62,7 +63,7 @@ func mainWithError() (err error) {
 				panic(fmt.Sprintf("%#v\n", microerror.Maskf(err, "service.New")))
 			}
 
-			go newService.Boot()
+			go newService.Boot(context.Background())
 		}
 
 		// New custom server that bundles microkit endpoints.
--- a/main.go
+++ b/main.go
@@ -31,6 +31,8 @@ func main() {
 }
 
 func mainWithError() (err error) {
+	ctx := context.Background()
+
 	// Create a new logger that is used by all packages.
 	var newLogger micrologger.Logger
 	{
@@ -63,7 +65,7 @@ func mainWithError() (err error) {
 				panic(fmt.Sprintf("%#v\n", microerror.Maskf(err, "service.New")))
 			}
 
-			go newService.Boot(context.Background())
+			go newService.Boot(ctx)
 		}
 
 		// New custom server that bundles microkit endpoints.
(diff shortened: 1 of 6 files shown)


service/collector/app_operator.go
```diff
@@ -0,0 +1,56 @@
...
+
+// AppOperatorConfig is this collector's configuration struct.
+type AppOperatorConfig struct {
+	K8sClient kubernetes.Interface
+	G8sClient versioned.Interface
+	Logger    micrologger.Logger
+}
+
+// AppOperator is the main struct for this collector.
+type AppOperator struct {
```
tomahawk28: Can't think about the right name but atm I just set it as `AppOperator`

rossf7: I think the name should be `AppResource` to match what we do in chart-operator. https://github.com/giantswarm/chart-operator/blob/master/service/collector/chart_resource.go

I think this will be the first metric but its fine for it to be just the empty structure for now.

service/service.go
```diff
@@ -159,6 +176,8 @@ func New(config Config) (*Service, error) {
 // Boot starts top level service implementation.
 func (s *Service) Boot() {
 	s.bootOnce.Do(func() {
+		go s.operatorCollector.Boot(context.Background())
```
rossf7: The wiring here is outdated. The context should be passed in like here.

https://github.com/giantswarm/cluster-operator/blob/0ed29d8467a3cd7d1c2c46406dcd2dbd0b6b0542/service/service.go#L283

main.go
```diff
@@ -62,7 +63,7 @@ func mainWithError() (err error) {
 				panic(fmt.Sprintf("%#v\n", microerror.Maskf(err, "service.New")))
 			}
 
-			go newService.Boot()
+			go newService.Boot(context.Background())
```
rossf7: I think the context should be defined at the top of `mainWithError` rather than inline. As we might add more logic there later.

service/collector/app_resource.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/client-go/kubernetes"
+)
+
+// AppResourceConfig is this collector's configuration struct.
+type AppResourceConfig struct {
+	K8sClient kubernetes.Interface
```
rossf7: Params should be ordered.

service/service.go
```diff
@@ -129,6 +131,20 @@ func New(config Config) (*Service, error) {
 		}
 	}
 
+	var resourceCollector *collector.Set
```
rossf7:
```suggestion
	var operatorCollector *collector.Set
```

I think its better to use operatorCollector like we do in aws-operator. As the collector will have multiple metrics not just for resources.

PR #79, opened 2019-04-29
Merge catalog and app level secrets

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/5885

This PR implements merging catalog and app level secrets. Currently we error if the app CR has both levels of config.

The app level secrets override any catalog level secrets. A later PR will implement user level secrets which allows customers to override these values per tenant cluster.

### DIFF
--- a/service/controller/app/v1/resource/secret/desired.go
+++ b/service/controller/app/v1/resource/secret/desired.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/helmclient"
 	"github.com/giantswarm/microerror"
+	yaml "gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


service/controller/app/v1/resource/secret/desired.go
```diff
@@ -35,17 +37,34 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
-	data, err := r.getSecretData(ctx, cr)
+	secretsYAML, err := yaml.Marshal(mergedData)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
 
 	secret := &corev1.Secret{
-		Data: data,
+		Data: map[string][]byte{
+			"values": secretsYAML,
```
rossf7: The key name can have any name. The only restriction in chart-operator is only a single key with YAML is supported.

service/controller/app/v1/resource/secret/desired_test.go
```diff
@@ -256,10 +334,23 @@ func Test_Resource_GetDesiredState(t *testing.T) {
 				if !reflect.DeepEqual(secret.ObjectMeta, tc.expectedSecret.ObjectMeta) {
 					t.Fatalf("want matching objectmeta \n %s", cmp.Diff(secret.ObjectMeta, tc.expectedSecret.ObjectMeta))
 				}
-				if !reflect.DeepEqual(secret.StringData, tc.expectedSecret.StringData) {
-					t.Fatalf("want matching data \n %s", cmp.Diff(secret.StringData, tc.expectedSecret.StringData))
+				if !reflect.DeepEqual(secret.Data, tc.expectedSecret.Data) {
+					data := toStringMap(secret.Data)
+					expectedData := toStringMap(tc.expectedSecret.Data)
```
rossf7: This is an extra step to have better diffs. Byte array diffs are evil.

PR #81, opened 2019-04-30
Use key package for release names

rossf7:
We need to add e2e test coverage for the values support. This is some prep to use the `key` package for more Helm release names.

### DIFF
--- a/integration/test/app/basic/simple_test.go
+++ b/integration/test/app/basic/simple_test.go
@@ -3,13 +3,15 @@
 package basic
 
 import (
+	"context"
 	"fmt"
-	"golang.org/x/net/context"
 	"testing"
 
 	"github.com/giantswarm/e2e-harness/pkg/release"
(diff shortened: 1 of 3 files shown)


integration/test/app/basic/simple_test.go
```diff
@@ -3,21 +3,21 @@
...
 
 const (
-	namespace                 = "giantswarm"
-	customResourceReleaseName = "apiextensions-app-e2e-chart"
-	chartOperatorVersion      = "chart-operator.giantswarm.io/version"
-	testAppReleaseName        = "kubernetes-test-app-chart"
-	testAppCatalogReleaseName = "test-app-catalog"
+	namespace            = "giantswarm"
+	chartOperatorVersion = "chart-operator.giantswarm.io/version"
+	testAppCatalogName   = "test-app-catalog"
```
rossf7: I renamed this from `testAppCatalogReleaseName` because it isn't actually a release name. More of these constants can move to the key package but I didn't want to make the diff too noisy.

PR #82, opened 2019-04-30
Fix bug when configmap name or secret name is empty

rossf7:
Fixes https://github.com/giantswarm/giantswarm/issues/5919

Fixes a problem @webwurst hit when installing an app in `talos` earlier. This is a bug I introduced with my recent config merging changes. 

Error is `resource name may not be empty` when fetching a configmap or secret. So now we return early if the name is empty.

### DIFF
--- a/service/controller/app/v1/resource/configmap/desired.go
+++ b/service/controller/app/v1/resource/configmap/desired.go
@@ -79,8 +79,8 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 
 func (r *Resource) getConfigMap(ctx context.Context, configMapName, configMapNamespace string) (map[string]string, error) {
 	if configMapName == "" {
-		// Return early.
-		return map[string]string{}, nil
+		// Return early as no configmap has been specified.
+		return nil, nil
 	}
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("looking for configmap %#q in namespace %#q", configMapName, configMapNamespace))
--- a/service/controller/app/v1/resource/configmap/desired.go
+++ b/service/controller/app/v1/resource/configmap/desired.go
@@ -77,7 +77,12 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 	return configMap, nil
 }
 
-func (r *Resource) getConfigMap(ctx context.Context, configMapName, configMapNamespace string) (*corev1.ConfigMap, error) {
+func (r *Resource) getConfigMap(ctx context.Context, configMapName, configMapNamespace string) (map[string]string, error) {
+	if configMapName == "" {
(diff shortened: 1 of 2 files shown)


service/controller/app/v1/resource/configmap/desired.go
```diff
@@ -77,7 +77,12 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 	return configMap, nil
 }
 
-func (r *Resource) getConfigMap(ctx context.Context, configMapName, configMapNamespace string) (*corev1.ConfigMap, error) {
+func (r *Resource) getConfigMap(ctx context.Context, configMapName, configMapNamespace string) (map[string]string, error) {
+	if configMapName == "" {
+		// Return early.
+		return map[string]string{}, nil
```
xh3b4sd: Does it has to be an initialized empty map? Zero value would be nil if no other reason applies to do otherwise.

rossf7: Good point on the zero value. The data will be run through `helmclient.MergeValues` so the empty map helps there. But actually I think `MergeValues` should handle nil values.

I can make it nil and also update `helmclient`. WDYT?

xh3b4sd: Code in general should be able to deal with zero values. Makes it more resilient and easier to comprehend here. Otherwise it looks like there is some deeper meaning to it that nobody gets.

rossf7: Good point, thanks. I'm going to change `helmclient.MergeValues` to handle nils and then simplify here by returning nil.

service/controller/app/v1/resource/configmap/desired.go
```diff
@@ -77,7 +77,12 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 	return configMap, nil
 }
 
-func (r *Resource) getConfigMap(ctx context.Context, configMapName, configMapNamespace string) (*corev1.ConfigMap, error) {
+func (r *Resource) getConfigMap(ctx context.Context, configMapName, configMapNamespace string) (map[string]string, error) {
+	if configMapName == "" {
```
kopiczko: Why can this be an empty string?

rossf7: All levels of config in the app CR are optional. So it will be empty if no config has been set. 

I think it's better to check once here and return early. Rather than checking separately for catalog, app and user config.

kopiczko: 👍

service/controller/app/v1/resource/configmap/desired.go
rossf7: Returning nil works without needing changes to helmclient. So going with that to simplify.

PR #84, opened 2019-05-01
Merge user configmaps

rossf7:
Towards giantswarm/giantswarm#5885

Same as https://github.com/giantswarm/app-operator/pull/86 but for configmaps.

### DIFF
--- a/service/controller/app/v1/key/key.go
+++ b/service/controller/app/v1/key/key.go
@@ -69,6 +69,18 @@ func Namespace(customResource v1alpha1.App) string {
 	return customResource.Spec.Namespace
 }
 
+// UserSecretName returns the name of the secret that stores user level
+// secrets for the provided app CR.
+func UserSecretName(customResource v1alpha1.App) string {
+	return customResource.Spec.UserConfig.Secret.Name
+}
+
+// UserSecretNamespace returns the namespace of the secret that stores user
+// level secrets for the provided app CR.
+func UserSecretNamespace(customResource v1alpha1.App) string {
+	return customResource.Spec.UserConfig.Secret.Namespace
+}
+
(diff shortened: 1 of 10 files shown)


service/controller/app/v1/key/key.go
```diff
@@ -96,6 +84,30 @@ func ToCustomResource(v interface{}) (v1alpha1.App, error) {
...
+func UserSecretName(customResource v1alpha1.App) string {
+	return customResource.Spec.UserConfig.Secret.Name
+}
+
+// UserSecretNamespace returns the namespace of the secret that stores user
+// level secrets for the provided app CR.
+func UserSecretNamespace(customResource v1alpha1.App) string {
+	return customResource.Spec.UserConfig.Secret.Namespace
+}
+
```
pipo02mix: why it is added again?

rossf7: Hmm because I got the merge wrong somehow 😄

Thx checking!

rossf7: OK its because the ordering was wrong before. This fixes it.

service/controller/app/v1/key/key.go
```diff
@@ -69,18 +69,6 @@ func Namespace(customResource v1alpha1.App) string {
...
-// secrets for the provided app CR.
-func UserSecretName(customResource v1alpha1.App) string {
-	return customResource.Spec.UserConfig.Secret.Name
-}
-
-// UserSecretNamespace returns the namespace of the secret that stores user
-// level secrets for the provided app CR.
-func UserSecretNamespace(customResource v1alpha1.App) string {
-	return customResource.Spec.UserConfig.Secret.Namespace
-}
```
rossf7: Old funcs get deleted here to fix the ordering.

PR #85, opened 2019-05-02
Fix chart namespace

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/5923

The issue was operator searching the `chart` CR in a wrong namespace.

### DIFF
--- a/service/controller/app/v1/resource/chart/delete.go
+++ b/service/controller/app/v1/resource/chart/delete.go
@@ -26,7 +26,7 @@ func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange inte
 			return microerror.Mask(err)
 		}
 
-		err = cc.G8sClient.ApplicationV1alpha1().Charts(r.chartNamespace).Delete(chart.Name, &metav1.DeleteOptions{})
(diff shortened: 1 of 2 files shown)


service/controller/app/v1/resource/chart/delete.go
```diff
@@ -32,7 +26,7 @@ func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange inte
 			return microerror.Mask(err)
 		}
 
-		err = cc.G8sClient.ApplicationV1alpha1().Charts(cr.GetNamespace()).Delete(chart.Name, &metav1.DeleteOptions{})
+		err = cc.G8sClient.ApplicationV1alpha1().Charts(r.chartNamespace).Delete(chart.Name, &metav1.DeleteOptions{})
```
kopiczko: Can it be following?

```suggestion
		err = cc.G8sClient.ApplicationV1alpha1().Charts(chart.Namespace).Delete(chart.Name, &metav1.DeleteOptions{})
```

tomahawk28: Well that's same but looks nicer than before :)

PR #86, opened 2019-05-02
Merge user secrets

rossf7:
Towards giantswarm/giantswarm#5885

This PR implemented merging user secrets for app CRs. Next PR will be the same for configmaps.

### DIFF
--- a/service/controller/app/v1/resource/secret/desired.go
+++ b/service/controller/app/v1/resource/secret/desired.go
@@ -30,8 +30,9 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 
 	appSecretName := key.AppSecretName(cr)
 	catalogSecretName := appcatalogkey.SecretName(cc.AppCatalog)
+	userSecretName := key.UserSecretName(cr)
 
-	if appSecretName == "" && catalogSecretName == "" {
+	if appSecretName == "" && catalogSecretName == "" && userSecretName == "" {
 		// Return early as there is no secret.
 		return nil, nil
 	}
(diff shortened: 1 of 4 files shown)


service/controller/app/v1/resource/secret/desired.go
```diff
@@ -30,8 +30,9 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 
 	appSecretName := key.AppSecretName(cr)
 	catalogSecretName := appcatalogkey.SecretName(cc.AppCatalog)
+	userSecretName := key.UserSecretName(cr)
 
-	if appSecretName == "" && catalogSecretName == "" {
+	if appSecretName == "" && catalogSecretName == "" && userSecretName == "" {
```
rossf7: If all 3 levels of config are empty we don't create a secret for the chart CR.

service/controller/app/v1/resource/secret/desired.go
```diff
@@ -55,6 +56,21 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
+			return nil, microerror.Mask(err)
+		}
+
+		// Secrets are merged again and in case of intersecting values the user
+		// level secrets are preferred.
+		mergedData, err = values.MergeSecretData(mergedData, userData)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
```
rossf7: We do a 2nd merge if a user secret has been configured.

The 2 step process is because most app CRs won't have user config. It is only for managed apps where we or the customer need to override values per tenant cluster.

service/controller/app/v1/resource/secret/desired_test.go
```diff
@@ -267,6 +267,7 @@ func Test_Resource_GetDesiredState(t *testing.T) {
 			},
 			expectedSecret: &corev1.Secret{
 				Data: map[string][]byte{
+					// "test: app" overrides "test: catalog".
```
rossf7: The comment is because the test is a bit subtle. I don't want to use big blocks of YAML and we test `helmclient.MergeValues` in the library tests.

service/controller/app/v1/resource/secret/desired_test.go
```diff
@@ -278,6 +279,88 @@ func Test_Resource_GetDesiredState(t *testing.T) {
...
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-user-secrets",
+						Namespace: "giantswarm",
+					},
+				},
+			},
+			expectedSecret: &corev1.Secret{
+				Data: map[string][]byte{
+					// "test: user" overrides "test: catalog" and "test: app".
```
rossf7: New case for the user config.

PR #87, opened 2019-05-03
Deactivate index resource

tomahawk28:
Toward: https://github.com/giantswarm/giantswarm/issues/5932

- We added `empty` resource since we can't keep the controller structure w/o resources.


rossf7:
We should also delete the index resource. No point keeping the dead code.

PR #89, opened 2019-05-07
First app resource metrics

tomahawk28:
- Adding initial `app` metrics into app-operator


service/collector/set.go
```diff
@@ -12,6 +12,8 @@ type SetConfig struct {
 	G8sClient versioned.Interface
 	K8sClient kubernetes.Interface
 	Logger    micrologger.Logger
+
+	WatchNamespace string
```
tomahawk28: Adding the watchNamesapce to follow up the `app` CR location.

service/collector/app_resource.go
```diff
@@ -38,19 +61,63 @@ func NewAppResource(config AppResourceConfig) (*AppResource, error) {
...
+
+	var wg sync.WaitGroup
+
+	for _, collectFunc := range collectFuncs {
+		wg.Add(1)
+
+		go func(collectFunc func(ctx context.Context, ch chan<- prometheus.Metric)) {
+			defer wg.Done()
+			collectFunc(ctx, ch)
+		}(collectFunc)
```
xh3b4sd: All this `collectFunc` stuff is not necessary. We do this in the `aws-operator` because of the different AWS accounts we need to cover for BYOC. This is ugly in general and not necessary for apps. The logic of `collectAppStatus` should be what we want.

service/collector/set.go
xh3b4sd: Why do we need that though? Why can't we list all apps from all namespaces?

tomahawk28:
@xh3b4sd Even if we watch all namespaces, `app-operator` only operates on a specific namespace that we set by a flag, so I think it makes sense that we do this way.

service/collector/app_resource.go
tomahawk28: Thanks, I will delete the whole waitgroup section of this area.

xh3b4sd:
The question was why.

service/collector/set.go
tomahawk28: Can we do that using k8sclient?

tomahawk28: Okay I found at https://github.com/kubernetes/client-go/issues/159#issuecomment-288624475
I will follow it as all namespaces.

xh3b4sd: Yeah but again, can you please explain why we have the namespace limitation at all in the first place? There might be a reason for that. So before we change code back and forth let'S understand why things are in a certain way.

tomahawk28: `app-operator` is watching the namespace (e.g. giantswarm) which is the user set as a flag. So I thought that `collector` also should do the same thing.

xh3b4sd: I understand that, but that does not explain why it is like that and we both do apparently not understand why it is like that. So it would be really good if you could figure out why it is like that, so that we can then figure out how it should be for the collector. Do not just follow existing code. Let's understand and have a reason. Talk to people and see what you can get together. Your team should be able to help you in the first place.

tomahawk28: I don't think this is the thing that I should ask my team to find out the reason since I wrote most of the codes in this operator perhaps the question was not clear. 
Anyway, In the current situation, `app` CR is installed inside cluster namespace, so namespace limitation inside `collector` is wrong I believe.

service/collector/app_resource.go
```diff
@@ -1,13 +1,31 @@
...
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	"github.com/prometheus/client_golang/prometheus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
+var (
+	appDesc *prometheus.Desc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, "", "app"),
```
JosephSalisbury: So this is `app_operator_app`? I'd expect just `app_info`

service/collector/app_resource.go
```diff
@@ -43,14 +61,40 @@ func NewAppResource(config AppResourceConfig) (*AppResource, error) {
 	return c, nil
 }
 
+func (c *AppResource) collectAppStatus(ctx context.Context, ch chan<- prometheus.Metric) {
+	r, err := c.g8sClient.ApplicationV1alpha1().Apps("").List(metav1.ListOptions{})
+	if err != nil {
+		c.logger.LogCtx(ctx, "level", "error", "message", "could not get apps", "stack", fmt.Sprintf("%#v", err))
```
JosephSalisbury: If we error here, is `r` `nil`? Because then you'll panic later

service/collector/app_resource.go
tomahawk28: Good catch, I'll change the name in here suffix from `app` into `info`

service/collector/set.go
xh3b4sd: > I don't think this is the thing that I should ask my team to find out the reason since I wrote most of the codes in this operator

And yet neither of us seems to have a convincing explanation. I stop bothering you now. Do what you think is right.

tomahawk28:
```#!/bin/sh -eo pipefail
# Blocked due to plan-no-credits-available Please purchase a new credit block then push a new commit or call the API to run a new build.
# 
# -------
# Warning: This configuration was auto-generated to show you the message above.
# Don't rerun this job. Rerunning will have no effect.
false```
Marcel already contact to CircleCI team regarding this issue.

tomahawk28:
@JosephSalisbury PTAL

service/collector/app_resource.go
tomahawk28: I fixed it, thanks for notice.

service/collector/app_resource.go
JosephSalisbury: The `_info` suffix is useful for these kinds of metrics, so i'd like it to be `app_info`

PR #90, opened 2019-05-09
Add reason field

tomahawk28:
Toward suggestion made in https://github.com/giantswarm/giantswarm/issues/5751
- Update reason when helm chart had been crashed.


service/controller/app/v1/resource/status/create.go
```diff
@@ -45,6 +45,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		AppVersion: chartStatus.AppVersion,
 		Release: v1alpha1.AppStatusRelease{
 			LastDeployed: *chartStatus.Release.LastDeployed.DeepCopy(),
```
xh3b4sd: FMI why does that need a deep copy?

tomahawk28: I think it was @kopiczko who introduce this practice on chart-operator, can you tell me the reason?

kopiczko: I don't remember me doing that and if I did I don't remember why :D

xh3b4sd: When there is no reason we should drop it and keep it simple. [`LastDeployed`](https://godoc.org/github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1#AppStatusRelease) is only `DeepCopyTime`.

tomahawk28: Do you mind if I open another PR for that, I would like to keep the subject like this

xh3b4sd: Sure

PR #94, opened 2019-05-10
Clear status when chart is not ready

tomahawk28:
- When the operator could not find related `chart`, it should clear status in `app` CR so the user would not misunderstand the current situation.

### DIFF
--- a/service/controller/app/v1/resource/status/create.go
+++ b/service/controller/app/v1/resource/status/create.go
@@ -32,6 +32,21 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	if errors.IsNotFound(err) {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart %#q in namespace %#q", name, r.chartNamespace))
 		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+
+		currentCR, err := r.getCurrentCR(cr)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		currentCR.Status = v1alpha1.AppStatus{}
+
+		_, err = r.g8sClient.ApplicationV1alpha1().Apps(cr.Namespace).UpdateStatus(currentCR)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("status clear for app %#q in namespace %#q", name, r.chartNamespace))
+
 		return nil
 
 	} else if err != nil {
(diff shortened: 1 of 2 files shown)


service/controller/app/v1/resource/status/create.go
```diff
@@ -32,6 +32,21 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	if errors.IsNotFound(err) {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart %#q in namespace %#q", name, r.chartNamespace))
 		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+
+		currentCR, err := r.getCurrentCR(cr)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		currentCR.Status = v1alpha1.AppStatus{}
```
tomahawk28: Should we check it first? 

```golang
if currentCR.Status == v1alpha.AppStatus{} {
  // only clear status in this case
}
```

service/controller/app/v1/resource/status/create.go
```diff
@@ -74,3 +89,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 	return nil
 }
+
+func (r *Resource) getCurrentCR(cr v1alpha1.App) (*v1alpha1.App, error) {
```
xh3b4sd: I would vote to not do this. The benefit is not really given IMO.

service/controller/app/v1/resource/status/create.go
```diff
@@ -32,6 +32,21 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+		}
+
+		currentCR.Status = v1alpha1.AppStatus{}
+
+		_, err = r.g8sClient.ApplicationV1alpha1().Apps(cr.Namespace).UpdateStatus(currentCR)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("status clear for app %#q in namespace %#q", name, r.chartNamespace))
```
xh3b4sd: There should be an opening and closing log, just like everywhere else in the operators. The wording should also be reworked.

service/controller/app/v1/resource/status/create.go
xh3b4sd: Probably yes.

tomahawk28:
@xh3b4sd In `ginger`, I saw that one of `app` CR had a status as `DEPLOYED` even if that tenant cluster had been deleted already. In this case, I think it makes sense to clear the status.

xh3b4sd:
Why is there still an App CR when the TC is deleted?

puja108:
> Why is there still an App CR when the TC is deleted?

That is a good question, we need to make sure that App CRs get cleaned on cluster deletion.not sure how we do this for other resources that belong to clusters but I guess a component like cluster operator needs to take care of this smh

tomahawk28:
I think we can close this PR after merging https://github.com/giantswarm/app-operator/pull/95 since that was the main reason why app CR was remaining there.

PR #95, opened 2019-05-14
Fix app CR deletion logic

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/5979

1. In `GetCurrentState` on each resource, the operator should not return errors otherwise `app` CR would remain in the namespace. 

2. Simply deletion logic. Do not need to compare in this case.

### DIFF
--- a/service/controller/app/v1/resource/chart/current.go
+++ b/service/controller/app/v1/resource/chart/current.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/giantswarm/errors/tenant"
 	"github.com/giantswarm/microerror"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -31,9 +32,11 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	if errors.IsNotFound(err) {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart %#q in namespace %#q", name, r.chartNamespace))
 		return nil, nil
-	} else if err != nil {
+	} else if tenant.IsAPINotAvailable(err) {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("could not find chart %#q in namespace %#q due to an error: %#q", name, r.chartNamespace, err))
 		return nil, nil
+	} else if err != nil {
+		return nil, microerror.Mask(err)
 	}
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found chart %#q", name))
--- a/service/controller/app/v1/resource/chart/current.go
+++ b/service/controller/app/v1/resource/chart/current.go
@@ -33,7 +33,7 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart %#q in namespace %#q", name, r.chartNamespace))
 		return nil, nil
 	} else if tenant.IsAPINotAvailable(err) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("could not find chart %#q in namespace %#q due to an error: %#q", name, r.chartNamespace, err))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("could not find chart %#q in namespace %#q", name, r.chartNamespace))
 		return nil, nil
 	} else if err != nil {
 		return nil, microerror.Mask(err)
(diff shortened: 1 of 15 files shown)


service/controller/app/v1/resource/configmap/delete.go
```diff
@@ -40,37 +40,22 @@ func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange inte
 }
 
 func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desiredState interface{}) (*controller.Patch, error) {
-	delete, err := r.newDeleteChange(ctx, obj, currentState, desiredState)
+	del, err := r.newDeleteChange(ctx, obj, currentState, desiredState)
```
tomahawk28: `delete` is reserved in golang, so I changed to `del`

service/controller/app/v1/resource/secret/desired.go
```diff
@@ -18,6 +18,15 @@ import (
 )
 
 func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
+	deleted, err := key.IsDeleted(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	if deleted {
```
tomahawk28: Return early when its `app` CR is going to be deleted.

service/controller/app/v1/resource/chart/current.go
```diff
@@ -29,10 +29,11 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 
 	chart, err := cc.G8sClient.ApplicationV1alpha1().Charts(r.chartNamespace).Get(name, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart %#q", name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart %#q in namespace %#q", name, r.chartNamespace))
 		return nil, nil
 	} else if err != nil {
-		return nil, microerror.Mask(err)
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("could not find chart %#q in namespace %#q due to an error: %#q", name, r.chartNamespace, err))
+		return nil, nil
```
xh3b4sd: I don't think this is right. When there are unexpected errors we should return them to the caller. We only seem to have problems with errors that we kind of expect, so we should match them and handle them specifically. The "not found error" is such a special case already. We only need to add another special case.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -17,6 +17,15 @@ import (
 )
 
 func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
+	deleted, err := key.IsDeleted(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	if deleted {
+		return nil, nil
```
xh3b4sd: That does not make sense at all and when you test this you will notice that you just return the desired state in `newDeleteChange`, which now is nil. So nothing will be deleted in fact. `GetDesiredState` should not be changed at all IMO.

service/controller/app/v1/resource/chart/current.go
tomahawk28: In the original issue, the operator got an error as below. ```x509: certificate is valid for ingress.local, not api.hd39g.k8s.ginger.eu-central-1.aws.gigantic.io``` I'm not sure which cases we should not report as an error in order to not prevent namespace deletion.

xh3b4sd: Well, this case at least I think. We can use something like this here: https://godoc.org/github.com/giantswarm/errors/tenant.

tomahawk28: Wow thanks! Great hint!

tomahawk28:
@xh3b4sd PTAL

service/controller/app/v1/resource/chart/current.go
```diff
@@ -29,7 +30,10 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 
 	chart, err := cc.G8sClient.ApplicationV1alpha1().Charts(r.chartNamespace).Get(name, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart %#q", name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart %#q in namespace %#q", name, r.chartNamespace))
+		return nil, nil
+	} else if tenant.IsAPINotAvailable(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("could not find chart %#q in namespace %#q due to an error: %#q", name, r.chartNamespace, err))
```
xh3b4sd: I would remove the error from the log and keep it consistent with the log above.

tomahawk28:
@xh3b4sd We had discussed that already and we decided to not use the generic resource in here since this resource is targetting tenant cluster, not its own cluster. When it failed to reconcile the resource, we may need to do some backup for TC. That's what I understood from the previous discussion. Correct if I'm wrong, @kopiczko

tomahawk28:
Sorry I will take care more of it and reopen this PR when I'm done.

PR #97, opened 2019-05-16
Adding kubeconfig finalizer

tomahawk28:
- Adding custom finalizer into kubeconfig secret to prevent it from deleting before we clean up all depending app CRs.

### DIFF
--- a/service/controller/app/v1/key/key.go
+++ b/service/controller/app/v1/key/key.go
@@ -65,6 +65,18 @@ func InCluster(customResource v1alpha1.App) bool {
 	return customResource.Spec.KubeConfig.InCluster
 }
 
+func KubeConfigFinalizer(customResource v1alpha1.App) string {
+	return fmt.Sprintf("%s/%s", label.FinalizerPrefix, customResource.GetName())
+}
+
+func KubecConfigSecretName(customResource v1alpha1.App) string {
+	return customResource.Spec.KubeConfig.Secret.Name
+}
+
+func KubecConfigSecretNamespace(customResource v1alpha1.App) string {
+	return customResource.Spec.KubeConfig.Secret.Namespace
+}
+
 func Namespace(customResource v1alpha1.App) string {
 	return customResource.Spec.Namespace
 }
--- a/service/controller/app/v1/key/key.go
+++ b/service/controller/app/v1/key/key.go
@@ -66,7 +66,7 @@ func InCluster(customResource v1alpha1.App) bool {
 }
 
 func KubeConfigFinalizer(customResource v1alpha1.App) string {
-	return fmt.Sprintf("%s/%s", label.FinalizerPrefix, customResource.GetName())
+	return fmt.Sprintf("app-operator.giantswarm.io/%s", customResource.GetName())
 }
 
 func KubecConfigSecretName(customResource v1alpha1.App) string {
--- a/service/controller/app/v1/resource/kubeconfigfinalizer/create.go
+++ b/service/controller/app/v1/resource/kubeconfigfinalizer/create.go
@@ -18,7 +18,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	}
 
 	if key.InCluster(cr) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q do not use kubeconfig secret since it would install the chart in the same cluster", cr.GetName()))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q does not use a kubeconfig secret for apps in the control plane", cr.GetName()))
 		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
 		return nil
 	}
--- a/service/controller/app/v1/resource/kubeconfigfinalizer/create.go
+++ b/service/controller/app/v1/resource/kubeconfigfinalizer/create.go
@@ -17,6 +17,12 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	if key.InCluster(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("In-cluster used"))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("no need to put the finalizer"))
+		return nil
+	}
+
 	name := key.KubecConfigSecretName(cr)
 	namespace := key.KubecConfigSecretNamespace(cr)
 
--- a/service/controller/app/v1/resource/kubeconfigfinalizer/create.go
+++ b/service/controller/app/v1/resource/kubeconfigfinalizer/create.go
@@ -3,6 +3,7 @@ package kubeconfigfinalizer
 import (
 	"context"
 	"fmt"
+
 	"github.com/giantswarm/microerror"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -36,7 +37,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	finalizerTag := key.KubeConfigFinalizer(cr)
 
 	if !contains(kubeConfig.Finalizers, finalizerTag) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting finalizer for kubeconfig %#q in namespace %#q", name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting finalizer for kubeconfig %#q", name))
 
 		kubeConfig.Finalizers = append(kubeConfig.Finalizers)
 
@@ -45,7 +46,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			return microerror.Mask(err)
 		}
 	} else {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer already set for kubeconfig %#q in namespace %#q", name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer already set for kubeconfig %#q", name))
 	}
 	return nil
 }
--- a/service/controller/app/v1/resource/kubeconfigfinalizer/create.go
+++ b/service/controller/app/v1/resource/kubeconfigfinalizer/create.go
@@ -1,0 +1,60 @@
+package kubeconfigfinalizer
+
+import (
+	"context"
+	"fmt"
+	"github.com/giantswarm/microerror"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	name := key.KubecConfigSecretName(cr)
+	namespace := key.KubecConfigSecretNamespace(cr)
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding kubeconfig secret for app %#q", name))
+
+	kubeConfig, err := r.k8sClient.CoreV1().Secrets(namespace).Get(name, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find kubeconfig %#q", name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+		return nil
+
+	} else if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found kubeconfig secret for app %#q", name))
+
+	finalizerTag := key.KubeConfigFinalizer(cr)
+
+	if !contains(kubeConfig.Finalizers, finalizerTag) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting finalizer for kubeconfig %#q in namespace %#q", name))
+
+		kubeConfig.Finalizers = append(kubeConfig.Finalizers)
+
+		_, err := r.k8sClient.CoreV1().Secrets(namespace).Update(kubeConfig)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	} else {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer already set for kubeconfig %#q in namespace %#q", name))
+	}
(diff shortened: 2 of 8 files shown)


service/controller/app/v1/resource_set.go
```diff
@@ -131,6 +132,19 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		}
 	}
 
+	var kubeconfigFinalizerResource controller.Resource
```
tomahawk28: This name is a bit more, but I don't want to make it shorten since we already have `kubeconfig` library we imported in here.

service/controller/app/v1/resource_set.go
```diff
@@ -174,6 +188,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		secretResource,
 		chartResource,
 		statusResource,
+		kubeconfigFinalizerResource,
```
tomahawk28: Put this resource on the last so we give chances to other resources to reconciliation with deletion events.

service/controller/app/v1/resource/kubeconfigfinalizer/delete.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if key.InCluster(cr) {
+		return nil
```
tomahawk28: Can we skip the logging in here I guess?

service/controller/app/v1/key/key.go
```diff
@@ -65,6 +65,18 @@ func InCluster(customResource v1alpha1.App) bool {
 	return customResource.Spec.KubeConfig.InCluster
 }
 
+func KubeConfigFinalizer(customResource v1alpha1.App) string {
+	return fmt.Sprintf("%s/%s", label.FinalizerPrefix, customResource.GetName())
```
xh3b4sd: Technically the finalizer prefix is not a label. I would not add it to the label package, but just add it here, maybe even inline.

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if key.InCluster(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("In-cluster used"))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("no need to put the finalizer"))
```
xh3b4sd: I do not understand what this is about. What is "in cluster"? Further, the "canceling resource" log is missing.

service/controller/app/v1/resource/kubeconfigfinalizer/delete.go
xh3b4sd: Why should we?

tomahawk28: Because we already put the logs on `create.go` so I was thinking that I put too many logs on it, but you think it's fine?

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
tomahawk28: `In-cluster` is for when you would like to install `chart` in CP, not in TC. 
https://github.com/giantswarm/giantswarm/blob/master/roadmap/MANAGED-SERVICES-CATALOG.MD#kubeconfig

service/controller/app/v1/resource/kubeconfigfinalizer/delete.go
xh3b4sd: When we do some action we should properly log about it. Doesn't matter what happens during creation when this here is about deletion. Just put yourself into the shoes of somebody who debugs the operator and does not know how it works. So this somebody watches the logs and wants to understand what happens. When we do not log anything, nobody has a clue what the program is doing.

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
xh3b4sd: So when it is "in cluster" there is no secret to be maintained during app deletion?

service/controller/app/v1/resource/kubeconfigfinalizer/delete.go
tomahawk28: Okay, I will put the same logs as creation, but adding `cancelling resource` as well.

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
tomahawk28: Yes, we don't have any kubeconfig secrets in `in-cluster` case.

xh3b4sd: Ok then I would like to have better logging and maybe even a comment in the code to explain all of this in a couple of sentences.

service/controller/app/v1/key/key.go
tomahawk28: Sorry I don't get what you mean in here, how we can put it as inline?

xh3b4sd: I thought about something like this. And then drop the changes in the label package. 

```suggestion
	return fmt.Sprintf("app-operator.giantswarm.io/%s", customResource.GetName())
```

xh3b4sd: I thought about something like this. And then drop the changes in the label package. 

```suggestion
	return fmt.Sprintf("app-operator.giantswarm.io/%s", customResource.GetName())
```

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
```diff
@@ -0,0 +1,67 @@
...
+		return nil
+	}
+
+	name := key.KubecConfigSecretName(cr)
+	namespace := key.KubecConfigSecretNamespace(cr)
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding kubeconfig secret for app %#q", name))
```
kopiczko: This is the app or the secret name? If this is the secret printing namespace would be useful. Printing app name here isn't very informative as this is already injected into `object` key in logs.

```suggestion
	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding kubeconfig secret for app %#q", name))
```

service/controller/app/v1/resource/kubeconfigfinalizer/delete.go
```diff
@@ -0,0 +1,65 @@
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	} else {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("kubeconfig %#q do not have matching finalizer", name))
+	}
+	return nil
+}
+
+func filter(finalizers []string, matching string) (ret []string) {
```
kopiczko: Please declare the `ret` below. Named return variables should be used for documentation purposes only AFAIK.

service/controller/app/v1/resource/kubeconfigfinalizer/delete.go
```diff
@@ -0,0 +1,65 @@
...
+
+		kubeConfig.Finalizers = filter(kubeConfig.Finalizers, finalizerTag)
+
+		_, err := r.k8sClient.CoreV1().Secrets(namespace).Update(kubeConfig)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	} else {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("kubeconfig %#q do not have matching finalizer", name))
+	}
```
kopiczko: I wonder if this is ok to just remove this finalizer here. Are you sure access to the TC isn't needed when we reach this resource? What if the Chart CR deletion on the TC is still in progress? I think there should be some sort of communication between resources or reconciliation cancellation.

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
tomahawk28: This is `app` name actually so I would change it to `secret` name with namespace. Thanks.

service/controller/app/v1/resource/kubeconfigfinalizer/delete.go
tomahawk28: Is it possible other resources are in progress still when we tried to delete finalizer in here? 
I thought we put this `kubeconfigfinalizer` at the last step so there should be no remaining tasks.

kopiczko: No but other resources should not block. They may cancel operation if Chart CR deletion is in progress.

tomahawk28: the operator does not check chart CR deletion in progress cases yet. I can add in `chart/current.go` to check CR deletion status and could cancel reconciliation at that point. 
But I would like to keep this PR as smaller as possible, Do you mind if I would start this after this PR merged?

xh3b4sd: We should start to think about a concept of reconciliation result so that we have a standardized way of communicating state of resource implementations at runtime. Our current primitive is the controller context but that is ugly and a bit unstructured.

kopiczko: Okay but let's keep this in mind.

tomahawk28:
@xh3b4sd @kopiczko I have addressed all the comments although there might be wording problems. Can you take another look?

Thanks!

tomahawk28:
Ah, find that delete.go do not update the same logging for namespace, sorry.

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if key.InCluster(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q do not use kubeconfig secret since it would install the chart in the same cluster", cr.GetName()))
```
xh3b4sd: Does that semantically make sense? 

```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q does not use a kubeconfig secret for apps in the control plane", cr.GetName()))
```

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
```diff
@@ -0,0 +1,67 @@
...
+	} else if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found kubeconfig secret %#q in namespace %#q", name, namespace))
+
+	finalizerTag := key.KubeConfigFinalizer(cr)
+
+	if !contains(kubeConfig.Finalizers, finalizerTag) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting finalizer for kubeconfig %#q in namespace %#q", name, namespace))
```
xh3b4sd: The closing log is missing.

service/controller/app/v1/resource/kubeconfigfinalizer/error.go
```diff
@@ -0,0 +1,23 @@
...
+var invalidConfigError = &microerror.Error{
+	Kind: "invalidConfigError",
+}
+
+// IsInvalidConfig asserts invalidConfigError.
+func IsInvalidConfig(err error) bool {
+	return microerror.Cause(err) == invalidConfigError
+}
+
+var notFoundError = &microerror.Error{
```
xh3b4sd: That error is not used for anything.

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
tomahawk28: Nice 🇬🇧

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if key.InCluster(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q does not use a kubeconfig secret for apps in the control plane", cr.GetName()))
```
TheoBrigitte: I would have had a more generic message here, mentioning control plane here is a bit confusing.
I understand that app using InCluster are installed on the control-plane, but I feel like this bit of information is too much here.

```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q does not use a kubeconfig secret", cr.GetName()))
```

service/controller/app/v1/resource/kubeconfigfinalizer/delete.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if key.InCluster(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q do not use kubeconfig secret since it installed the chart in the same cluster", cr.GetName()))
```
TheoBrigitte: Same direction as my previous comment, mentioning chart here feels confusing.

```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q do not use kubeconfig secret", cr.GetName()))
```

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
tomahawk28: Agreed, maybe it's too much information 👍

service/controller/app/v1/resource/kubeconfigfinalizer/resource.go
```diff
@@ -0,0 +1,44 @@
+package kubeconfigfinalizer
+
+import (
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	Name = "kubeconfigv1"
```
xh3b4sd: Just noticed that the resource name is wrong. It should be `kubeconfigfinalizerv1`.

rossf7: @tomahawk28 Can you update this pls?

tomahawk28: https://github.com/giantswarm/app-operator/pull/98 Sure, I just added

rossf7: @tomahawk28 Thx! ❤️

PR #99, opened 2019-05-28
app/v1/key: make finalizer prefix longer to avoid conflicts

kopiczko:
I asked for that in the original PR. I think this is reasonable to make
this prefix a bit longer since app names are arbitrary. That may (even
though unlikely) cause conflicts with finalizers added in the future.

### DIFF
--- a/service/controller/app/v1/key/key.go
+++ b/service/controller/app/v1/key/key.go
@@ -66,7 +66,7 @@ func InCluster(customResource v1alpha1.App) bool {
 }
 
 func KubeConfigFinalizer(customResource v1alpha1.App) string {
-	return fmt.Sprintf("app-operator.giantswarm.io/%s", customResource.GetName())
+	return fmt.Sprintf("app-operator.giantswarm.io/app-%s", customResource.GetName())
 }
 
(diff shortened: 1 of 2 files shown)


service/controller/app/v1/key/key.go
```diff
@@ -66,7 +66,7 @@ func InCluster(customResource v1alpha1.App) bool {
 }
 
 func KubeConfigFinalizer(customResource v1alpha1.App) string {
-	return fmt.Sprintf("app-operator.giantswarm.io/%s", customResource.GetName())
+	return fmt.Sprintf("app-operator.giantswarm.io/app-%s", customResource.GetName())
```
xh3b4sd: So how to deal with the differently named finalizers we already have?

kopiczko: I don't think we use this in production. Only for testing in some test installations.

I requested the review from @giantswarm/team-batman to confirm.

tomahawk28: Yes, we only have few cases for testing. We can go with it.

rossf7: EFK is running in prod in one CP but we're at an early stage. So I think we can change this and good to get it straight.

PR #100, opened 2019-06-11
Use alpine 3.9 as base image

rossf7:
I noticed this when doing the docs PR earlier. Getting it straight.

Not sure how important updating the other 3.8 images is?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.8
+FROM alpine:3.9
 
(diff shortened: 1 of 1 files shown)


marians:
Can't judge this here

rossf7:
Agreed. The is no security reason to upgrade AFAIK. So IMO this should be a passive "best efforts" upgrade rather than an active "do PRs for all the operators" type of upgrade.

PR #101, opened 2019-06-21
Resource cancelling when timeout & tenant API down

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/6289

- For this PR, the operator would check whether error matches `Timeout`,`TenantAPIError` cases and would cancel resource reconciliation. 

- First I would like to get approval from `configmap` first and *apply in general to other resources.*

### DIFF
--- a/service/controller/app/v1/resource/configmap/create.go
+++ b/service/controller/app/v1/resource/configmap/create.go
@@ -11,15 +11,9 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
-	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
 )
 
 func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
-	cr, err := key.ToCustomResource(obj)
-	if err != nil {
-		return microerror.Mask(err)
-	}
-
 	configMap, err := toConfigMap(createChange)
 	if err != nil {
 		return microerror.Mask(err)
@@ -36,16 +30,8 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		_, err = cc.K8sClient.CoreV1().ConfigMaps(configMap.Namespace).Create(configMap)
 		if apierrors.IsAlreadyExists(err) {
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("already created configmap %#q in namespace %#q", configMap.Name, configMap.Namespace))
-		} else if apierrors.IsTimeout(err) {
-			r.logger.LogCtx(ctx, "level", "debug", "message", "cluster api timeout.")
-
-			// We should not hammer API if it is not available, the cluster
-			// might be initializing. We will retry on next reconciliation loop.
-			resourcecanceledcontext.SetCanceled(ctx)
-			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
-
 			return nil
-		} else if !key.InCluster(cr) && tenant.IsAPINotAvailable(err) {
+		} else if tenant.IsAPINotAvailable(err) {
 			r.logger.LogCtx(ctx, "level", "debug", "message", "tenant cluster is not available.")
 
 			// We should not hammer tenant API if it is not available, the tenant cluster
--- a/service/controller/app/v1/resource/configmap/create.go
+++ b/service/controller/app/v1/resource/configmap/create.go
@@ -16,6 +16,10 @@ import (
 
 func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
 	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
 	configMap, err := toConfigMap(createChange)
 	if err != nil {
 		return microerror.Mask(err)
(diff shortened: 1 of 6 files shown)


service/controller/app/v1/resource/configmap/create.go
```diff
@@ -28,6 +32,24 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
...
+		} else if apierrors.IsTimeout(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "cluster api timeout.")
+
+			// We should not hammer API if it is not available, the cluster
+			// might be initializing. We will retry on next reconciliation loop.
+			resourcecanceledcontext.SetCanceled(ctx)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+
+			return nil
+		} else if !key.InCluster(cr) && tenant.IsAPINotAvailable(err) {
```
tomahawk28: I would check first whether it's InCluster or not.

service/controller/app/v1/resource/configmap/current.go
```diff
@@ -33,8 +34,23 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		// Return early as configmap does not exist.
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find configmap %#q in namespace %#q", name, r.chartNamespace))
 		return nil, nil
-	} else if tenant.IsAPINotAvailable(err) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find configmap %#q in namespace %#q", name, r.chartNamespace))
```
tomahawk28: The original log message in here was wrong, it should about the tenant API.

service/controller/app/v1/resource/configmap/create.go
```diff
@@ -4,14 +4,18 @@ import (
...
+	"github.com/giantswarm/operatorkit/controller/context/resourcecanceledcontext"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
 )
 
 func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
+	cr, err := key.ToCustomResource(obj)
```
xh3b4sd: I think when you review your own PR upfront you could recognize that this error here is not handled. Please make sure that such things are covered, as such things occur frequently in your PRs.

service/controller/app/v1/resource/configmap/create.go
```diff
@@ -28,6 +32,24 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		_, err = cc.K8sClient.CoreV1().ConfigMaps(configMap.Namespace).Create(configMap)
 		if apierrors.IsAlreadyExists(err) {
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("already created configmap %#q in namespace %#q", configMap.Name, configMap.Namespace))
+		} else if apierrors.IsTimeout(err) {
```
xh3b4sd: `tenant.IsAPINotAvailable` also checks for certain timeout errors. Why do we need to explicitly use `apierrors.IsTimeout` here now?

service/controller/app/v1/resource/configmap/create.go
xh3b4sd: I do not understand this condition. Why `!key.InCluster(cr)`? Which problem does this solve?

tomahawk28:
@xh3b4sd `!key.InCluster(cr)` is introduced there since we don't need to check `tenant` error in case if we want to install this chart in CP which  `InCluster = true` means (I think we talk about before about `InCluster`?)

service/controller/app/v1/resource/configmap/create.go
tomahawk28: I didn't know that `apierrors` check on it, since I saw the best practice on here https://github.com/giantswarm/cluster-operator/blob/4161b4dfdcb02b73c4574d1c6f9335ac0f514587/pkg/v17/resource/namespace/current.go#L57-L75

Did you add timeoutError later then? Should I update cluster-operator?

service/controller/app/v1/resource/configmap/create.go
tomahawk28: Shit, will cover this.

service/controller/app/v1/resource/configmap/create.go
tomahawk28: @rossf7 > Do you prefer to delete Timeout err check in general, since `tenant.IsAPINotAvailable` covers all the cases in here. (I can do update cluster-operator after this)

rossf7: Sorry, I'm late here and I added this when trying to fix the persistent connection problems we get in `viking`.

I think Tim is right here and we should just use `tenant.IsAPINotAvailable`. Yes please also remove from the WIP version of cluster-operator. <3

rossf7:
Adding my POV on the  `InCluster` questions. `app-operator` is a special snowflake because it needs to connect to both control plane and tenant clusters.

For managed apps in TCs we set the kubeconfig secret in the app CR.

```
      kubeConfig:
        context:
          name: "giantswarm-eggs2"
        secret:
          name: “giantswarm-eggs2”
          namespace: "eggs2"
```

For VOO we set the InCluster flag to true.

```
kubeConfig:
  context:
    name: "giantswarm-12345"
  inCluster: true
```

So if we want to do additional error handling for TCs the in cluster flag is the best one to use.

My question is why we can't use same error handling for all app CRs?

tomahawk28:
@rossf7 I think we can check errors in CP, just like we do in tenant cluster. After all, it's same k8s API request/response.

rossf7:
Good call let’s do that. Thx!

tomahawk28:
@xh3b4sd I deleted all APITimeoutError checking case since `tenant` package do all this, Thanks for checking!

PR #102, opened 2019-06-24
add a PSP and run as non-root

glitchcrab:
towards giantswarm/giantswarm/issues/6013


glitchcrab:
Not entirely sure why the CI tests are failing here

PR #104, opened 2019-07-11
Copy cordon related annotation into chart CR

tomahawk28:
Toward giantswarm/giantswarm#6078

It would copy all annotation from `app` CR into `chart` CR after operator-related prefix changes.

### DIFF
--- a/pkg/annotation/annotation.go
+++ b/pkg/annotation/annotation.go
@@ -4,10 +4,10 @@ package annotation
 
 const (
 	// CordonReason is the name of the annotation that indicates
-	// the reason of why app-operator should not apply any update on this app CR.
+	// the reason of why app-operator should not apply any update to this app CR.
 	CordonReason = "app-operator.giantswarm.io/cordon-reason"
 
 	// CordonUntil is the name of the annotation that indicates
-	// the expiration date of rule of this cordon.
+	// the expiration date for this cordon rule.
 	CordonUntil = "app-operator.giantswarm.io/cordon-until"
 )
--- a/pkg/annotation/annotation.go
+++ b/pkg/annotation/annotation.go
@@ -1,0 +1,13 @@
+// Package annotation contains common Kubernetes metadata. These are defined in
+// https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md.
+package annotation
+
+const (
+	// CordonReason is the name of the annotation that indicates
+	// the reason of why app-operator should not apply any update on this app CR.
+	CordonReason = "app-operator.giantswarm.io/cordon-reason"
+
+	// CordonUntilDate is the name of the annotation that indicates
+	// the expiration date of rule of this cordon.
+	CordonUntilDate = "app-operator.giantswarm.io/cordon-until"
+)
--- a/pkg/annotation/annotation.go
+++ b/pkg/annotation/annotation.go
@@ -7,7 +7,7 @@ const (
 	// the reason of why app-operator should not apply any update on this app CR.
 	CordonReason = "app-operator.giantswarm.io/cordon-reason"
 
-	// CordonUntilDate is the name of the annotation that indicates
+	// CordonUntil is the name of the annotation that indicates
 	// the expiration date of rule of this cordon.
-	CordonUntilDate = "app-operator.giantswarm.io/cordon-until"
+	CordonUntil = "app-operator.giantswarm.io/cordon-until"
 )
--- a/service/controller/app/v1/key/key.go
+++ b/service/controller/app/v1/key/key.go
@@ -6,6 +6,7 @@ import (
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/microerror"
 
+	"github.com/giantswarm/app-operator/pkg/annotation"
 	"github.com/giantswarm/app-operator/pkg/label"
 )
 
@@ -61,8 +62,28 @@ func ChartSecretName(customResource v1alpha1.App) string {
 	return fmt.Sprintf("%s-chart-secrets", customResource.GetName())
 }
 
+func CordonReason(customObject v1alpha1.App) string {
+	return customObject.GetAnnotations()[annotation.CordonReason]
+}
+
+func CordonUntil(customObject v1alpha1.App) string {
+	return customObject.GetAnnotations()[annotation.CordonUntilDate]
+}
+
 func InCluster(customResource v1alpha1.App) bool {
 	return customResource.Spec.KubeConfig.InCluster
+}
+
+func IsCordoned(customObject v1alpha1.App) bool {
+	_, reasonOk := customObject.Annotations[annotation.CordonReason]
+	_, untilOk := customObject.Annotations[annotation.CordonUntilDate]
+
+	if reasonOk && untilOk {
+		return true
+	} else {
+		return false
+	}
+
 }
 
 func KubeConfigFinalizer(customResource v1alpha1.App) string {
--- a/service/controller/app/v1/key/key.go
+++ b/service/controller/app/v1/key/key.go
@@ -83,7 +83,6 @@ func IsCordoned(customObject v1alpha1.App) bool {
 	} else {
 		return false
 	}
-
 }
 
 func KubeConfigFinalizer(customResource v1alpha1.App) string {
--- a/service/controller/app/v1/key/key.go
+++ b/service/controller/app/v1/key/key.go
@@ -67,7 +67,7 @@ func CordonReason(customObject v1alpha1.App) string {
 }
 
 func CordonUntil(customObject v1alpha1.App) string {
-	return customObject.GetAnnotations()[annotation.CordonUntilDate]
+	return customObject.GetAnnotations()[annotation.CordonUntil]
 }
 
 func InCluster(customResource v1alpha1.App) bool {
@@ -76,7 +76,7 @@ func InCluster(customResource v1alpha1.App) bool {
 
 func IsCordoned(customObject v1alpha1.App) bool {
 	_, reasonOk := customObject.Annotations[annotation.CordonReason]
-	_, untilOk := customObject.Annotations[annotation.CordonUntilDate]
+	_, untilOk := customObject.Annotations[annotation.CordonUntil]
 
 	if reasonOk && untilOk {
 		return true
--- a/service/controller/app/v1/key/key_test.go
+++ b/service/controller/app/v1/key/key_test.go
@@ -214,7 +214,7 @@ func Test_CordonUntil(t *testing.T) {
 	obj := v1alpha1.App{
 		ObjectMeta: metav1.ObjectMeta{
 			Annotations: map[string]string{
-				annotation.CordonUntilDate: "2019-12-31T23:59:59Z",
+				annotation.CordonUntil: "2019-12-31T23:59:59Z",
 			},
 		},
 	}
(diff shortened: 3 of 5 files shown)


service/controller/app/v1/resource/chart/desired.go
```diff
@@ -106,6 +109,31 @@ func hasSecret(cr v1alpha1.App, appCatalog v1alpha1.AppCatalog) bool {
...
 
+func replacePrefix(from string) string {
+	return strings.ReplaceAll(from, "app-operator.giantswarm.io", "chart-operator.giantswarm.io")
+}
+
+// processAnnotations ensures necessary app-operator.giantswarm.io annotations
+// are copied into the proper prefix in chart CR.
+func processAnnotations(cr v1alpha1.App) map[string]string {
+
+	if cr.GetAnnotations() == nil {
```
tomahawk28: if it's nil at the beginning, return early.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -106,6 +109,31 @@ func hasSecret(cr v1alpha1.App, appCatalog v1alpha1.AppCatalog) bool {
 	return false
 }
 
+func replacePrefix(from string) string {
+	return strings.ReplaceAll(from, "app-operator.giantswarm.io", "chart-operator.giantswarm.io")
```
tomahawk28: To change all prefix, this is efficient.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -106,6 +109,31 @@ func hasSecret(cr v1alpha1.App, appCatalog v1alpha1.AppCatalog) bool {
 	return false
 }
 
+func replacePrefix(from string) string {
+	return strings.Replace(from, "app-operator.giantswarm.io", "chart-operator.giantswarm.io", 1)
```
tomahawk28: Used `ReplaceAll` but failed in CircleCI that function was introduced from 1.12.

pkg/annotation/annotation.go
```diff
@@ -0,0 +1,13 @@
...
+package annotation
+
+const (
+	// CordonReason is the name of the annotation that indicates
+	// the reason of why app-operator should not apply any update on this app CR.
+	CordonReason = "app-operator.giantswarm.io/cordon-reason"
+
+	// CordonUntilDate is the name of the annotation that indicates
+	// the expiration date of rule of this cordon.
+	CordonUntilDate = "app-operator.giantswarm.io/cordon-until"
```
xh3b4sd:
```suggestion
	CordonUntil = "app-operator.giantswarm.io/cordon-until"
```

pkg/annotation/annotation.go
```diff
@@ -0,0 +1,13 @@
+// Package annotation contains common Kubernetes metadata. These are defined in
+// https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md.
+package annotation
+
+const (
+	// CordonReason is the name of the annotation that indicates
+	// the reason of why app-operator should not apply any update on this app CR.
+	CordonReason = "app-operator.giantswarm.io/cordon-reason"
+
+	// CordonUntilDate is the name of the annotation that indicates
```
xh3b4sd:
```suggestion
	// CordonUntil is the name of the annotation that indicates
```

service/controller/app/v1/key/key.go
```diff
@@ -61,10 +62,29 @@ func ChartSecretName(customResource v1alpha1.App) string {
 	return fmt.Sprintf("%s-chart-secrets", customResource.GetName())
 }
 
+func CordonReason(customObject v1alpha1.App) string {
+	return customObject.GetAnnotations()[annotation.CordonReason]
+}
+
+func CordonUntil(customObject v1alpha1.App) string {
+	return customObject.GetAnnotations()[annotation.CordonUntilDate]
```
xh3b4sd:
```suggestion
	return customObject.GetAnnotations()[annotation.CordonUntil]
```

service/controller/app/v1/key/key.go
```diff
@@ -61,10 +62,29 @@ func ChartSecretName(customResource v1alpha1.App) string {
...
+}
+
 func InCluster(customResource v1alpha1.App) bool {
 	return customResource.Spec.KubeConfig.InCluster
 }
 
+func IsCordoned(customObject v1alpha1.App) bool {
+	_, reasonOk := customObject.Annotations[annotation.CordonReason]
+	_, untilOk := customObject.Annotations[annotation.CordonUntilDate]
```
xh3b4sd:
```suggestion
	_, untilOk := customObject.Annotations[annotation.CordonUntil]
```

service/controller/app/v1/key/key_test.go
```diff
@@ -190,6 +192,38 @@ func Test_ChartConfigMapName(t *testing.T) {
...
+	}
+}
+
+func Test_CordonUntil(t *testing.T) {
+	expectedCordonUntil := "2019-12-31T23:59:59Z"
+
+	obj := v1alpha1.App{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: map[string]string{
+				annotation.CordonUntilDate: "2019-12-31T23:59:59Z",
```
xh3b4sd:
```suggestion
				annotation.CordonUntil: "2019-12-31T23:59:59Z",
```

service/controller/app/v1/key/key_test.go
```diff
@@ -204,6 +238,39 @@ func Test_InCluster(t *testing.T) {
...
+		chart          v1alpha1.App
+		expectedResult bool
+	}{
+		{
+			name: "case 0: app cordoned",
+			chart: v1alpha1.App{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotation.CordonReason:    "testing manual upgrade",
+						annotation.CordonUntilDate: "2019-12-31T23:59:59Z",
```
xh3b4sd:
```suggestion
						annotation.CordonUntil: "2019-12-31T23:59:59Z",
```

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -106,6 +109,31 @@ func hasSecret(cr v1alpha1.App, appCatalog v1alpha1.AppCatalog) bool {
...
+		return nil
+	}
+
+	newAnnotations := map[string]string{}
+
+	if key.IsCordoned(cr) {
+		cordonReasonKey := replacePrefix(annotation.CordonReason)
+		newAnnotations[cordonReasonKey] = key.CordonReason(cr)
+
+		cordonUntilKey := replacePrefix(annotation.CordonUntilDate)
```
xh3b4sd:
```suggestion
		cordonUntilKey := replacePrefix(annotation.CordonUntil)
```

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -149,6 +150,90 @@ func Test_Resource_GetDesiredState(t *testing.T) {
 			},
 			errorMatcher: IsExecutionFailed,
 		},
+		{
+			name: "case 2: cordoned app",
+			obj: &v1alpha1.App{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotation.CordonReason:    "migrating job in process",
+						annotation.CordonUntilDate: "2019-12-31T23:59:59Z",
```
xh3b4sd:
```suggestion
						annotation.CordonUntil: "2019-12-31T23:59:59Z",
```

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -106,6 +109,31 @@ func hasSecret(cr v1alpha1.App, appCatalog v1alpha1.AppCatalog) bool {
 	return false
 }
 
+func replacePrefix(from string) string {
```
xh3b4sd: So what are we using this for? Why do we replace the prefix? Which problem does that solve? I saw the comment in `processAnnotations` but I still don't understand it.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -106,6 +109,31 @@ func hasSecret(cr v1alpha1.App, appCatalog v1alpha1.AppCatalog) bool {
...
+	return strings.Replace(from, "app-operator.giantswarm.io", "chart-operator.giantswarm.io", 1)
+}
+
+// processAnnotations ensures necessary app-operator.giantswarm.io annotations
+// are copied into the proper prefix in chart CR.
+func processAnnotations(cr v1alpha1.App) map[string]string {
+
```
xh3b4sd:
```suggestion
```

service/controller/app/v1/resource/chart/desired.go
tomahawk28: All the annotations values we had put in `app` CR has prefix as `app-operator.giantswarm.io/` and we would like to change the value for chart-operator as it should manage that in TC. 

But as you questioned what it solves, it just changed the name but nothing. I have a suggestion maybe we could get rid of prefix at all, just using `giantswarm.io/`. 
so, for example, cordon-until key would be `giantswarm.io/cordon-until`
WDYT? @rossf7

pkg/annotation/annotation.go
```diff
@@ -0,0 +1,13 @@
+// Package annotation contains common Kubernetes metadata. These are defined in
+// https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md.
+package annotation
+
+const (
+	// CordonReason is the name of the annotation that indicates
+	// the reason of why app-operator should not apply any update on this app CR.
```
rossf7:
```suggestion
	// the reason of why app-operator should not apply any update to this app CR.
```

pkg/annotation/annotation.go
```diff
@@ -0,0 +1,13 @@
...
+	CordonReason = "app-operator.giantswarm.io/cordon-reason"
+
+	// CordonUntil is the name of the annotation that indicates
+	// the expiration date of rule of this cordon.
```
rossf7:
```suggestion
	// the expiration date for this cordon rule.
```

service/controller/app/v1/resource/chart/desired.go
rossf7: Damn thanks for trying and we should move to Go 1.12 at some point.

service/controller/app/v1/resource/chart/desired.go
rossf7: I think the approach is fine here. For context, we want to apply the annotations to the app CR not the chart CR. As the app CR is the public interface we expose both internally to release-operator and to customers.

So we want to translate the app-operator.* annotations to chart-operator.* and I think this is a good way to achieve that. We have the same use case for the version label.

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -459,3 +544,42 @@ func Test_processLabels(t *testing.T) {
...
 }
+
+func Test_processAnnotations(t *testing.T) {
+	tests := []struct {
+		name                string
+		obj                 *v1alpha1.App
+		expectedAnnotations map[string]string
+	}{
+		{
+			name: "case 0: basic match",
```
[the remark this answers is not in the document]
rossf7: OK, I thought of another one. You could add some other annotations and make sure they are not returned.

For example, the annotation kubectl uses for the previous state. If the annotations we don't care about are propagated from the app CR to the chart CR then weird / bad things will happen.

service/controller/app/v1/resource/chart/desired.go
xh3b4sd: I like what Ji suggested towards the label naming, but you can decide that within your team.

kopiczko: My 2c is when the annotation is supposed to be set by the user it probably makes sense to skip `*-operator` prefix.

rossf7: @kopiczko Late to the party but this isn't customer facing.

The cordon will be applied by our users via opsctl and programmatically by cluster-operator for the migration from chartconfig to app CRs. So I think using the usual prefix makes sense.

PR #106, opened 2019-07-16
Adding app cordon metrics

tomahawk28:
Toward giantswarm/giantswarm#6078

Adding app cordon metrics for alarming. Most of the structure changes introduced in this PR is almost same as in chart-operator.

### DIFF
--- a/service/collector/app_resource.go
+++ b/service/collector/app_resource.go
@@ -3,7 +3,6 @@ package collector
 import (
 	"context"
 	"fmt"
-	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
 	"strings"
 	"time"
 
@@ -13,6 +12,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
 )
 
 var (
--- a/service/collector/app_resource.go
+++ b/service/collector/app_resource.go
@@ -74,6 +74,25 @@ func NewAppResource(config AppResourceConfig) (*AppResource, error) {
 	return c, nil
 }
 
+// Collect is the main metrics collection function.
+func (c *AppResource) Collect(ch chan<- prometheus.Metric) error {
+	ctx := context.Background()
+
+	c.logger.LogCtx(ctx, "level", "debug", "message", "collecting metrics")
+
+	c.collectAppStatus(ctx, ch)
+
+	c.logger.LogCtx(ctx, "level", "debug", "message", "finished collecting metrics")
+	return nil
+}
+
+// Describe emits the description for the metrics collected here.
+func (c *AppResource) Describe(ch chan<- *prometheus.Desc) error {
+	ch <- appDesc
+	ch <- appCordonExpireTimeDesc
+	return nil
+}
+
 func (c *AppResource) collectAppStatus(ctx context.Context, ch chan<- prometheus.Metric) {
 	r, err := c.g8sClient.ApplicationV1alpha1().Apps("").List(metav1.ListOptions{})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


service/collector/app_resource.go
```diff
@@ -13,8 +16,8 @@ import (
 )
 
 var (
-	appDesc *prometheus.Desc = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, "", "app_info"),
+	appDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, "app", "info"),
```
tomahawk28: Using subsystem as `app` is a better approach I think

service/collector/app_resource_test.go
```diff
@@ -0,0 +1,49 @@
+package collector
+
+import (
+	"reflect"
+	"testing"
+	"time"
+)
+
+func Test_convertToTime(t *testing.T) {
```
tomahawk28: copy same unittests from chart-operator

tomahawk28:
@xh3b4sd I'm looking into aws-operator collector structure, and I see that the difference between collectForAccount and collectAppStatus are having a goroutine or not but not interface as it is. is that right?  https://github.com/giantswarm/aws-operator/blob/027a12a38d51996f9e39ec8d9ded2ed5c0066061/service/collector/trusted_advisor.go#L135

xh3b4sd:
Sorry I did not look closely enough. I see that there is this collector set thingy. This is nice. Let's fix the method order though and then `collectAppStatus` should also have proper error handling. The collector interface supports errors for a reason.

tomahawk28:
@xh3b4sd Thanks, I just got figure it out what was wrong with error handling, PTAL!

service/collector/app_resource.go
```diff
@@ -79,24 +113,38 @@ func (c *AppResource) collectAppStatus(ctx context.Context, ch chan<- prometheus
...
+	if len(split) == 0 {
+		return time.Time{}, microerror.Maskf(invalidExecutionError, "%#q must have at least one item in order to collect metrics for the cordon expiration", datetime)
+	}
 
-	c.logger.LogCtx(ctx, "level", "debug", "message", "finished collecting metrics")
-	return nil
-}
+	t, err := time.Parse(layout, split[0])
+	if err != nil {
+		return time.Time{}, microerror.Maskf(invalidExecutionError, "unavailable to %#q parsing: %#v", split[0], err.Error())
```
rossf7:
```suggestion
		return time.Time{}, microerror.Maskf(invalidExecutionError, "parsing timestamp %#q failed: %#v", split[0], err.Error())
```

service/collector/app_resource.go
```diff
@@ -79,24 +113,38 @@ func (c *AppResource) collectAppStatus(ctx context.Context, ch chan<- prometheus
...
+			prometheus.GaugeValue,
+			float64(t.Unix()),
+			key.AppName(app),
+		)
+	}
+	return nil
+}
 
-	c.logger.LogCtx(ctx, "level", "debug", "message", "collecting metrics")
+func convertToTime(datetime string) (time.Time, error) {
```
rossf7:
```suggestion
func convertToTime(input string) (time.Time, error) {
```

Nitpick but I find the var name also being a "type" a bit confusing.

service/collector/app_resource.go
```diff
@@ -79,24 +113,38 @@ func (c *AppResource) collectAppStatus(ctx context.Context, ch chan<- prometheus
...
+	layout := "2006-01-02T15:04:05"
```
rossf7: In the unit test it has milliseconds `2020-01-02T15:04:05.000`. Is that required? 

We are going to set the annotation in opsctl but I'd like to be clearer on the datetime format.

service/collector/app_resource.go
```diff
@@ -79,24 +113,38 @@ func (c *AppResource) collectAppStatus(ctx context.Context, ch chan<- prometheus
...
 
-	c.collectAppStatus(ctx, ch)
+	split := strings.Split(datetime, ".")
```
rossf7: Sorry more questions. Why is `.` the separator here?

service/collector/app_resource_test.go
```diff
@@ -0,0 +1,49 @@
...
+		{
+			name:     "case 1: normal timestamp parsing",
+			datetime: "2019-12-31T23:59:59.000",
+			expected: expectedTime,
+		},
+		{
+			name:         "case 2: parsing error since unknown ",
+			datetime:     "2019-12-31T23:59:59Z",
+			errorMatcher: IsInvalidExecution,
+		},
```
rossf7: I think some more failure cases would be good. e.g. 2019-13-31T23:59:59Z

service/collector/app_resource.go
tomahawk28: This is also from cert-operator that checking the timestamps in secret. AFAIK this is to split millisecond parts from `datetime`. Most of cases we get the timestamps as `2019-12-31T23:59:59.000` but we don't need that part of milliseconds.

rossf7: OK cool that makes sense. For the `cordon-until` we don't care about milliseconds so good to strip them. 👍

service/collector/app_resource.go
tomahawk28: As we discussed below, we would strip milliseconds part.

PR #108, opened 2019-07-18
Removing kubeconfig finaliser

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/6522

We decided to remove this finalizer on the kubeconfig secret resource, and we would implement another logic to avoid reconciling app CRs against soon-deleting clusters. 

Before we remove `kubeconfigfinaliser` resource, we would like to clean up the remaining finaliser.

### DIFF
--- a/service/controller/app/v1/resource/kubeconfigfinalizer/create.go
+++ b/service/controller/app/v1/resource/kubeconfigfinalizer/create.go
@@ -53,18 +53,17 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	finalizerTag := key.KubeConfigFinalizer(cr)
 
 	if !contains(kubeConfig.Finalizers, finalizerTag) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting finalizer for kubeconfig %#q in namespace %#q", name, namespace))
-
-		kubeConfig.Finalizers = append(kubeConfig.Finalizers, finalizerTag)
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer already clear for kubeconfig secret %#q in namespace %#q", name, namespace))
+	} else {
+		// TODO: Before we remove this resource, let's delete all finalizers that we put on kubeconfig secrets
+		kubeConfig.Finalizers = filter(kubeConfig.Finalizers, finalizerTag)
 
 		_, err := r.k8sClient.CoreV1().Secrets(namespace).Update(kubeConfig)
 		if err != nil {
 			return microerror.Mask(err)
 		}
 
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer set for kubeconfig %#q in namespace %#q", name, namespace))
-	} else {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer already set for kubeconfig secret %#q in namespace %#q", name, namespace))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer cleared for kubeconfig %#q in namespace %#q", name, namespace))
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


service/controller/app/v1/resource/kubeconfigfinalizer/create.go
```diff
@@ -53,18 +53,17 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	finalizerTag := key.KubeConfigFinalizer(cr)
 
 	if !contains(kubeConfig.Finalizers, finalizerTag) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting finalizer for kubeconfig %#q in namespace %#q", name, namespace))
-
-		kubeConfig.Finalizers = append(kubeConfig.Finalizers, finalizerTag)
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer already clear for kubeconfig secret %#q in namespace %#q", name, namespace))
+	} else {
+		// TODO: Before we remove this resource, let's delete all finalizers that we put on kubeconfig secrets
```
tomahawk28: TODO is okay in here? just for a sign that I need to do it by today 😆

rossf7: If this is a short lived TODO can we just do it before we merge this?

tomahawk28: I didn't make it clear; What I would like to put in TODO list is removing this resource from app-operator.

rossf7: Sorry @tomahawk28 I'm still confused here. How will the finalizers be removed?

Also I'm fine with adding a TODO if you think its necessary. But please follow the process and create an issue. https://github.com/giantswarm/fmt/blob/master/go/todo_comments.md

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
```diff
@@ -53,18 +53,17 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	finalizerTag := key.KubeConfigFinalizer(cr)
 
 	if !contains(kubeConfig.Finalizers, finalizerTag) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting finalizer for kubeconfig %#q in namespace %#q", name, namespace))
-
-		kubeConfig.Finalizers = append(kubeConfig.Finalizers, finalizerTag)
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer already clear for kubeconfig secret %#q in namespace %#q", name, namespace))
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer already unset for kubeconfig secret %#q in namespace %#q", name, namespace))
```

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
```diff
@@ -53,18 +53,17 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
 
 		_, err := r.k8sClient.CoreV1().Secrets(namespace).Update(kubeConfig)
 		if err != nil {
 			return microerror.Mask(err)
 		}
 
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer set for kubeconfig %#q in namespace %#q", name, namespace))
-	} else {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer already set for kubeconfig secret %#q in namespace %#q", name, namespace))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finalizer cleared for kubeconfig %#q in namespace %#q", name, namespace))
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("removed finalizer for kubeconfig secret %#q in namespace %#q", name, namespace))
```

I think this is clearer and aligned with the other log entry. I think there should also be a `removing finalizer for kubeconfig secret` log entry.

service/controller/app/v1/resource/kubeconfigfinalizer/create.go
rossf7: OK I think I get this now. So we merge this to remove the finalizers. After that we can do a follow up PR to remove the resource. Is that correct?

tomahawk28:
Tested on ginger.

PR #109, opened 2019-07-23
Remove kubeconfigfinalizer resource

tomahawk28:
Toward giantswarm/giantswarm#6522

As we are going to put another logic on cluster deletion (checking deletion timestamp on cluster namespace`), we don't need to use `kubeconfigfinalizer` resource.


tomahawk28:
I checked that `gollum` installations but it stop during process `configmap` resource since there's an error ( not a bug) but it didn't continue to other resources, is it expected?

tomahawk28:
Okay, it was due to app CR had been modified manually after deployment. I check `ginger` and it didn't have any finalizers remain at there.

PR #110, opened 2019-07-23
Stop reconcile when namespace is going to be deleted soon.

tomahawk28:
Toward giantswarm/giantswarm#6522

The operator will remove all finalizers in `app` CR in case of namespace contains that `app` CR is going to be deleted soon. 
If this affected, cluster namespace would not hang in `Terminating` status.


service/controller/app/v1/resource/configmap/current_test.go
```diff
@@ -132,6 +132,8 @@ func Test_Resource_GetCurrentState(t *testing.T) {
 				objs = append(objs, tc.configMap)
 			}
 
+			objs = append(objs, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "kube-system"}})
```
tomahawk28: Adding default namespace resource for unittest passing.

service/controller/app/v1/resource/configmap/current.go
```diff
@@ -20,6 +21,31 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
+	ns, err := r.k8sClient.CoreV1().Namespaces().Get(key.Namespace(cr), metav1.GetOptions{})
```
tomahawk28: Whenever cluster got deleted, we will cancel all reconciliation since network issue in app-operator, For the next loop in reconciliation, we would start from `configmap` resource. 
So I think checking the namespace condition only in `configmap` makes sense, WDYT?

service/controller/app/v1/resource/configmap/current.go
```diff
@@ -20,6 +21,31 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+		return nil, microerror.Mask(err)
+	}
+
+	if ns.GetDeletionTimestamp() != nil {
+		currentCR, err := r.g8sClient.ApplicationV1alpha1().Apps(cr.Namespace).Get(cr.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		currentCR.Finalizers = []string{}
```
tomahawk28: cleaning finalizers as empty, so not to call app-operator for deletion event again.

xh3b4sd: That fucks with all operators adding their own finalizers. So that is definitely not what we want. What a resource should do is to remove its own finalizer if there is anything like that. If there is a project wide finalizer there should be a separate resource managing it. I think when you only want to stop reconciling on deletion the resources should be canceled here, and then each should do that on their own.

rossf7: 100% i discussed that with @tomahawk28 in slack. Ji is going to post in #sig-operator to discuss and find a solution that works.

tomahawk28:
Gt an error as below from circleCI, should add namespace in RBAC.
```{"caller":"github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/resource/wrapper/retryresource/crud_resource_ops_wrapper.go:71","controller":"app-operator","event":"update","function":"GetCurrentState","level":"warning","loop":"1","message":"retrying due to error","object":"/apis/application.giantswarm.io/v1alpha1/namespaces/giantswarm/apps/kubernetes-test-app-chart","resource":"configmapv1","stack":"[{/go/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/resource/wrapper/retryresource/crud_resource_ops_wrapper.go:64: } {/go/src/github.com/giantswarm/app-operator/service/controller/app/v1/resource/configmap/current.go:25: } {namespaces \"giantswarm\" is forbidden: User \"system:serviceaccount:giantswarm:app-operator\" cannot get namespaces in the namespace \"giantswarm\"}]","time":"2019-07-24T10:50:43.209266+00:00","underlyingResource":"configmapv1","version":"512"}```

**Comment by tomahawk28** [comment]
@giantswarm/team-batman , @giantswarm/sig-operator; Would you mind to review this again, I applied all the changes we agreed except a helper function; It's hard to find a default location for that since I need k8sClient to fetch namespace.

File: service/controller/app/v1/resource/chart/create_test.go [hunk 4]
```diff
@@ -71,6 +72,7 @@ func Test_Resource_newCreateChange(t *testing.T) {
 
 	c := Config{
 		G8sClient: fake.NewSimpleClientset(),
```
xh3b4sd:
```suggestion
		G8sClient: g8sfake.NewSimpleClientset(),
```

service/controller/app/v1/resource/chart/current.go
```diff
@@ -19,6 +20,20 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
+	ns, err := r.k8sClient.CoreV1().Namespaces().Get(cr.Namespace, metav1.GetOptions{})
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	if ns.GetDeletionTimestamp() != nil {
```
xh3b4sd: Brainwave, what we can do is to have a resource that gathers information in the controller context which other resources then make use of. So we could have a resource to fetch the namespace and put it into the controller context or directly have it put the boolean. So this here would then only be something like this. 

```
if cc.Status.TenantCluster.Namespace.IsDeleting {
```

We make use of this concept in a lot of places already. See for instance the following. 

* https://github.com/giantswarm/aws-operator/tree/master/service/controller/clusterapi/v29/resource/accountid
* https://github.com/giantswarm/aws-operator/tree/master/service/controller/clusterapi/v29/resource/awsclient
* https://github.com/giantswarm/aws-operator/tree/master/service/controller/clusterapi/v29/resource/region

tomahawk28: Thanks for the reference! This is great, I think new resource should be located in the first place in the resource_set, right?

service/controller/app/v1/resource/namespace/delete.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
```
tomahawk28: I think we don't need to take care of `EnsureCreated` case, does it?

service/controller/app/v1/resource/chart/current.go
```diff
@@ -26,6 +27,15 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
+	if cc.Status.TenantCluster.IsDeleting {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("namespace %#q is going to be deleted, no need to reconcile resource", cr.Namespace))
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("namespace %#q is being deleted, no need to reconcile resource", cr.Namespace))
```

Nit but I prefer this.

service/controller/app/v1/resource/namespace/delete.go
rossf7: Yes I don't think we need to implement EnsureCreated.

But we have a naming collision here. 😢 

For bootstrapping chart-operator in tenant clusters we need a namespace resource which will create the giantswarm namespace in new tenant clusters.

Can we rename this to `appnamespace`? Please also add godoc so the purpose of this resource is clear.

tomahawk28: @rossf7 Can you suggest where you want to see the godoc description? I'm missing in here.

service/controller/app/v1/resource/appnamespace/delete.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
```
rossf7:
```suggestion
// EnsureDeleted checks whether the namespace this app CR belongs to
// is being deleting currently.
func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
```

tomahawk28:
I tested on ginger and found a problem in the initial reconciliation step. 
https://gigantic.slack.com/archives/C76JX6YLQ/p1563979865085400

The operator could not find `kubeconfig` secret resource since clusters are being deleted at that moment. 
https://github.com/giantswarm/app-operator/blob/515e4b51bc05b9bdf0c527b899f18b9ef3cad47d/service/controller/app/v1/resource_set.go#L229-L232 

So instead of adding new resource (I already did though), I suggest adding namespace checking mechanism into `initCtxFunc` and leave two context variables as nil only if `cc.Status.TenantCluster.IsDeleting` is `true`
WDYT?

rossf7:
@tomahawk28 Good catch and I'd missed this when we discussed the fix.

We added the initCtxFunc because we want to fetch the kubeconfig secret once per loop. Currently in cluster-operator we do this lots of times and its wasteful in terms of API calls and slows things down.

So I think the solution you're proposing is the right one. It's OK to leave the values as nil because we first check cc.Status.TenantCluster.IsDeleting and cancel the resource if it is true.

tomahawk28:
@rossf7 Thanks, the only thing holding back the idea is I should use `config.K8sClient` variable directly from this `initCtxFunc`. I'm worried about whether this is a design flaw.

rossf7:
`config.K8sClient` is safe to use. It is the client for the control plane cluster where the operator runs so it's always available.

tomahawk28:
Testing again on ginger.

tomahawk28:
Got cleaned up ```D 07/24 15:46:15 /apis/application.giantswarm.io/v1alpha1/namespaces/gb97o/apps/kubernetes-elastic-stack-elastic-logging configmapv1.GetCurrentState namespace `gb97o` is being deleted, no need to reconcile resource | app-operator/service/controller/app/v1/resource/configmap/current.go:31 | controller=app-operator | event=delete | loop=0 | version=140033100``` 🎉 
Is it safe to merge? 
Since there are lots of changes after you two had approved. @xh3b4sd , @rossf7

service/controller/app/v1/resource_set.go
```diff
@@ -209,25 +210,44 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		var k8sClient kubernetes.Interface
 		var g8sClient versioned.Interface
 
-		restConfig, err := kubeConfig.NewRESTConfigForApp(ctx, cr)
-		if err != nil {
-			return nil, microerror.Mask(err)
-		}
+		var isDeleting bool
```
xh3b4sd: I see you took the shortcut. Things like this are the reason why I would like to get rid of the init ctx func. It is regularly abused like that. I made an effort to remove all logic from the init ctx func in the `aws-operator` already. It would be nice if we could not add more logic here. I think we should have a resource for this. I do not understand the problem that you had with the resource you already implemented. 

We could have a resource for the namespace deletion and a resource for the kube config magic. Then the kube config resource uses the information about the namespace deletion to decide what to do, just like all other resources do now. Then the init ctx func would be empty I think. In `aws-operator` we have the same approach with AWS clients that we gather in a separate resource which puts the clients into the controller context.

rossf7: This has nothing to do with shortcuts. So I'd appreciate it if we don't use that using that kind of language in reviews.

We are trying to find a solution to a problem that exists in production and has been reported by 3 separate people just this week.

If we want to deprecate this functionality why is there no deprecation notice in operatorkit? How are Ji and I supposed to know this? To me that is a communication problem.

https://github.com/giantswarm/operatorkit/blob/1a065237841c6c15cc0a3005d3a29a1c8d490b95/framework/framework.go#L19-L25

If this solves the issue in production I'd like us to merge it. We will then do a follow up PR to address these concerns and so we can deprecate this functionality from operatorkit.

cc @puja108

xh3b4sd: It was not my intention to trigger your frustration here. I approved the PR twice already and do not block anything. So I assume I am not the root cause for whatever overwhelms you now. Do what you think is right and let me know when I can help with anything. I only provided suggestions as always, and not because I want to control the progress here, but because I have been asked for it.

xh3b4sd: I put a card in the SIG Operator Board to document the init ctx func magic. You are right that this is likely not be well documented.

rossf7: Yes I got frustrated because I don't think we were taking a shortcut. So I wanted to push back a bit on that but I think my reaction was an overreaction and not helpful. So I'm sorry about that. ❤️ 

I like the design you're proposing and as always I really appreciate getting feedback that lets us improve what we do.

tomahawk28: After merging this, I will open another PR to migrate from initCtxFunc to another resource. Thanks for your input!

rossf7: Great thanks @tomahawk28 much appreciated and sorry I let the discussion get heated in your PR. ❤️

PR #111, opened 2019-07-25
Move out from initCtxFunc in controller

tomahawk28:
From the previous PR, I got an input stating that initCtxFunc would be deprecated eventually. 

@xh3b4sd's Comment link: https://github.com/giantswarm/app-operator/pull/110#discussion_r306899415

So I'm moving to the separate resource with proper ordering in the resource set.


service/controller/app/v1/resource/appcatalog/delete.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+// EnsureDeleted gets the appCatalog CR specified in the provided app CR.
```
tomahawk28: I don't think we need appCatalog in deletion event, but some resources refer it inside `desired.go`. @rossf7 found out we need to pass it through :)

service/controller/app/v1/resource/appcatalog/resource.go
```diff
@@ -11,62 +11,71 @@ import (
...
+
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
 	catalogName := key.CatalogName(customResource)
 
-	a.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("looking for appCatalog %#q in namespace %#q", catalogName, a.watchNamespace))
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("looking for appCatalog %#q in namespace %#q", catalogName, "default"))
```
tomahawk28: Since this is Cluser-scoped resource, no need to put `watchNamespace`.

tomahawk28:
Tested on ginger, cluser creation & cluster deletion.

service/controller/app/v1/resource/appcatalog/resource.go
```diff
@@ -11,62 +11,71 @@ import (
...
 
-// GetCatalogForApp gets the appCatalog CR specified in the provided app CR.
-func (a *AppCatalog) GetCatalogForApp(ctx context.Context, customResource v1alpha1.App) (*v1alpha1.AppCatalog, error) {
+func (*Resource) Name() string {
+	return Name
+}
+
+// getCatalogForApp gets the appCatalog CR specified in the provided app CR.
+func (r *Resource) getCatalogForApp(ctx context.Context, customResource v1alpha1.App) error {
+
```
xh3b4sd:
```suggestion
```

service/controller/app/v1/resource/clients/resource.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// addClientsToContext gets the k8sClient and g8sClient instance for Tenant cluster access.
+func (r *Resource) addClientsToContext(ctx context.Context, cr v1alpha1.App) error {
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if !cc.Status.TenantCluster.IsDeleting {
```
xh3b4sd: We can do an early return when we negate the expression.

service/controller/app/v1/resource/clients/resource.go
```diff
@@ -0,0 +1,99 @@
...
+
+		if cc.K8sClient == nil {
+			k8sClient, err := kubernetes.NewForConfig(restConfig)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+			cc.K8sClient = k8sClient
+		}
+	}
+	return nil
```
xh3b4sd:
```suggestion

	return nil
```

service/controller/app/v1/resource_set.go
```diff
@@ -109,6 +105,18 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		}
 	}
 
+	var clientsResource controller.Resource
+	{
+		c := clients.Config{
+			K8sClient: config.K8sClient,
+			Logger:    config.Logger,
+		}
+		clientsResource, err = clients.New(c)
```
xh3b4sd:
```suggestion

		clientsResource, err = clients.New(c)
```

service/controller/app/v1/resource/clients/resource.go
tomahawk28: That is much cleaner before, Thanks for the input.

service/controller/app/v1/resource/appcatalog/resource.go
```diff
@@ -11,62 +11,70 @@ import (
...
+func (r *Resource) getCatalogForApp(ctx context.Context, customResource v1alpha1.App) error {
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
 	catalogName := key.CatalogName(customResource)
 
-	a.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("looking for appCatalog %#q in namespace %#q", catalogName, a.watchNamespace))
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("looking for appCatalog %#q in namespace %#q", catalogName, "default"))
```
rossf7:
```suggestion
	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("looking for appCatalog %#q", catalogName))
```

I think we can just drop the namespace from the logging. Since its cluster scoped I think including default is more confusing.

service/controller/app/v1/resource/appnamespace/create.go
```diff
@@ -0,0 +1,7 @@
+package appnamespace
+
+import "context"
+
+func (*Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
```
rossf7: Let's have a comment explaining why we don't need to implement `EnsureCreated`.

tomahawk28: I will put the comment inside the function.

rossf7: But will it be godoc? Is this for the internal function?

I'd like to have it here too. As when I'm trying to figure out what a resource does this is where I look first.

tomahawk28: I will put a godoc as ```// EnsureCreated does not need to fetch the status of namespace since operator only
// need to take consider of deletion event.``` WDYT?

rossf7: Thanks and can we go with this?

> EnsureCreated is not implemented because we only need to consider the status of the namespace on deletion events.

service/controller/app/v1/resource/appcatalog/resource.go
```diff
@@ -11,62 +11,70 @@ import (
...
 	catalogName := key.CatalogName(customResource)
 
-	a.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("looking for appCatalog %#q in namespace %#q", catalogName, a.watchNamespace))
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("looking for appCatalog %#q", catalogName))
 
-	appCatalog, err := a.g8sClient.ApplicationV1alpha1().AppCatalogs().Get(catalogName, metav1.GetOptions{})
+	appCatalog, err := r.g8sClient.ApplicationV1alpha1().AppCatalogs().Get(catalogName, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		return nil, microerror.Maskf(notFoundError, "appCatalog %#q in namespace %#q", catalogName, "default")
+		return microerror.Maskf(notFoundError, "appCatalog %#q in namespace %#q", catalogName, "default")
```
rossf7: Lets remove the namespace here too.

PR #114, opened 2019-07-27
Minor changes to godoc

rossf7:


service/controller/app/v1/resource/clients/create.go
```diff
@@ -8,7 +8,8 @@ import (
 	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
 )
 
-// EnsureCreated gets the k8sClient and g8sClient instance for Tenant cluster access.
+// EnsureCreated adds g8s and k8s clients to the controller context based on the
+// kubeconfig settings for the app CR.
```
rossf7: I updated this as its not always tenant cluster access. For aws-operator etc it will be the control plane cluster because inCluster is set to true in the kubeconfig spec.

service/controller/app/v1/resource/appnamespace/delete.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 // EnsureDeleted checks whether the namespace this app CR belongs to
-// is being deleting currently.
+// is currently being deleted.
```
rossf7: Just a 🇺🇸 nit.

rossf7:
Some minor godoc stuff I noticed while making other changes.

PR #115, opened 2019-07-27
Extract getting configmap data to values service

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/6639

This is prep for extending app-operator to bootstrap chart-operator.

We need to provide values when installing chart-operator. So this moves the logic from the configmap resource to the values service. This will be used in the next PR.

### DIFF
--- a/service/controller/app/v1/key/key.go
+++ b/service/controller/app/v1/key/key.go
@@ -62,21 +62,21 @@ func ChartSecretName(customResource v1alpha1.App) string {
 	return fmt.Sprintf("%s-chart-secrets", customResource.GetName())
 }
 
-func CordonReason(customObject v1alpha1.App) string {
-	return customObject.GetAnnotations()[annotation.CordonReason]
+func CordonReason(customResource v1alpha1.App) string {
+	return customResource.GetAnnotations()[annotation.CordonReason]
 }
 
-func CordonUntil(customObject v1alpha1.App) string {
-	return customObject.GetAnnotations()[annotation.CordonUntil]
+func CordonUntil(customResource v1alpha1.App) string {
+	return customResource.GetAnnotations()[annotation.CordonUntil]
 }
 
 func InCluster(customResource v1alpha1.App) bool {
 	return customResource.Spec.KubeConfig.InCluster
 }
 
-func IsCordoned(customObject v1alpha1.App) bool {
-	_, reasonOk := customObject.Annotations[annotation.CordonReason]
-	_, untilOk := customObject.Annotations[annotation.CordonUntil]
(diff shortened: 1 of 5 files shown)


service/controller/app/v1/key/key.go
```diff
@@ -62,21 +62,21 @@ func ChartSecretName(customResource v1alpha1.App) string {
 	return fmt.Sprintf("%s-chart-secrets", customResource.GetName())
 }
 
-func CordonReason(customObject v1alpha1.App) string {
-	return customObject.GetAnnotations()[annotation.CordonReason]
+func CordonReason(customResource v1alpha1.App) string {
```
tomahawk28: <3

service/controller/app/v1/values/configmap.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// We get the user level values if configured and merge them.
+	if key.UserConfigMapName(app) != "" {
+		userData, err := v.getUserConfigMapForApp(ctx, app)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		// Config is merged again and in case of intersecting values the user
```
tomahawk28: Can we summarize this godocs and put it into the first place?

rossf7:
@tomahawk28 There are unit tests in the desired state of the configmap resource. The tests are quite big so I'd prefer not to duplicate them. What about if I move them to the values service?

Good shout on the godoc I'll see if it can be improved.




https://github.com/giantswarm/app-operator/blob/master/service/controller/app/v1/resource/configmap/desired_test.go

service/controller/app/v1/values/configmap.go
rossf7: Done.

service/controller/app/v1/values/configmap_test.go
```diff
@@ -0,0 +1,358 @@
...
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/micrologger/microloggertest"
+	"github.com/google/go-cmp/cmp"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgofake "k8s.io/client-go/kubernetes/fake"
+)
+
+func Test_MergeConfigMapData(t *testing.T) {
```
rossf7: This basically copy pasta from the configmap desired state test. In the next PR I will simplify the UT there.

PR #119, opened 2019-08-06
Add helmclient on controllercontext

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/6639

This is preparing step we move `chartoperator` into app-operator. 
We need helmclient instance to setup chart-operator at TC.

### DIFF
--- a/service/controller/app/v1/resource/clients/create.go
+++ b/service/controller/app/v1/resource/clients/create.go
@@ -21,10 +21,5 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	err = r.addHelmClientToContext(ctx, cr)
-	if err != nil {
-		return microerror.Mask(err)
-	}
-
 	return nil
 }
--- a/service/controller/app/v1/resource/clients/create.go
+++ b/service/controller/app/v1/resource/clients/create.go
@@ -21,5 +21,10 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	err = r.addHelmClientToContext(ctx, cr)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
 	return nil
 }
--- a/service/controller/app/v1/resource/clients/resource.go
+++ b/service/controller/app/v1/resource/clients/resource.go
@@ -2,10 +2,10 @@ package clients
 
 import (
 	"context"
-	"github.com/giantswarm/helmclient"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
+	"github.com/giantswarm/helmclient"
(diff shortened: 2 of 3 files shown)


service/controller/app/v1/resource/clients/resource.go
```diff
@@ -66,6 +78,10 @@ func (r *Resource) addClientsToContext(ctx context.Context, cr v1alpha1.App) err
 		return nil
 	}
 
+	if cc.G8sClient != nil && cc.K8sClient != nil {
```
tomahawk28: Return early rather than propagate two indentations blocks.

xh3b4sd: Why should the clients be set at this point here? Is the resource called multiple times within the same reconciliation loop? That should not be the case and whenever the resource is called it should mutate the context IMO. Anything else would be confusing and complex.

service/controller/app/v1/resource/clients/create.go
```diff
@@ -21,5 +21,10 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	err = r.addHelmClientToContext(ctx, cr)
```
xh3b4sd: Why don't we add the client in the method that already exists? There are clients added to the context already.

service/controller/app/v1/resource/clients/resource.go
tomahawk28: That's right, I should not check `nil` at this point, Thanks to pointing out.

service/controller/app/v1/resource/clients/create.go
tomahawk28: Not sure which clients in the method you are meaning, you mean `helmClient`?

xh3b4sd: There is `addClientsToContext` already. It adds the `g8sClient` and the `k8sClient` to the context. Why do we need another method to add the helm client to the context? Why can't we add the helm client to the context in `addClientsToContext`?

tomahawk28: I thought I need to differentiate two clients and helm client, but seems that's not the good practice, I'm merging two functions as you said, Thanks a lot!

tomahawk28: @xh3b4sd PTAL :)

service/controller/app/v1/resource/clients/resource.go
```diff
@@ -84,21 +97,31 @@ func (r *Resource) addClientsToContext(ctx context.Context, cr v1alpha1.App) err
...
-		g8sClient, err := versioned.NewForConfig(restConfig)
-		if err != nil {
-			return microerror.Mask(err)
-		}
-		cc.G8sClient = g8sClient
+	g8sClient, err := versioned.NewForConfig(restConfig)
+	if err != nil {
+		return microerror.Mask(err)
 	}
+	cc.G8sClient = g8sClient
```
xh3b4sd: When we could now have this code within a scope block I would be so happy. 🙈

PR #120, opened 2019-08-06
Moving chartoperator resource into app-operator

tomahawk28:
Toward giantswarm/giantswarm#6639

Moving chart-operator from cluster-operator to app-operator. 

This resource is not implemented CRUD operations since we would like to install chart-operator when TC is a new cluster and don't have `chart-operator` yet.

We would reconcile the chart-operator in TC by using `app` resource in cluster-operator.


integration/setup/config.go
```diff
@@ -14,8 +14,7 @@ import (
 )
 
 const (
-	namespace       = "giantswarm"
-	tillerNamespace = "kube-system"
+	namespace = "giantswarm"
```
tomahawk28: Don't need to have a separate namespace for tillerNamespace. Make it simple

helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -70,6 +70,18 @@ rules:
...
+- apiGroups:
+    - ""
+  resources:
+    - pods
+  verbs:
+    - list
+- apiGroups:
+    - ""
+  resources:
+    - pods/portforward
```
tomahawk28: Without these permission, operator could not install `chart-operator` inside TC.

service/controller/app/v1/resource/chartoperator/delete.go
```diff
@@ -0,0 +1,7 @@
+package chartoperator
+
+import "context"
+
+func (r Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
```
tomahawk28: Should we delete chart-operator if no `apps` are there?

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -0,0 +1,153 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/helm/pkg/helm"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+const (
+	Name = "chartoperatorv1"
+
+	defaultClusterDNSIP    = "172.31.0.10"
```
tomahawk28: Adding a default value, but we surely have this value for newer cluster.

integration/setup/setup.go
```diff
@@ -60,13 +59,6 @@ func installResources(ctx context.Context, config Config) error {
 		}
 	}
 
-	{
```
tomahawk28: We are installing chart-operator in the resource level, so don't need to care it from e2e setup.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// Check whether TC has chart-operator deployment yet.
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding chartoperator release %#q in tenant cluster", chartOperatorRelease))
```
tomahawk28: This file is a bit crowed with logs, but please let me know if coding style needs to change.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
```
xh3b4sd:
```suggestion
```

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,46 @@
...
+				return microerror.Mask(err)
+			}
+
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("installed chartoperator release %#q in tenant cluster", chartOperatorRelease))
+		} else if err != nil {
+			return microerror.Mask(err)
+		} else {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found chartoperator release %#q", chartOperatorRelease))
+		}
+	}
```
xh3b4sd:
```suggestion
	}

```

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// Check whether TC has chart-operator deployment yet.
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding chartoperator release %#q in tenant cluster", chartOperatorRelease))
+		_, err := cc.HelmClient.GetReleaseContent(ctx, chartOperatorRelease)
+		if helmclient.IsReleaseNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("no chartoperator release %#q in tenant cluster", chartOperatorRelease))
```
xh3b4sd: I would say let's either drop the log or align it with the one above so that it reads "did not find chart-opertor ...".

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,46 @@
...
+
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("installaing chartoperator release %#q in tenant cluster", chartOperatorRelease))
```
xh3b4sd: I would like to change this everywhere. 

```suggestion
			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("installaing chart-operator release %#q in tenant cluster", chartOperatorRelease))
```

service/controller/app/v1/resource/chartoperator/delete.go
xh3b4sd: The resource in `cluster-operator` does not implement deletion but when there is an app and it is deleted whatever is managed with that app should be deleted IMO. Just to ensure app creation and deletion works.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -0,0 +1,153 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/helm/pkg/helm"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+const (
+	Name = "chartoperatorv1"
```
xh3b4sd:
```suggestion
	Name = "chartoperatorv1"
)
```

service/controller/app/v1/resource/chartoperator/resource.go
xh3b4sd:
```suggestion
const (
	defaultClusterDNSIP    = "172.31.0.10"
```

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -0,0 +1,153 @@
...
+		}()
+	}
+
+	var clusterDNSIP string
+	{
+		name := key.ClusterValuesConfigMapName(cr)
+		cm, err := r.k8sClient.CoreV1().ConfigMaps(cr.Namespace).Get(name, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "warning", "message", fmt.Sprintf("no cluster-value %#q in control plane, operator will use default clusterDNSIP value", name))
+			clusterDNSIP = defaultClusterDNSIP
```
xh3b4sd: Why should that actually ever happen? And in case it ever happens, why can we set the value statically? If I am not mistaken we derive the DNS IP from the Cluster IP range and this is configured via the installations repo. So I assume it can be different. I would either drop the defaulting here as I do not see why we need it, but if I am missing something the defaulting should not be done based on some static value. Or as a third option, what we do in the `cluster-operator` is not necessary but then we should clean up the mess there.

service/controller/app/v1/resource/chartoperator/spec.go
```diff
@@ -0,0 +1,23 @@
+package chartoperator
+
+// Values represents the values to be passed to Helm commands related to
+// chart-operator chart.
+type Values struct {
+	ChartResource ChartResource `json:"resource"`
```
xh3b4sd: This was bugging me already in `cluster-operator`. Can we go with the naming as we do in `apiextensions`? 

```suggestion
	ChartResource ValuesChartResource `json:"resource"`
```

service/controller/app/v1/resource_set.go
```diff
@@ -97,6 +103,19 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		}
 	}
 
+	var chartOperatorResource controller.Resource
+	{
+		c := chartoperator.Config{
+			Fs:        config.Fs,
```
xh3b4sd: I would like to call this `FileSystem`.

helm/app-operator-chart/templates/rbac.yaml
rossf7: Ack. That makes sense because app-operator creates a port forward connection using https://github.com/giantswarm/k8sportforward. Once we get rid of tiller we can remove this.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,46 @@
...
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// Check whether TC has chart-operator deployment yet.
```
rossf7:
```suggestion
	// Check whether tenant cluster has a chart-operator helm release yet.
```

Nit but this is more accurate I think as we are using helmclient to interact with the cluster.

service/controller/app/v1/resource/chartoperator/create.go
rossf7: +1 that is aligned with https://github.com/giantswarm/fmt/blob/master/go/log_messages.md#fact-gathering.

service/controller/app/v1/resource/chartoperator/create.go
rossf7: Agree but please fix the `installaing ` typo too! 😄

service/controller/app/v1/resource/chartoperator/delete.go
rossf7: I think `EnsureDeleted` should be a no-op like it is here.

For me, it's a question of which operator is responsible. 

- Every TC needs a chart-operator which is what we are bootstrapping here.
- When a TC gets deleted aws-operator etc are responsible for deleting the TC and chart-operator in the TC goes away.

service/controller/app/v1/resource/chartoperator/resource.go
rossf7: Hmm yes, this is for bootstrapping CoreDNS and it should come from the cluster configmap. 

What happens if this is not set?

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+const (
+	Name = "chartoperatorv1"
+
+	defaultClusterDNSIP    = "172.31.0.10"
+	chartOperatorNamespace = "giantswarm"
+	chartOperatorRelease   = "chart-operator"
+	chartOperatorVersion   = "0.9.0"
```
rossf7: This version should come from the app CR we have created for chart-operator in cluster-operator. We can then remove this hardcoding.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -0,0 +1,153 @@
...
+func (r Resource) Name() string {
+	return Name
+}
+
+func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App, helmClient helmclient.Interface) error {
+	var err error
+
+	var tarballPath string
+	{
+		tarballURL := fmt.Sprintf("https://giantswarm.github.io/giantswarm-catalog/chart-operator-%s.tgz", chartOperatorVersion)
```
rossf7: We should have a key function for this so we have the logic once and reuse it in the chart resource.

service/controller/app/v1/resource/chartoperator/resource.go
rossf7: Yes like I replied before this is for bootstrapping CoreDNS. That is only needed when chart-operator is running in a TC.

If chart-operator is running locally in kind or minikube then there is already a CoreDNS. So yes I think we should drop this.

service/controller/app/v1/resource/chartoperator/spec.go
rossf7: I would actually drop these types completely and use a `map[string]interface{}`. This was a suggestion from Mr @kopiczko and I like it a lot.

service/controller/app/v1/resource_set.go
rossf7: Yes that makes sense to me.

rossf7:
> As a general note, I did refactoring of this resource in cluster-operator. Why don't we reuse it from there? This looks like you wrote in on your own again and I think this is not necessary.

@xh3b4sd The design changes quite a bit. In cluster-operator we use a CRUD resource and implement create and update.

The new design is we use the simple resource and just implement create. There is now a chart CR for chart-operator and it updates itself like draughtsman does. 

@tomahawk28 But if there are improvements in the cluster-operator resource we should try to bring them across.

service/controller/app/v1/resource/chartoperator/delete.go
xh3b4sd: One step back. Why do we have a separate resource for installing `chart-operator` at all? Why isn't it a generic operation to install whatever chart should be installed upon an `App` CR? Why is this special?

service/controller/app/v1/resource/chartoperator/resource.go
tomahawk28: Then it would use default value in here, https://github.com/giantswarm/chart-operator/blob/854a6b727ed86ee05e172d84bc6a03e8e1429ab7/helm/chart-operator-chart/values.yaml#L11

Therefore I can safely remove this value.

service/controller/app/v1/resource/chartoperator/resource.go
tomahawk28: So I found out we don't need to have default value as separate, we keep this value in chart values.

tomahawk28:
> We need to be careful how we deploy this. The code should only be active if there is an app CR for chart-operator. 

@rossf7 I didn't know this, I was thinking about we are actually bootstrapping in here with the initial version (0.9.0 in here), and `app` CR will do a necessary update after all. 
If app-operator awaits for `app` CR for the chart, we should know the object name of `app` CR that would relate to `chart-operator` and that's not the good pattern I would say. 

Why `chartoperator` resource needs to do exactly the same thing that `chart-operator/chart` resource would do anyway?

Moreover, we need to have chart-operator CR for every e2e test by then.

I would like to go with not to wait for `chart-operator` app CR for these reasons, WDYT?

service/controller/app/v1/resource/chartoperator/delete.go
tomahawk28: `app-operator` watches `app CR and create `chart` CR related to that, but first of all, we need an operator that creating `chart` *CRD*. `chartoperator` would bring this chart-operator into the new TC.

xh3b4sd: With you. It is for the bootstrapping.

rossf7:
@tomahawk28 This is looking nice and I think the main thing to resolve is the app CR.

I see what you mean but we can name the app CR `chart-operator` so it's predictable and the same as the helm release.

I'd prefer to have the app CR so we don't need to hardcode the chart-operator version in the operator. We just set the version in the app CR.

If there is no chart-operator then this resource installs it. If its an update the chart resource updates the chart CR and chart-operator updates itself.

It's a good point the e2e test needs to create 2 app CRs but we can reuse the helm chart that creates the test CR.

tomahawk28:
@rossf7 So `EnsureCreated` would install chart-operator in case of all below conditions are met.
  1. TC does not have chart-operator
  2. CP has a `chart-operator` `app` CR
This way, we don't need to care about upgrading the version of chart-operator.

rossf7:
@tomahawk28 This resource handles installing chart-operator when there is no chart-operator.

For updating chart-operator.

- If there is already a chart-operator this resource doesn't do anything.
- The chart resource will process the app CR which updates the chart CR.
- chart-operator processes its own chart CR and updates itself.

I think that works and is my preferred option but there is another approach.

- This resource handles installing and updating chart-operator.
- In that case we need to change the chart resource so it doesn't create a CR for chart-operator.

Either way we need special case logic for the bootstrapping. Does that make sense? If not lets do a call on this.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -30,7 +27,7 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 	}
 
 	config := generateConfig(cr, cc.AppCatalog, r.chartNamespace)
-	tarballURL, err := generateTarballURL(appcatalogkey.AppCatalogStorageURL(cc.AppCatalog), key.AppName(cr), key.Version(cr))
+	tarballURL, err := appcatalogkey.GenerateTarballURL(appcatalogkey.AppCatalogStorageURL(cc.AppCatalog), key.AppName(cr), key.Version(cr))
```
rossf7: Why is this is in `appcatalogkey` and not `key`? 

It's only needed for the `app` controller isn't it? If so it would be good to move it.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	cc, err := controllercontext.FromContext(ctx)
```
rossf7: Nit but newline pls.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// Check whether tenant cluster has a chart-operator helm release yet.
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding chart-operator release %#q in tenant cluster", chartOperatorRelease))
```
rossf7: Newline here too pls.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,46 @@
...
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// Check whether tenant cluster has a chart-operator helm release yet.
```
rossf7: Yes, we only want to bootstrap chart-operator in TCs. But we also need to process app CRs in the CP for VOO. Is this handled?

If not I think we should cancel the resource if InCluster is true. In the CP chart-operator is deployed by draughtsman.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,46 @@
...
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding chart-operator release %#q in tenant cluster", chartOperatorRelease))
+		_, err := cc.HelmClient.GetReleaseContent(ctx, chartOperatorRelease)
+		if helmclient.IsReleaseNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find charto-perator release %#q in tenant cluster", chartOperatorRelease))
```
rossf7:
```suggestion
			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart-operator release %#q in tenant cluster", chartOperatorRelease))
```

integration/test/app/kubeconfig/kubeconfig_test.go
```diff
@@ -60,12 +61,27 @@ func TestAppLifecycleUsingKubeconfig(t *testing.T) {
 	{
 		config.Logger.LogCtx(ctx, "level", "debug", "message", "creating kubeconfig secret")
 
-		restConfig := config.Host.RestConfig()
-		bytes, err := kubeconfig.NewKubeConfigForRESTConfig(ctx, restConfig, "test-cluster", targetNamespace)
-		if err != nil {
-			t.Fatalf("expected nil got %#v", err)
+		// Transform kubeconfig file to restconfig and flatten
```
rossf7: Good catch on finding the problem with minikube.

Makes sense to do this now. It gets simpler with the migration to KIND.

service/controller/app/v1/resource/chartoperator/create.go
tomahawk28: I didn't know that it deployed by draughtsman, so I can skip this resource on inCluster mode as you said. Thanks for pointing it out.

service/controller/app/v1/key/error.go
```diff
@@ -11,6 +11,15 @@ func IsEmptyValueError(err error) bool {
 	return microerror.Cause(err) == emptyValueError
 }
 
+var executionFailedError = &microerror.Error{
+	Kind: "executionFailed",
+}
+
+// IsExecutionFailed asserts executionFailedError.
+func IsExecutionFailed(err error) bool {
```
xh3b4sd: Please check `aws-operator` as reference implementation and see https://github.com/giantswarm/fmt/blob/master/go/errors.md. We do not want to have the matcher for this and the error should move to the top with a comment.

service/controller/app/v1/key/key.go
```diff
@@ -153,3 +159,15 @@ func VersionLabel(customResource v1alpha1.App) string {
 		return ""
 	}
 }
+
+func GenerateTarballURL(baseURL string, appName string, version string) (string, error) {
```
xh3b4sd: IMO we should stop adding key functions that return errors. When the implementation is so complex it may not be a key function.

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -13,6 +13,7 @@ import (
 
 	"github.com/giantswarm/app-operator/pkg/annotation"
 	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+	appcatalogkey "github.com/giantswarm/app-operator/service/controller/appcatalog/v1/key"
```
xh3b4sd: Why should we alias this import? That does not make sense to me.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/giantswarm/micrologger"
+	"github.com/spf13/afero"
+	"gopkg.in/yaml.v2"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/helm/pkg/helm"
+
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+	appcatalogkey "github.com/giantswarm/app-operator/service/controller/appcatalog/v1/key"
```
xh3b4sd: I think when controllers import different key packages something went really wrong in the design. Let's please not do that. I saw this in `cluster-operator` everywhere and it is just painful to work with because it is so complicated.

service/controller/app/v1/key/key.go
rossf7: That makes sense but we need to reuse this logic between the chart and chartoperator resources. What about defining a tarball package for this?

service/controller/app/v1/resource/chartoperator/resource.go
rossf7: Agree I think everything should be in `service/controller/app/v1/key`.  As the appcatalog controller has no resources and is just for creating the CRD.

service/controller/app/v1/key/key.go
xh3b4sd: `pkg/tarball` sounds nice.

service/controller/app/v1/resource/chart/desired_test.go
```diff
@@ -22,7 +21,7 @@ func Test_Resource_GetDesiredState(t *testing.T) {
 		obj           *v1alpha1.App
 		appCatalog    v1alpha1.AppCatalog
 		expectedChart *v1alpha1.Chart
-		errorMatcher  func(error) bool
```
tomahawk28: Change it to simple bool type so we don't have to match against `executionFailedError`

service/controller/app/v1/resource/chartoperator/create.go
tomahawk28: @rossf7 I found out this results in an error from e2etest, since we are not installing fresh chart-operator from `simple-test`. So I revered this change. Nevertheless, we will always look into helm charts list in CP, then we will skip installing chart-operator whenever they are already there. WDYT?

service/controller/app/v1/key/error.go
tomahawk28: Thanks, I removed the matcher function and also change the unittest to not to match with this specific error.

service/controller/app/v1/resource/chartoperator/create.go
rossf7: OK cool. Good point that there is the chart-operator release created by draughtsman.

tomahawk28:
@xh3b4sd PTAL, fixed all the issues that you pointed out :)

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -28,7 +25,7 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 	}
 
 	config := generateConfig(cr, cc.AppCatalog, r.chartNamespace)
-	tarballURL, err := generateTarballURL(appcatalogkey.AppCatalogStorageURL(cc.AppCatalog), key.AppName(cr), key.Version(cr))
+	tarballURL, err := tarball.GenerateTarballURL(key.AppCatalogStorageURL(cc.AppCatalog), key.AppName(cr), key.Version(cr))
```
xh3b4sd: That name stutters from a caller point of view. 

```suggestion
	tarballURL, err := tarball.NewURL(key.AppCatalogStorageURL(cc.AppCatalog), key.AppName(cr), key.Version(cr))
```

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
```
xh3b4sd: Trying that again. 😛 

```suggestion
```

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,48 @@
...
+	// Check whether tenant cluster has a chart-operator helm release yet.
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding chart-operator release %#q in tenant cluster", chartOperatorRelease))
+
+		_, err := cc.HelmClient.GetReleaseContent(ctx, chartOperatorRelease)
+		if helmclient.IsReleaseNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart-perator release %#q in tenant cluster", chartOperatorRelease))
+
```
xh3b4sd:
```suggestion
```

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -0,0 +1,48 @@
...
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("installing chart-operator release %#q in tenant cluster", chartOperatorRelease))
+			err = r.installChartOperator(ctx, cr, cc.HelmClient)
```
xh3b4sd:
```suggestion

			err = r.installChartOperator(ctx, cr, cc.HelmClient)
```

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+)
+
+const (
+	Name = "chartoperatorv1"
+)
+
+const (
+	chartOperatorNamespace = "giantswarm"
+	chartOperatorRelease   = "chart-operator"
```
xh3b4sd: I would just go with `namespace` and `release` if there is no strong reason for being explicit here. The package scope is already `chartoperator`.

service/controller/appcatalog/v1/key/error.go
```diff
@@ -11,6 +11,15 @@ func IsEmptyValueError(err error) bool {
 	return microerror.Cause(err) == emptyValueError
 }
 
+var executionFailedError = &microerror.Error{
+	Kind: "executionFailed",
+}
+
+// IsExecutionFailed asserts executionFailedError.
+func IsExecutionFailed(err error) bool {
```
xh3b4sd: Please drop the matcher and move the error up.

service/controller/app/v1/resource/chartoperator/create.go
tomahawk28: Ah, thought that you meant other lines :)

PR #122, opened 2019-08-13
New cordonchart resource

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/6724

It would add new cordonchart resource, which would run `patch` operation on charts CR. 

By using `patch` operation instead, we can avoid overwriting into the existing annotations.


service/controller/app/v1/resource/cordonchart/create.go
```diff
@@ -0,0 +1,74 @@
...
+	Value map[string]string `json:"value"`
+}
+
+func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if !key.IsCordoned(cr) {
```
tomahawk28: Early return if app is not cordoned.

tomahawk28: I figured out I should remove cordon annotations in chart, should not return in here.

service/controller/app/v1/resource/cordonchart/create.go
```diff
@@ -0,0 +1,61 @@
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		orig := chart.GetAnnotations()
+		_, ok1 := orig[replacePrefix(annotation.CordonUntil)]
+		_, ok2 := orig[replacePrefix(annotation.CordonReason)]
+
+		if !ok1 || !ok2 {
+			return nil
```
tomahawk28: Check cordon annotations placed in `chart` CR and if not, return early.

service/controller/app/v1/resource/cordonchart/resource.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return nil
+}
+
+func replacePrefix(from string) string {
+	return strings.Replace(from, "app-operator.giantswarm.io", "chart-operator.giantswarm.io", 1)
+}
+
+func replaceToEscape(from string) string {
+	return strings.Replace(from, "/", "~1", -1)
```
tomahawk28: All `/` characters should change to `~1` according to JSON Merge RFC documentations. https://tools.ietf.org/html/rfc6902#appendix-A.14

service/controller/app/v1/resource/cordonchart/create.go
tomahawk28: I'm adding this logic since if we are trying to delete cordon key from annotations when it actually doesn't exist, it results in errors.

tomahawk28:
I once added unittest for this resource, but `fakeClientset` is not supporting well enough for JSON merge patch, so I simply delete it.

tomahawk28:
Need to merge this at the same time, https://github.com/giantswarm/chart-operator/pull/274

service/controller/app/v1/resource/cordonchart/resource.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleted cordon annotations for chart CR %#q in namespace %#q", cr.Name, r.chartNamespace))
+
+	return nil
+}
+
+func replacePrefix(from string) string {
+	return strings.Replace(from, "app-operator.giantswarm.io", "chart-operator.giantswarm.io", 1)
```
xh3b4sd: I feel like this is done in a lot of different places. How can we make this a bit better? Why aren't we using the label package here at least? I am also still wondering why we need to do this at all. We patch a chart CR. So why does the chart CR have an `app-operator` annotation at all?

service/controller/app/v1/resource/cordonchart/resource.go
```diff
@@ -0,0 +1,139 @@
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	_, err = client.ApplicationV1alpha1().Charts(r.chartNamespace).Patch(cr.GetName(), types.JSONPatchType, patchByte)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleted cordon annotations for chart CR %#q in namespace %#q", cr.Name, r.chartNamespace))
```
xh3b4sd: The opening log is missing.

service/controller/app/v1/resource/cordonchart/resource.go
```diff
@@ -0,0 +1,139 @@
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	_, err = client.ApplicationV1alpha1().Charts(r.chartNamespace).Patch(cr.GetName(), types.JSONPatchType, patchByte)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("added cordon annotations for chart CR %#q in namespace %#q", cr.Name, r.chartNamespace))
```
xh3b4sd: Opening log missing.

service/controller/app/v1/resource/cordonchart/resource.go
xh3b4sd: Why do we need that though? In the `statusresource` and I am sure in other places this works without this replacement. See https://github.com/giantswarm/statusresource/blob/master/resource.go.

service/controller/app/v1/resource/cordonchart/create.go
```diff
@@ -0,0 +1,62 @@
...
+
+		return nil
+	}
+
+	if key.IsCordoned(cr) {
+		err := r.addCordon(ctx, cr, cc.G8sClient)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	} else {
```
xh3b4sd: We can early return/cancel and not have the else clause.

service/controller/app/v1/resource/cordonchart/resource.go
tomahawk28: if key name contains `/` or `~`, We should convert it to escape characters otherwise the key name `chart-operator.giantswarm.io/cordon-until` would have two depth even though it is one depth in real. So far I don't see the same case in `statusresource`

service/controller/app/v1/resource/cordonchart/resource.go
tomahawk28: @xh3b4sd If we are moving this function to another package, I think it should be `annotations`. 
And this function is to convert `app-operator.giantswarm.io` prefix into `chart-operator.giantswarm.io` so `chart` CR would have no `app-operator` annotations when the operator is done with copying these.

service/controller/app/v1/resource/cordonchart/resource.go
xh3b4sd: If I understand the code correctly what happens here is that the patch path `/metadata/annotations/...` becomes `~1metadata~1annotations~1...` and this looks pretty wrong to me. Did you test this? What is the use case for this weird escaping here? `/` is a path separator AFAIK which is exactly what happens in the `statusresource`.

service/controller/app/v1/resource/cordonchart/resource.go
xh3b4sd: My question was why the chart CR still haves the `app-operator` annotations when it got copied and why we do this at all. My question was if you can explain the whole thing to me a bit more.

xh3b4sd: Also note that I did not say we should move this function to the annotation package. I asked the question how we could improve this and how it should be so that we have a plan or idea or vision for this.

service/controller/app/v1/resource/cordonchart/resource.go
tomahawk28: After I changed the code for adding `annotation`, I didn't test it enough for this.
That's wrong I should only change the last key name which contains `/` so
`/metadata/annotations/"chart-operator.giantswarm.io/cordon-until"` should change into 
`/metadata/annotations/"chart-operator.giantswarm.io~1cordon-until"`

service/controller/app/v1/resource/cordonchart/resource.go
tomahawk28: Annotations related to cordon features are from `app` CRs, and we copy the same cordon annotations into `chart` CRs in TC to stop affecting into helm charts. But only we need to change the prefix in annotations so hiding the information the app-operator is manipulating these cordon job from the background. 
Also, Having `chart` CR with other operator's prefix is wrong in the design concept I believe. 

That's why I thought about removing this prefix for this cordon annotation once for all if we need to copy this information from CRs to CRs in another type. 
What makes sense to have prefix from the start in this case?
@xh3b4sd WDYT?

service/controller/app/v1/resource/cordonchart/resource.go
xh3b4sd: That makes sense yes.

service/controller/app/v1/resource/cordonchart/resource.go
xh3b4sd: Who decides when to cordon an app? How does this process look like? I am wondering why two different CRs and two different operator responsibilities need to be involved here. This copy back and forth is something I do not get into my head. I do not understand that. Why do we need the cordon annotations in two different places?

tomahawk28: `chart` CRs in TC are merely the reconciliation result from `app` CR, therefore, we need to set annotations into `app` CRs to have it in `chart` CRs. That's why we involved two different operators in here. 
If you have more questions I think the technical spec for phase 2 would be very helpful. 
https://github.com/giantswarm/giantswarm/blob/master/specs/app-catalog-phase-2.md
Or we can jump into a joint call about this.

xh3b4sd: Thank you for the spec link. I start to understand it. What the spec does not answer apparently is why we cordon at the start of the responsibility chain instead of at the end. Why don't we "ignore" `app-operator` in that regard and only cordon the `Chart` CRs in the TCs?

tomahawk28: @xh3b4sd What if we trying to change the user config or chart version inside `app` CR? I think we need to mitigate those changes before it affects `chart` CRs in TC. So far only `cordon` related annotation can stop this, WDYT?

tomahawk28:
I talked with @xh3b4sd and agreed that on cordoning on `app` CRs for chain effects on to `chart` CRs is too much, rather we would like to put cordon annotation on `chart` CR directly. 
So I'm closing this PR.

tomahawk28:
Slack: https://gigantic.slack.com/archives/C2MS2VB37/p1565861859171500

PR #125, opened 2019-08-28
do not installing chart-operator in case of in-cluster 

tomahawk28:
Fixing a problem trying to install chart-operator in case of in-cluster mode.

### DIFF
--- a/integration/templates/chart_operator_values.go
+++ b/integration/templates/chart_operator_values.go
@@ -3,4 +3,9 @@
 package templates
 
 // ChartOperatorValues values required by chart-operator-chart.
-const ChartOperatorValues = `clusterDNSIP: 10.96.0.10`
+const ChartOperatorValues = `
+clusterDNSIP: 10.96.0.10
+resource:
+  tiller:
+    namespace: "giantswarm"
+`
--- a/integration/test/app/basic/simple_test.go
+++ b/integration/test/app/basic/simple_test.go
@@ -215,6 +215,10 @@ func TestAppLifecycle(t *testing.T) {
 		config.Logger.LogCtx(ctx, "level", "debug", "message", "checked tarball URL in chart spec")
 	}
 
+	// TODO: Delete a release instead of app CR only if all resources not result in errors.
+	//
+	// See: https://github.com/giantswarm/giantswarm/issues/6825
+	//
 	{
 		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting %#q app CR", key.TestAppReleaseName()))
 
--- a/integration/test/app/basic/simple_test.go
+++ b/integration/test/app/basic/simple_test.go
@@ -70,6 +70,10 @@ func TestAppLifecycle(t *testing.T) {
 
 		var tarballPath string
 		{
+			// TODO: Remove hard-coded tarballURL when VOO feature are ready to use.
+			//
+			//	See https://github.com/giantswarm/giantswarm/issues/6824
+			//
 			tarballURL := "https://giantswarm.github.io/giantswarm-catalog/chart-operator-0.9.0.tgz"
 			tarballPath, err = config.HelmClient.PullChartTarball(ctx, tarballURL)
 			if err != nil {
--- a/integration/test/app/basic/simple_test.go
+++ b/integration/test/app/basic/simple_test.go
@@ -9,14 +9,18 @@ import (
 
 	"github.com/giantswarm/e2e-harness/pkg/release"
 	"github.com/giantswarm/e2etemplates/pkg/chartvalues"
+	"github.com/spf13/afero"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/helm/pkg/helm"
 
 	"github.com/giantswarm/app-operator/integration/key"
+	"github.com/giantswarm/app-operator/integration/templates"
 )
 
 const (
-	namespace            = "giantswarm"
 	chartOperatorVersion = "chart-operator.giantswarm.io/version"
+	chartOperatorRelease = "chart-operator"
+	namespace            = "giantswarm"
 	testAppCatalogName   = "test-app-catalog"
 )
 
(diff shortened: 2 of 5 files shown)


integration/templates/chart_operator_values.go
```diff
@@ -3,4 +3,9 @@
 package templates
 
 // ChartOperatorValues values required by chart-operator-chart.
-const ChartOperatorValues = `clusterDNSIP: 10.96.0.10`
+const ChartOperatorValues = `
+clusterDNSIP: 10.96.0.10
+resource:
+  tiller:
+    namespace: "giantswarm"
```
tomahawk28: Adding it otherwise it would use kube-system as a default.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -21,6 +22,15 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	if key.InCluster(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q is targeting a control plance, no need to setup a chart operator", cr.Name))
+
+		resourcecanceledcontext.SetCanceled(ctx)
```
tomahawk28: Canceling like this is okay? or just return nil at this point?

xh3b4sd: https://github.com/giantswarm/operatorkit/blob/master/docs/control_flow_primitives.md#cancel-resources

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -21,6 +22,15 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
```
xh3b4sd: Please first logging then function call.

service/controller/app/v1/resource/chartoperator/create.go
tomahawk28: > The resources being configured after the canceled resource are then executed and the canceled resource is executed again on the next reconciliation loop once the resync period kicks in again.
I will remove this line then, Thanks.

integration/templates/chart_operator_values.go
rossf7: Makes sense. We need to always configure the namespace for tenant clusters. One day we will be tiller-less. Way better than server-less IMO 😄

integration/test/app/basic/simple_test.go
```diff
@@ -212,25 +212,14 @@ func TestAppLifecycle(t *testing.T) {
 	}
 
 	{
-		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting release %#q", key.CustomResourceReleaseName()))
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting %#q app CR", key.TestAppReleaseName()))
 
-		err := config.Release.Delete(ctx, key.CustomResourceReleaseName())
+		err = config.Host.G8sClient().ApplicationV1alpha1().Apps(namespace).Delete(key.TestAppReleaseName(), &metav1.DeleteOptions{})
```
tomahawk28: This is for fixing the issue that entire configmap, secret and appCatalog deleted before we tried to delete `chart` CR.

integration/test/app/basic/simple_test.go
```diff
@@ -61,6 +65,33 @@ func TestAppLifecycle(t *testing.T) {
 		Namespace: namespace,
 	}
 
+	{
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("installing chart operator"))
+
+		var tarballPath string
+		{
+			tarballURL := "https://giantswarm.github.io/giantswarm-catalog/chart-operator-0.9.0.tgz"
```
rossf7: The hardcoded version here is a problem. In the VOO spec the design is to have a label called chart-operator.giantswarm.io/version="control-plane". that always maps to master.

https://github.com/giantswarm/giantswarm/blob/master/specs/versioning-of-operators.md#bootstrap

Can we have a TODO issue to get this straight later?

integration/test/app/basic/simple_test.go
rossf7: I think the operator needs to be able to handle this. Don't we have logic in the configmap resource etc that checks if the app CR is deleted?

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -21,6 +21,13 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	if key.InCluster(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q is targeting a control plance, no need to setup a chart operator", cr.Name))
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q in %#q uses InCluster kubeconfig no need to install chart operator", cr.Name, cr.Namespace))
```

🇺🇸 nitpick but this is a bit more generic and matches the if statement.

integration/test/app/basic/simple_test.go
tomahawk28: If we deleted the entire release, helm would delete related configmaps, secrets and appCatalogs. 
But if helm deletes other resources earlier than `app` CR in deletion event, app-operator could not find the resources related to this release, so it hangs. 
It happened time by time and break e2etest some cases.

service/controller/app/v1/resource/chartoperator/create.go
tomahawk28: Thanks 🇬🇧

integration/test/app/basic/simple_test.go
rossf7: I think that's a bug then. We can't 100% control the order of deletion and if this happens in production there may be alerts.

When the app CR is deleted the operator needs to handle these cases. Please create a bug issue and link another TODO here. So we can get it added to our team board and then fixed.

integration/test/app/basic/simple_test.go
tomahawk28: Sure, I added 2 TODOs on the code.

PR #126, opened 2019-08-28
Tiller resource 

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/6078

Moving tiller resource from cluster-operator.


integration/test/app/basic/simple_test.go
```diff
@@ -65,6 +65,17 @@ func TestAppLifecycle(t *testing.T) {
 		Namespace: namespace,
 	}
 
+	{
```
tomahawk28: e2e would install tiller unless it's using `inCluster` mode, hence we install this manually in case of basic test.

service/controller/app/v1/resource/tiller/resource.go
```diff
@@ -0,0 +1,96 @@
...
+		"spec.template.spec.tolerations[0].operator=Exists",
+	}
+
+	var err error
+
+	err = helmClient.EnsureTillerInstalledWithValues(ctx, values)
+	if tenantcluster.IsTimeout(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "timeout fetching certificates")
+
+		// A timeout error here means that the app-operator certificate
```
tomahawk28: These comments are from cluster-operator, are these okay?

service/controller/app/v1/resource_set.go
```diff
@@ -209,10 +210,22 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
 	resources := []controller.Resource{
 		appNamespaceResource,
 		appcatalogResource,
 		clientsResource,
+		tillerResource,
```
tomahawk28: Put it right after `clients` so we can install tiller after helmclient are up.

integration/setup/setup.go
```diff
@@ -52,13 +52,6 @@ func installResources(ctx context.Context, config Config) error {
 		}
 	}
 
-	{
```
tomahawk28: Exclude this step from setup and move it to basic test only.

tomahawk28: I was wrong, if we exclude it from e2e test, we can setup app-operator.

tomahawk28:
Tested on ginger

service/controller/app/v1/resource/tiller/resource.go
rossf7: We can remove this case because we don't use tenantcluster. It was replaced by kubeconfig and the clients resource handles creating the helmclient.

Other cases are fine.

service/controller/app/v1/resource_set.go
rossf7: Nice 👍

PR #128, opened 2019-08-29
Namespace resource

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/6078

Moving `namespace` resource from cluster-operator.


service/controller/app/v1/resource/namespace/current.go
```diff
@@ -0,0 +1,67 @@
...
+func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	if key.InCluster(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q in %#q uses InCluster kubeconfig no need to install a namespace", cr.Name, cr.Namespace))
+
+		resourcecanceledcontext.SetCanceled(ctx)
```
tomahawk28: app-operator have several codes that calling `SetCanceled` earlier than logs. is that okay?

service/controller/app/v1/resource/namespace/desired.go
```diff
@@ -0,0 +1,55 @@
...
+func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	var clusterID, org string
+	{
+		name := key.ClusterValuesConfigMapName(cr)
+		cm, err := r.k8sClient.CoreV1().ConfigMaps(cr.Namespace).Get(name, metav1.GetOptions{})
```
tomahawk28: Fetching `clusterconfigmap` for getting labels from it.

service/controller/app/v1/resource_set.go
```diff
@@ -225,6 +244,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 		appNamespaceResource,
 		appcatalogResource,
 		clientsResource,
+		namespaceResource,
```
tomahawk28: Next to the `clients` resource to initiate namespace.

service/controller/app/v1/resource/namespace/desired.go
xh3b4sd: Why does the `App` CR not just have the cluster ID and organization ID labels?

tomahawk28: @xh3b4sd `app` CRs are located in control plane, so it supposed to not contain `cluster` label, although I'm not sure `organization` at there. 
We could fetch two labels on `clusterconfigmap` so I chose to fetching from this object.

xh3b4sd: I don't follow that. Check any resource in the CP we manage for the TCs. They all have these labels.

xh3b4sd: Also see https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md#common-labels.

tomahawk28: @xh3b4sd You are right
```
 k get apps --all-namespaces -o yaml
apiVersion: v1
items:
- apiVersion: application.giantswarm.io/v1alpha1
  kind: App
  metadata:
    annotations:
      chart-operator.giantswarm.io/force-helm-upgrade: "true"
    creationTimestamp: "2019-08-29T07:50:00Z"
    finalizers:
    - operatorkit.giantswarm.io/app-operator
    generation: 1
    labels:
      app: kube-state-metrics
      app-operator.giantswarm.io/version: 1.0.0
      giantswarm.io/cluster: mbnc3
      giantswarm.io/managed-by: cluster-operator
      giantswarm.io/organization: giantswarm
      giantswarm.io/service-type: managed
    name: kube-state-metrics
```
I can make it to work with `app` CR, not involved any configMaps in here, Thanks.

tomahawk28:
@xh3b4sd I had brought all the changes from the latest cluster-operator, PTAL! Thanks.

service/controller/app/v1/resource/namespace/current.go
xh3b4sd: Not a big issue but the majority is the other way around. So also the code below for example. 😛

service/controller/app/v1/resource/namespace/current.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	if key.InCluster(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q in %#q uses InCluster kubeconfig no need to install a namespace", cr.Name, cr.Namespace))
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q in %#q uses InCluster kubeconfig no need to create namespace", cr.Name, cr.Namespace))
```

Lets also remove the blank line.

service/controller/app/v1/resource/namespace/resource.go
```diff
@@ -0,0 +1,64 @@
...
+
+// Config represents the configuration used to create a new namespace resource.
+type Config struct {
+	K8sClient kubernetes.Interface
+	Logger    micrologger.Logger
+}
+
+// Resource implements the namespace resource.
+type Resource struct {
+	k8sClient kubernetes.Interface
```
rossf7: I think this k8sClient can be removed because we get it from the context.

service/controller/app/v1/resource/namespace/desired.go
sslavic: @rossf7 while analyzing some alerts and related logs, I found examples of Apps where assumption above about common labels being present on the App CRs does not hold - on anubis CP there are currently 3 Apps (2 aqua-app-enforcer Apps for 2 TCs, and 1 cluster-operator-2.0.0 App as configured in [kvm-app-collection](https://github.com/giantswarm/kvm-app-collection/blob/master/helm/kvm-app-collection-chart/templates/cluster-operator-2.0.0.yaml#L6)), all of which have only `app-operator.giantswarm.io/version=1.0.0` label; because of this log messages in e.g. in `namespace/current.go` logging cluster ID currently log empty strings there. On the other hand two`*-cluster-values` ConfigMaps have both `giantswarm.io/cluster` and `giantswarm.io/organization` label set.

Are these Apps without standard labels temporary exception? If not, should we switch back to deriving cluster and organization info from cluster values ConfigMaps?

rossf7: Yes these app CRs are being created by different components which is why the labels are not consistent.

Having gatekeeper do the defaulting would help here. @puja108 WDYT?

aqua-app-enforcer was probably created by the api and it misses adding the cluster label.
https://github.com/giantswarm/api/blob/260f363ae1d94d088aabb487a0a96d71362df050/service/app/creator/service.go#L102

The cluster-operator CR is probably because the helm chart misses the label. The namespace resource only interacts with tenant clusters but we need to make sure the resource doesn't error and logs OK.

rossf7: The cluster values CM is not used by VOO app CRs. So I think we should add the missing labels.

PR #130, opened 2019-08-30
No-op in appCatalog resource

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/6825 

We should not fetch appCatalog in case of deletion since it could be the part of helm chart release.

### DIFF
--- a/integration/test/app/basic/simple_test.go
+++ b/integration/test/app/basic/simple_test.go
@@ -235,19 +235,15 @@ func TestAppLifecycle(t *testing.T) {
 		config.Logger.LogCtx(ctx, "level", "debug", "message", "checked tarball URL in chart spec")
 	}
 
-	// TODO: Delete a release instead of app CR only if all resources not result in errors.
-	//
-	// See: https://github.com/giantswarm/giantswarm/issues/6825
-	//
 	{
-		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting %#q app CR", key.TestAppReleaseName()))
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting release %#q", key.CustomResourceReleaseName()))
 
-		err = config.K8sClients.G8sClient().ApplicationV1alpha1().Apps(namespace).Delete(key.TestAppReleaseName(), &metav1.DeleteOptions{})
+		err := config.Release.Delete(ctx, key.CustomResourceReleaseName())
 		if err != nil {
 			t.Fatalf("expected %#v got %#v", nil, err)
 		}
 
-		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleted %#q app CR", key.TestAppReleaseName()))
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleted release %#q", key.CustomResourceReleaseName()))
 	}
 
 	{
--- a/service/controller/app/v1/resource/chart/desired.go
+++ b/service/controller/app/v1/resource/chart/desired.go
@@ -2,6 +2,7 @@ package chart
 
 import (
 	"context"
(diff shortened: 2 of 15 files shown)


integration/test/app/basic/simple_test.go
```diff
@@ -235,19 +235,15 @@ func TestAppLifecycle(t *testing.T) {
 		config.Logger.LogCtx(ctx, "level", "debug", "message", "checked tarball URL in chart spec")
 	}
 
-	// TODO: Delete a release instead of app CR only if all resources not result in errors.
-	//
-	// See: https://github.com/giantswarm/giantswarm/issues/6825
-	//
 	{
-		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting %#q app CR", key.TestAppReleaseName()))
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting release %#q", key.CustomResourceReleaseName()))
```
tomahawk28: Revert it to delete a helm chart release.

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -24,6 +24,15 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
+	if cc.AppCatalog.Name == "" {
+		return &v1alpha1.Chart{
+			ObjectMeta: metav1.ObjectMeta{
```
tomahawk28: returning empty object but having an object name on it to trigger a deletion event.

xh3b4sd: The desired state should not be conditional in order to cause a different outcome. IMO the desired state should always be the same. The decision on when to delete something should be made when computing the delete patch.

tomahawk28: When an operator calculates the desired state, it needs to have cc.AppCatalog instance to make configMap, secret and charts. I was thinking about make related information as an empty string, (e.g. tarball) but that would affect creation events as well. 

@xh3b4sd Any suggestion on that?

xh3b4sd: When the desired state cannot be calculated the resource cannot work and should cancel until it is able to compute the desired state again. The desired state defines how the state should be. The current state defines how the state is right now. Deciding under which conditions to apply the desired state should be done when computing the create/update/delete patches. When we have conditionals when computing the desired state, the code is wrong.

tomahawk28: Maybe you already saw an issue that I linked in the description,  app-operator could not find appCatalog which is the part of helm chart release that deleted earlier. 
@xh3b4sd Based on what you said, I think it's bad design to put `appCatalog` in same helm chart, since we could not figure it out desired state. What do you think? 

BTW, I don't get that why we need to calculate full desired state in a deletion event.

tomahawk28:
We decided to do as follow, 
1. Keep desired state calculation get as far, even the operator could not find a relevant resource of it. In this case, this is appCatalog. 
2. If we have a problem in CRD validation, @rossf7 would take care of it to apply the same solution from he got yesterday.

tomahawk28:
Okay, it's happening on other resource, like configMap, secret since it's deleted earlier before than `app` CR. I don't think we can keep making the desired state in this case. 
Maybe we should think about applying finalizer in here?

rossf7:
> Okay, it's happening on other resource, like configMap, secret since it's deleted earlier before than app CR. I don't think we can keep making the desired state in this case.
Maybe we should think about applying finalizer in here?

Yes, the app CR depends on other resources like the appcatalog CR and the related configmaps and secrets. We should take the approach that we generate as much of the desired state as we can and leave the rest blank.

I'd suggest doing this one resource at a time. I don't see how adding a finalizer would help.

tomahawk28:
@xh3b4sd @rossf7 I make a change as we agreed on the call, PTAL.

PR #132, opened 2019-09-02
e2e: Remove unused guest framework and add e2esetup dep

rossf7:
Fixes a couple of things I found when testing the app CRD validation.


Gopkg.toml
```diff
@@ -48,6 +48,10 @@ required = [
   branch = "master"
   name = "github.com/giantswarm/apiextensions"
 
+[[constraint]]
+  branch = "master"
+  name = "github.com/giantswarm/e2esetup"
```
rossf7: Dep needed adding.

integration/setup/config.go
```diff
@@ -3,7 +3,6 @@
 package setup
 
 import (
-	"github.com/giantswarm/e2e-harness/pkg/framework"
```
rossf7: Dropping dead code. 🎉

PR #140, opened 2019-09-16
vendor: updated apiextensions

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/5407

### DIFF
--- a/helm/app-operator-chart/templates/rbac.yaml
+++ b/helm/app-operator-chart/templates/rbac.yaml
@@ -13,6 +13,12 @@ rules:
   verbs:
   - "*"
 - apiGroups:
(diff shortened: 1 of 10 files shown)


helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -12,6 +12,12 @@ rules:
   - customresourcedefinitions
   verbs:
   - "*"
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions/status
+  verbs:
+  - "*"
```
rossf7: Without permission to update the CRD status the operator crashes if there is already an app CRD.

PR #141, opened 2019-09-16
Improvements to app status resource

rossf7:
Some minor stuff to improve the status resource.

### DIFF
--- a/service/controller/app/v1/resource/status/create.go
+++ b/service/controller/app/v1/resource/status/create.go
@@ -21,8 +21,6 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	name := cr.GetName()
-
 	cc, err := controllercontext.FromContext(ctx)
 	if err != nil {
 		return microerror.Mask(err)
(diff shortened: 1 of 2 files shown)


service/controller/app/v1/resource/status/create.go
```diff
@@ -21,27 +21,24 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	name := cr.GetName()
-
```
rossf7: Aligning the approach so we get both the name and namespace from the CR.

service/controller/app/v1/resource/status/create.go
```diff
@@ -72,10 +69,10 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	}
 
 	if !equals(desiredStatus, key.AppStatus(cr)) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting status for app %#q in namespace %#q", name, r.chartNamespace))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("setting status for app %#q in namespace %#q", cr.Name, cr.Namespace))
```
rossf7: This was wrong. The chart namespace is `giantswarm` but the app CR is in the cluster namespace.

PR #142, opened 2019-09-17
Use values service in chart-operator resource

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/6874

On Azure and some onprem installations we can't use 8.8.8.8 for bootstrapping CoreDNS. This is not set in the default app catalog.

This is some refactoring to use the `values` package for generating the values used to install chart-operator. So we don't need to make code changes when we add new values.

### DIFF
--- a/integration/templates/chart_operator_values.go
+++ b/integration/templates/chart_operator_values.go
@@ -5,11 +5,8 @@ package templates
 // ChartOperatorValues values required by chart-operator-chart.
 const ChartOperatorValues = `
 clusterDNSIP: 10.96.0.10
-<<<<<<< HEAD
-=======
 image:
   registry: "quay.io"
->>>>>>> master
 tiller:
   namespace: "giantswarm"
 `
--- a/integration/templates/chart_operator_values.go
+++ b/integration/templates/chart_operator_values.go
@@ -5,6 +5,7 @@ package templates
 // ChartOperatorValues values required by chart-operator-chart.
 const ChartOperatorValues = `
 clusterDNSIP: 10.96.0.10
+e2e: true
 externalDNSIP: 8.8.8.8
 image:
   registry: "quay.io"
--- a/integration/templates/chart_operator_values.go
+++ b/integration/templates/chart_operator_values.go
@@ -5,6 +5,7 @@ package templates
 // ChartOperatorValues values required by chart-operator-chart.
 const ChartOperatorValues = `
 clusterDNSIP: 10.96.0.10
+externalDNSIP: 8.8.8.8
 image:
   registry: "quay.io"
 tiller:
--- a/integration/templates/chart_operator_values.go
+++ b/integration/templates/chart_operator_values.go
@@ -5,7 +5,6 @@ package templates
 // ChartOperatorValues values required by chart-operator-chart.
 const ChartOperatorValues = `
 clusterDNSIP: 10.96.0.10
-resource:
-  tiller:
-    namespace: "giantswarm"
+tiller:
+  namespace: "giantswarm"
 `
--- a/integration/test/app/kubeconfig/kubeconfig_test.go
+++ b/integration/test/app/kubeconfig/kubeconfig_test.go
@@ -155,7 +155,7 @@ func TestAppLifecycleUsingKubeconfig(t *testing.T) {
 				},
 				Name:      "chart-operator",
 				Namespace: "giantswarm",
-				Version:   "0.10.2",
+				Version:   "0.10.4",
 			},
 		})
 		if err != nil {
--- a/integration/test/app/kubeconfig/kubeconfig_test.go
+++ b/integration/test/app/kubeconfig/kubeconfig_test.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd/api"
 
 	"github.com/giantswarm/app-operator/integration/key"
+	"github.com/giantswarm/app-operator/integration/templates"
 )
 
 const (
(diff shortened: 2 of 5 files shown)


integration/templates/chart_operator_values.go
```diff
@@ -5,6 +5,8 @@ package templates
 // ChartOperatorValues values required by chart-operator-chart.
 const ChartOperatorValues = `
 clusterDNSIP: 10.96.0.10
+e2e: true
```
rossf7: This is so the `giantswarm-critical` priority class is created. Otherwise chart-operator won't start and e2e fails.

integration/test/app/kubeconfig/kubeconfig_test.go
```diff
@@ -121,16 +122,35 @@ func TestAppLifecycleUsingKubeconfig(t *testing.T) {
...
+			},
+		})
+		if err != nil {
+			t.Fatalf("expected nil got %#v", err)
+		}
+
 		_, err = config.K8sClients.G8sClient().ApplicationV1alpha1().AppCatalogs().Create(&v1alpha1.AppCatalog{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: "default",
+				Name: "default-test",
```
rossf7: TODO: Put back default catalog once new chart-operator release is created.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -69,9 +69,8 @@ func New(config Config) (*Resource, error) {
 	if config.Logger == nil {
 		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
 	}
-
-	if config.RegistryDomain == "" {
-		config.RegistryDomain = "quay.io"
```
rossf7: The image registry is in the catalog configmap.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -148,48 +154,21 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
...
 
-	var clusterDNSIP string
+	var chartOperatorValues []byte
 	{
-		name := key.ClusterValuesConfigMapName(cr)
-		cm, err := r.k8sClient.CoreV1().ConfigMaps(cr.Namespace).Get(name, metav1.GetOptions{})
-		if apierrors.IsNotFound(err) {
-			r.logger.LogCtx(ctx, "level", "warning", "message", fmt.Sprintf("no cluster-value %#q in control plane, operator will use default clusterDNSIP value", name))
-		} else if err != nil {
+		values, err := r.values.MergeAll(ctx, *chartOperatorAppCR, *appCatalogCR)
```
rossf7: Now we just merge all the values.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -148,48 +154,21 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
...
-	var chartOperatorValue []byte
-	{
-		v := map[string]interface{}{
-			"image": map[string]string{
-				"registry": r.registryDomain,
-			},
-			"tiller": map[string]string{
-				"namespace": namespace,
-			},
-		}
```
rossf7: We remove the hardcoding here.

rossf7:
Tested on `ghost`.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -98,36 +95,45 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
...
+	var appCatalogCR *v1alpha1.AppCatalog
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", "finding appCatalog CR")
 
+		catalogName := key.CatalogName(*chartOperatorAppCR)
+		appCatalogCR, err = r.g8sClient.ApplicationV1alpha1().AppCatalogs().Get(catalogName, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "can't find appCatalog CR")
 			r.logger.LogCtx(ctx, "level", "debug", "message", "cancelling the resource")
+			resourcecanceledcontext.SetCanceled(ctx)
```
tomahawk28: if no chart-operator in TC, does it make sense to continue to other resources? Why not simply cancel the whole reconciliation?

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -213,7 +192,7 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
 		// Wait for chart-operator to be deployed. If it takes longer than
 		// the timeout the chartconfig CRs will be created during the next
 		// reconciliation loop.
-		b := backoff.NewConstant(30*time.Second, 5*time.Second)
+		b := backoff.NewConstant(20*time.Second, 10*time.Second)
```
tomahawk28: I will close my PR then.

service/controller/app/v1/resource/chartoperator/resource.go
rossf7: Good point thanks. I want to merge this but I'll do a follow up PR.

service/controller/app/v1/resource/chartoperator/resource.go
rossf7: Thanks and yes I just copied these settings from your PR.

PR #143, opened 2019-09-18
Align values with chart-operator chart

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/6800

Chart values were changed in https://github.com/giantswarm/chart-operator/pull/287.

### DIFF
--- a/service/controller/app/v1/resource/chartoperator/resource.go
+++ b/service/controller/app/v1/resource/chartoperator/resource.go
@@ -170,13 +170,11 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
 	var chartOperatorValue []byte
 	{
 		v := map[string]interface{}{
-			"resource": map[string]interface{}{
-				"image": map[string]string{
-					"registry": r.registryDomain,
-				},
-				"tiller": map[string]string{
-					"namespace": namespace,
-				},
+			"image": map[string]string{
+				"registry": r.registryDomain,
+			},
+			"tiller": map[string]string{
+				"namespace": namespace,
 			},
 		}
 
--- a/integration/test/app/kubeconfig/kubeconfig_test.go
+++ b/integration/test/app/kubeconfig/kubeconfig_test.go
@@ -123,13 +123,13 @@ func TestAppLifecycleUsingKubeconfig(t *testing.T) {
 
 		_, err = config.K8sClients.G8sClient().ApplicationV1alpha1().AppCatalogs().Create(&v1alpha1.AppCatalog{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: "giantswarm",
+				Name: "default",
 			},
 			Spec: v1alpha1.AppCatalogSpec{
-				Title: "Giant Swarm Catalog",
+				Title: "Giant Swarm Default Catalog",
 				Storage: v1alpha1.AppCatalogSpecStorage{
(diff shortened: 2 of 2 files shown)


rossf7:
Tested on `ghost` but the kubeconfig e2e test is failing.

```
helm --tiller-namespace giantswarm get values chart-operator
clusterDNSIP: 172.31.0.10
image:
  registry: quay.io
tiller:
  namespace: giantswarm
```

```
k get cm chart-operator -o yaml
apiVersion: v1
data:
  config.yaml: |
    server:
      enable:
        debug:
          server: true
      listen:
        address: 'http://0.0.0.0:8000'
    service:
      cnr:
        address: 'https://quay.io'
      helm:
        tillerNamespace:  'giantswarm'
      kubernetes:
        incluster: true
        watch:
          namespace: 'giantswarm'
kind: ConfigMap
metadata:
  creationTimestamp: "2019-09-18T10:50:09Z"
  labels:
    app: chart-operator
    giantswarm.io/service-type: managed
  name: chart-operator
  namespace: giantswarm
```

@tomahawk28 Any ideas on why the test fails?

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -170,13 +170,11 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
...
-				"image": map[string]string{
-					"registry": r.registryDomain,
-				},
-				"tiller": map[string]string{
-					"namespace": namespace,
-				},
+			"image": map[string]string{
+				"registry": r.registryDomain,
+			},
+			"tiller": map[string]string{
```
tomahawk28: After this change, tiller had been installed in `kube-system` in e2e test. Did you align it with template changes?

sslavic: https://github.com/giantswarm/chart-operator/pull/287 ?

tomahawk28: ah, the problem is that we are pulling older chart-operator in e2e, we need to pull the latest in this case. @rossf7 Can you update that?

rossf7: Thanks! Yes makes sense it's the tiller namespace.

But the values structure needed to change. The resource block is just for installing multiple copies of the chart for VOO.

But it's changed in master for both charts. So what chart tarball is being used?

rossf7: @tomahawk28 Damn our updates crossed! 😆 Yes sure!

rossf7: Should be good now but let's see.🤞

rossf7: I didn't change the catalog.

picard-facepalm

tomahawk28: I think you need to change the name of here too https://github.com/giantswarm/app-operator/blob/790dcc062a9c4e79f3d900906063d6fb4a03591a/integration/test/app/kubeconfig/kubeconfig_test.go#L132

rossf7: Wow reading is hard today. Thx!

rossf7:
CI is green. 🎉 @tomahawk28 thx for the help! PTAL

PR #145, opened 2019-09-23
Use InstallWait in chartoperator resource

rossf7:
In the chartoperator resource in cluster-operator we use `helm.InstallWait`. In app-operator we aren't using it.

It could be related to weird problems we're seeing with chart-operator bootstrapping.


rossf7:
Not needed.

PR #147, opened 2019-09-26
Use externalDNSIP in chartoperator resource

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/6874

We use `8.8.8.8` for DNS to bootstrap CoreDNS in new tenant clusters. But on Azure we need to use `168.63.129.16`. We also need to use a different value in some onprem installations.

This adds the value when creating the helm release for chart-operator.


rossf7:
Replaced by #142.

PR #150, opened 2019-09-30
Update chart-operator in case of failed release

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/7095

Update a chart-operator release in case of failures so we don't keep in a loop for updating.

### DIFF
--- a/service/controller/app/v1/resource/chartoperator/create.go
+++ b/service/controller/app/v1/resource/chartoperator/create.go
@@ -58,7 +58,7 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart-perator release %#q in tenant cluster", release))
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("installing chart-operator release %#q in tenant cluster", release))
 
-			err = r.installChartOperator(ctx, cr)
+			err = r.installChartOperator(ctx, cr, false)
 			if err != nil {
 				return microerror.Mask(err)
 			}
@@ -68,6 +68,22 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			return microerror.Mask(err)
 		} else {
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found chart-operator release %#q", release))
+
+			releaseContent, err := cc.HelmClient.GetReleaseContent(ctx, release)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			if releaseContent.Status == "FAILED" {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("chart-operator release %#q failed to install", release))
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("updating a release %#q", release))
+				err = r.installChartOperator(ctx, cr, true)
+				if err != nil {
+					return microerror.Mask(err)
+				}
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("updated a release %#q", release))
+
+			}
 		}
 	}
 
--- a/service/controller/app/v1/resource/chartoperator/create.go
+++ b/service/controller/app/v1/resource/chartoperator/create.go
@@ -58,7 +58,7 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart-perator release %#q in tenant cluster", release))
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("installing chart-operator release %#q in tenant cluster", release))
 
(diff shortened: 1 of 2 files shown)


service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -89,7 +89,7 @@ func (r Resource) Name() string {
 	return Name
 }
 
-func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) error {
+func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App, update bool) error {
```
tomahawk28: `update` parameter added since we want to distinguish between initial install and retrial.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -168,9 +168,16 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
 	}
 
 	{
-		err = cc.HelmClient.InstallReleaseFromTarball(ctx, tarballPath, namespace, helm.ReleaseName(release), helm.ValueOverrides(chartOperatorValues))
-		if err != nil {
-			return microerror.Mask(err)
+		if update {
+			err = cc.HelmClient.UpdateReleaseFromTarball(ctx, release, tarballPath, helm.UpdateValueOverrides(chartOperatorValues), helm.UpgradeForce(true))
```
tomahawk28: `helm.UpgradeForce(true)` added so we are not updating revision number and to ensure that release not fall into the loop.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -68,6 +68,22 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found chart-operator release %#q", release))
+
+			releaseContent, err := cc.HelmClient.GetReleaseContent(ctx, release)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			if releaseContent.Status == "FAILED" {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("chart-operator release %#q failed to install", release))
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("updating a release %#q", release))
```
rossf7: Newline here pls.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -68,6 +68,22 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+				err = r.installChartOperator(ctx, cr, true)
+				if err != nil {
+					return microerror.Mask(err)
+				}
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("updated a release %#q", release))
```
rossf7: Another newline here. Makes it easier to read and matches the rest of the code.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -68,6 +68,22 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+
+			releaseContent, err := cc.HelmClient.GetReleaseContent(ctx, release)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			if releaseContent.Status == "FAILED" {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("chart-operator release %#q failed to install", release))
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("updating a release %#q", release))
+				err = r.installChartOperator(ctx, cr, true)
```
rossf7: I don't really like that this is `r.installChartOperator` which it's actually an update.

Most of the code is merging the values. That could be extracted to `r.mergeChartOperatorValues` and we could have a separate `r.updateChartOperator` without the need for the boolean flag.

tomahawk28: Thanks, I moved a bit to new functions.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -154,30 +134,100 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
...
+			return microerror.Mask(err)
+		}
+
+		defer func() {
+			err := r.fileSystem.Remove(tarballPath)
+			if err != nil {
+				r.logger.LogCtx(ctx, "level", "error", "message", fmt.Sprintf("deletion of %#q failed", tarballPath), "stack", fmt.Sprintf("%#v", err))
+			}
+		}()
+	}
```
rossf7: Pulling the chart tarball is also repeated for install and update. I think it can also be extracted to a `r.pullChartOperatorTarball` method.

tomahawk28: @rossf7 My only concern is that we are using `defer` to delete the tarball file after use. Even If we would move it as another function, that means we are repeating `defer` into two places which I don't think ideal. WDYT?

rossf7: Ack. Good point about the defer.

PR #153, opened 2019-10-02
Set force annotation for chart CR if it is set for the app CR

rossf7:
When upgrading chart-operator with helm 2.14.3 vendored it fails with this error.

```
/go/src/github.com/giantswarm/chart-operator/vendor/github.com/giantswarm/helmclient/helmclient.go:607:
	rpc error: code = Unknown desc = kind ClusterRole with the name "chart-operator-psp-user" already exists in the cluster and wasn't defined in the previous release. Before upgrading, please either delete the resource from the cluster or remove it from the chart
```

The fix is to use `helm.ForceUpgrade(true)` so the `chart-operator-psp-user` is deleted as part of the upgrade.

But the annotation that controls this is only set on the app CR. It needs to be passed on to the chart CR.

### DIFF
--- a/pkg/annotation/annotation.go
+++ b/pkg/annotation/annotation.go
@@ -10,4 +10,8 @@ const (
 	// CordonUntil is the name of the annotation that indicates
 	// the expiration date for this cordon rule.
 	CordonUntil = "app-operator.giantswarm.io/cordon-until"
+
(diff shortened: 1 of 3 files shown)


rossf7:
E2E now green.

PR #154, opened 2019-10-10
Search the latest catalog

tomahawk28:
Toward giantswarm/giantswarm#6824

Following the PR after https://github.com/giantswarm/appcatalog/pull/11


integration/test/app/basic/simple_test.go
```diff
@@ -90,11 +91,10 @@ func TestAppLifecycle(t *testing.T) {
 
 		var tarballPath string
 		{
-			// TODO: Remove hard-coded tarballURL when VOO feature are ready to use.
-			//
-			//	See https://github.com/giantswarm/giantswarm/issues/6824
-			//
-			tarballURL := "https://giantswarm.github.io/default-catalog/chart-operator-0.10.6.tgz"
+			tarballURL, err := appcatalog.GetLatestChart(ctx, "https://giantswarm.github.com/default-catalog", "chart-operator")
```
tomahawk28: No other way to get appcatalog storage URL in this case. So we are using hard-coding.

kopiczko: Maybe it's an idea to replace full URL with just the catalog name? Or we want to keep the library generic?

tomahawk28: In the previous PR, I was about to adding a catalog name on this function instead of a URL but revert it in favor of a storage URL. I think makes it like this generic is more usable for later use.

rossf7: As long the catalog name and the base URL as both parameters I'm fine with this. The library should work with the upgrade helm catalogs for example.

kopiczko: I don't have a strong opinion here. Just a thought.

rossf7: I think its OK with the storageURL and keeps it generic. But we should move the string to the key package. So we can reuse it between tests.

Like we do for release names.
https://github.com/giantswarm/app-operator/blob/master/integration/key/release_name.go

WDYT?

kopiczko: 👍

integration/key/release_name.go
```diff
@@ -10,6 +10,10 @@ func CustomResourceReleaseName() string {
 	return "apiextensions-app-e2e-chart"
 }
 
+func DefaultCatalogStorageURL() string {
+	return "https://giantswarm.github.com/default-catalog"
+}
```
rossf7: Let's move this to `catalog.go`. Feels weird to have it in release_name.go.

PR #155, opened 2019-10-14
Cancel reconciliation in chartoperator resource

rossf7:
Follow up to #142

### DIFF
--- a/service/controller/app/v1/resource/chartoperator/resource.go
+++ b/service/controller/app/v1/resource/chartoperator/resource.go
@@ -11,7 +11,7 @@ import (
 	"github.com/giantswarm/backoff"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
-	"github.com/giantswarm/operatorkit/controller/context/resourcecanceledcontext"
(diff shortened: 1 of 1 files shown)


service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -265,8 +265,9 @@ func (r *Resource) getAppCatalogCR(ctx context.Context, chartOperatorAppCR *v1al
 		appCatalogCR, err = r.g8sClient.ApplicationV1alpha1().AppCatalogs().Get(catalogName, metav1.GetOptions{})
 		if apierrors.IsNotFound(err) {
 			r.logger.LogCtx(ctx, "level", "debug", "message", "can't find appCatalog CR")
-			r.logger.LogCtx(ctx, "level", "debug", "message", "cancelling the resource")
-			resourcecanceledcontext.SetCanceled(ctx)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling the reconciliation")
+			reconciliationcanceledcontext.SetCanceled(ctx)
```
rossf7: We can't continue in either of these cases. So it's better to cancel the reconciliation.

PR #161, opened 2019-10-28
Cancel the reconciliation when no running tiller pod in TC

tomahawk28:
Following https://github.com/giantswarm/helmclient/pull/131

### DIFF
--- a/service/controller/app/v1/resource/tiller/resource.go
+++ b/service/controller/app/v1/resource/tiller/resource.go
@@ -65,7 +65,7 @@ func (r *Resource) ensureTillerInstalled(ctx context.Context, helmClient helmcli
 		reconciliationcanceledcontext.SetCanceled(ctx)
 
 		return nil
-	} else if helmclient.IsTillerStartingError(err) {
+	} else if helmclient.IsTillerNotRunningError(err) {
 		r.logger.LogCtx(ctx, "level", "debug", "message", "no running tiller pod")
 
 		// Can't find a tiller pod in starting phase, We will retry on next reconciliation loop.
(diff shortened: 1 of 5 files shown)


service/controller/app/v1/resource/tiller/resource.go
```diff
@@ -64,6 +64,14 @@ func (r *Resource) ensureTillerInstalled(ctx context.Context, helmClient helmcli
 		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
 		reconciliationcanceledcontext.SetCanceled(ctx)
 
+		return nil
+	} else if helmclient.IsTillerStartingError(err) {
```
tomahawk28: @rossf7 As we discussed in the previous PR, I added in here.

rossf7:
```suggestion
	} else if helmclient.IsTillerNotRunningError(err) {
```

The logic is fine here but the naming I find confusing.

Pods do have a phase but it is running not started. I'd like us to align this with the upstream terminology.

```
kk get po tiller-deploy-584df9b797-n47kw -o yaml | grep phase
  phase: Running
```

service/controller/app/v1/resource/tiller/resource.go
```diff
@@ -64,6 +64,14 @@ func (r *Resource) ensureTillerInstalled(ctx context.Context, helmClient helmcli
...
+		r.logger.LogCtx(ctx, "level", "debug", "message", "no tiller pod in starting phase")
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "no running tiller pod")
```

service/controller/app/v1/resource/tiller/resource.go
tomahawk28: Sure I will change the helmclient library first then.

PR #163, opened 2019-11-07
Add network policy

glitchcrab:
towards giantswarm/giantswarm/issues/4383

external egress is required.

### DIFF
--- a/helm/app-operator-chart/templates/network-policy.yaml
+++ b/helm/app-operator-chart/templates/network-policy.yaml
@@ -1,0 +1,20 @@
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  namespace: {{ .Values.namespace }}
+  name: {{ .Values.name }}-network-policy
+spec:
+  podSelector:
+    matchLabels:
+      app: {{ .Values.name }}
+  ingress:
+  - ports:
+    - port: {{ .Values.port }}
+      protocol: {{ .Values.protocol }}
+  egress:
+  - to:
+    - ipBlock:
+        cidr: 0.0.0.0/0
+  policyTypes:
+  - Egress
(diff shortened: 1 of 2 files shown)


helm/app-operator-chart/templates/network-policy.yaml
```diff
@@ -0,0 +1,20 @@
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  namespace: {{ .Values.namespace }}
+  name: {{ .Values.name }}-network-policy
```
kopiczko: Isn't it good to add `app` label also here?

corest: Even more, like this https://github.com/giantswarm/net-exporter/blob/master/helm/net-exporter/templates/np.yaml#L7-L8

helm/app-operator-chart/templates/network-policy.yaml
```diff
@@ -0,0 +1,20 @@
...
+    matchLabels:
+      app: {{ .Values.name }}
+  ingress:
+  - ports:
+    - port: {{ .Values.port }}
+      protocol: {{ .Values.protocol }}
+  egress:
+  - to:
+    - ipBlock:
+        cidr: 0.0.0.0/0
```
tomahawk28: I'm sorry to ask, but does it basically open to all address?

corest: Yes.
Also, we're usually doing just `{}`, so maybe let's stick to single style

rossf7:
@glitchcrab I think this is done. Can this PR be closed?

glitchcrab:
@rossf7 it looks that way - i'll rebase and re-deploy to test again

PR #169, opened 2019-11-26
Tidy up resource cancellation logic

rossf7:
Same as https://github.com/giantswarm/chart-operator/pull/317.

### DIFF
--- a/service/controller/app/v1/resource/status/create.go
+++ b/service/controller/app/v1/resource/status/create.go
@@ -7,7 +7,6 @@ import (
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/errors/tenant"
 	"github.com/giantswarm/microerror"
-	"github.com/giantswarm/operatorkit/controller/context/resourcecanceledcontext"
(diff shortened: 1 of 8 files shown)


service/controller/app/v1/resource/status/create.go
```diff
@@ -40,16 +37,12 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chart %#q in namespace %#q", cr.Name, r.chartNamespace))
 		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
-
 		return nil
 	} else if tenant.IsAPINotAvailable(err) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", "tenant cluster is not available.")
-
 		// We should not hammer tenant API if it is not available, the tenant cluster
 		// might be initializing. We will retry on next reconciliation loop.
-		resourcecanceledcontext.SetCanceled(ctx)
```
rossf7: Status uses the simple resource so returning nil is enough.

PR #172, opened 2019-12-02
Update project wiring

rossf7:
Same as https://github.com/giantswarm/chart-operator/pull/319.

Also removes unused setting from the index resource which was removed ages ago.


flag/service/service.go
```diff
@@ -3,14 +3,12 @@ package service
...
 	"github.com/giantswarm/operatorkit/flag/service/kubernetes"
 
-	"github.com/giantswarm/app-operator/flag/service/appcatalog"
 	"github.com/giantswarm/app-operator/flag/service/chart"
 	"github.com/giantswarm/app-operator/flag/service/image"
 )
 
 // Service is an intermediate data structure for command line configuration flags.
 type Service struct {
-	AppCatalog appcatalog.AppCatalog
```
rossf7: This was used by the index resource which was removed ages ago.

PR #173, opened 2019-12-05
Use k8sclient collection in wiring

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7963

Uses k8sclient collection. This prep for later PRs to update operatorkit internals.

### DIFF
--- a/service/controller/app/app.go
+++ b/service/controller/app/app.go
@@ -2,26 +2,22 @@ package app
 
 import (
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
-	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
+	"github.com/giantswarm/k8sclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	"github.com/giantswarm/operatorkit/client/k8scrdclient"
 	"github.com/giantswarm/operatorkit/controller"
 	"github.com/giantswarm/operatorkit/informer"
 	"github.com/spf13/afero"
-	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
-	"k8s.io/client-go/kubernetes"
 
 	"github.com/giantswarm/app-operator/pkg/project"
 	v1 "github.com/giantswarm/app-operator/service/controller/app/v1"
 )
 
 type Config struct {
-	Fs           afero.Fs
-	G8sClient    versioned.Interface
-	K8sClient    kubernetes.Interface
-	K8sExtClient apiextensionsclient.Interface
-	Logger       micrologger.Logger
+	Fs        afero.Fs
+	K8sClient k8sclient.Interface
+	Logger    micrologger.Logger
 
 	ChartNamespace string
 	ImageRegistry  string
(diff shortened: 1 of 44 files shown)


vendor/github.com/giantswarm/k8sclient/Gopkg.toml
```diff
@@ -42,10 +42,6 @@
   branch = "master"
   name = "github.com/giantswarm/micrologger"
 
-[[constraint]]
-  branch = "master"
```
tomahawk28: I don't understand how it got deleted, we still use operatorkit by vendoring, don't we?

service/service_test.go
```diff
@@ -1,15 +1,38 @@
...
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.Write(pods)
+	}
+	ts := httptest.NewServer(http.HandlerFunc(h))
+	defer ts.Close()
```
kopiczko: This part I don't get. Is the `k8sclient` lib setting up helm client as well?

rossf7: Sorry this was a bit rough. With the k8sclient collection the service test errors if a JSON response is not returned. This was copy pasta from chart-operator.

We know these service tests UTs can get complex. I'll update the comments but maybe it should be dropped like we did in other projects?

vendor/github.com/giantswarm/k8sclient/Gopkg.toml
rossf7: This is in the vendor dir. I just ran `dep ensure -update github.com/giantswarm/k8sclient`

Maybe this is from the circular dependencies we removed.

service/service_test.go
rossf7: Tidied up the UT a bit and fixed the comments.

kopiczko: Do you know what produces the request?

kopiczko: BTW it's not a blocker. Merge if you need. We con continue the discussion.

rossf7: @kopiczko Sorry was getting the rest of my PRs sorted. Its super weird it seems to be doing a DNS lookup because of https://github.com/giantswarm/k8sclient/blob/059a94fd04e59aceab384413cf9608ae1640a2ca/clients.go#L115.

I simplified to just return empty JSON as that is the minimum it will accept.

```
{"caller":"github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/client/k8srestconfig/k8s_rest_config.go:158","level":"debug","message":"creating out-cluster REST config","time":"2019-12-05T15:15:36.532146+00:00"}
{"caller":"github.com/giantswarm/app-operator/vendor/github.com/giantswarm/operatorkit/client/k8srestconfig/k8s_rest_config.go:173","level":"debug","message":"created out-cluster REST config","time":"2019-12-05T15:15:36.532243+00:00"}
--- FAIL: Test_Service_New (0.04s)
    --- FAIL: Test_Service_New/case_0:_valid_config_returns_no_error (0.04s)
        /Users/rossf7/.gvm/pkgsets/go1.13.1/global/src/github.com/giantswarm/app-operator/service/service_test.go:77: error == [{/Users/rossf7/.gvm/pkgsets/go1.13.1/global/src/github.com/giantswarm/app-operator/service/service.go:91: } {/Users/rossf7/.gvm/pkgsets/go1.13.1/global/src/github.com/giantswarm/app-operator/vendor/github.com/giantswarm/k8sclient/clients.go:117: } {Get http://kubernetes/api?timeout=10s: dial tcp: lookup kubernetes: no such host}], want nil
```

Going to go with this but maybe we can fix it in k8sclient somehow?

service/controller/app/app.go
```diff
@@ -112,7 +101,7 @@ func NewApp(config Config) (*App, error) {
 			ResourceSets: []*controller.ResourceSet{
 				resourceSetV1,
 			},
-			RESTClient: config.G8sClient.CoreV1alpha1().RESTClient(),
+			RESTClient: config.K8sClient.G8sClient().CoreV1alpha1().RESTClient(),
```
rossf7: The e2e tests broke when I used `config.K8sClient.RESTClient()` so I suspect something is wrong with the scheme builder.

I've reverted as I don't have time to investigate further. I'm off tomorrow so I need to get my PRs merged.

rossf7:
Tested on `godsmack`.

service/controller/app/app.go
xh3b4sd: No worries. We dropped the rest client there anyway. Will be gone in the next step.

tomahawk28:
Some wiring is still missing, I'm going to take over this PR for further wiring & implementation.

PR #174, opened 2019-12-06
Vendor operatorkit

tomahawk28:
Towards giantswarm/giantswarm#7963
After update k8sclient, we now need to use the latest operatorkit.


integration/setup/config.go
```diff
@@ -19,7 +18,7 @@ const (
 
 type Config struct {
 	HelmClient *helmclient.Client
-	K8s        *k8s.Setup
+	K8s        *k8sclient.Setup
```
tomahawk28: these are leftover from the previous PR.

tomahawk28:
Tested on `ghost`

PR #168, opened 2019-11-22
Vendor dependencies, update k8s to 1.16.3

sslavic:
Towards giantswarm/giantswarm#7675

PR #175, opened 2019-12-11
Helmclient k8s 1 16

tomahawk28:


sslavic:
@tomahawk28 why new PR? We already have https://github.com/giantswarm/app-operator/pull/168

tomahawk28:
@sslavic I would like to test it before I have it on that branch, but forgot to change base branch as k8s-1-16. is that okay for you?

tomahawk28:
... and in this PR, helmclient pointing another branch, not `master` .

PR #176, opened 2019-12-16
Delete unused watch_namespace var

tomahawk28:
`WatchNamespace` is not used anywhere. 
We can safely delete this.


tomahawk28:
Tested on geckon.

PR #179, opened 2019-12-20
Fix problems deleting chart CR, configmap and secrets

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/8187

### DIFF
--- a/integration/key/resource_name.go
+++ b/integration/key/resource_name.go
@@ -1,0 +1,7 @@
+// +build k8srequired
+
+package key
+
+func ChartConfigMapName() string {
+	return "kubernetes-test-app-chart-chart-values"
+}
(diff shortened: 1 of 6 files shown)


sslavic:
https://github.com/giantswarm/giantswarm/issues/8187 reproduced https://circleci.com/gh/giantswarm/app-operator/2440

rossf7:
Tested on `gauss`.

sslavic:
What do we do about existing ConfigMap and Secret orphaned resources laying around, labeled as managed by app-operator but not actually part of any existing App?

rossf7:
@sslavic Good point on the orphan resources. I think we should add metrics for orphan configmaps and secrets to the app-operator collector.

I can work on the cleanup when I'm back on Friday.

PR #181, opened 2019-12-23
Adjust e2e test timeout from default 10m to 20m

sslavic:
Basic e2e test often fails due to go test timeout (e.g. see [here](https://circleci.com/gh/giantswarm/app-operator/2453)), it seems to take longer. This PR extends the timeout to 20min.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ e2eTest: &e2eTest
             -e E2E_KUBECONFIG=/e2e/kubeconfig \
             -e CIRCLE_SHA1=${CIRCLE_SHA1} \
             -w /go/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME} \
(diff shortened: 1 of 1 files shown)


sslavic:
@rossf7 can this particular e2e test be sped up by reordering installation steps or operatorkit changed to eagerly resync on startup? When I was looking into it a bit recently, it seemed Chart CR for an App gets created before chart-operator gets installed; it seems the operator would get triggered only after resync, and not eagerly on startup.

rossf7:
@sslavic I want to rewrite the tests to just create the CRs using client-go without all the complicated helm chart to create CRs stuff. To me it now looks very over engineered.

Was the problem you saw with the kubeconfig test? It has a 2nd KIND cluster to simulate a tenant cluster. Process should be app-operator installs chart-operator and creates the chart CRD.

If the chart CRD is being created by the test setup then yes maybe the chart CR is being created too early.

sslavic:
Thumbs up for giving it a try to simplify tests, client-go/apiextensions may be enough to drive the test.

PR #182, opened 2019-12-30
Rename resource to tcnamespace and use simple resource

rossf7:
When the namespace resource was added to cluster-operator the CRUD resource was used but only create is needed. The code was then moved to app-operator.

This drops the `namespace` resource and adds a `tcnamespace` resource using the simple interface. The naming matches the `cpnamespace` resource in cluster-operator.


service/controller/app/v1/resource_set.go
```diff
@@ -235,12 +230,17 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 	}
 
 	resources := []resource.Interface{
+		// Following resources manage controller context information.
```
rossf7: I copied the commenting style in cluster-operator.

rossf7:
Tested on `ghost`.

PR #183, opened 2020-01-02
Rename file to align with test name

rossf7:


rossf7:
Integration tests are broken because the test app doesn't work with K8s 1.16.

Created https://github.com/giantswarm/giantswarm/issues/8217 to change tests to use kube-state-metrics and remove hardcoded versions.

sslavic:
I've released new versions of test app, k8s 1.16 compatible, but yeah since test app is actually kube-state-metrics it's better to replace test app with kube-state-metrics-app, and archive test app repo, so we don't have to maintain both.

rossf7:
Thanks Stevo I tested it with the release you created but it needs 2 releases for testing the update.

So yes I think it's better to clean up the tests properly and remove the usage of the sample catalog. By extending the appcatalog library to have GetPreviousVersion we can also remove the hardcoded versions.

sslavic:
I did create two release 0.7.0 which makes test app compatible with k8s 1.16 and 0.7.1 with essentially just version change, for testing purposes.
That part, where we have to have two versions for testing upgrade make me a bit reluctant to use kube-state-metrics-app for testing - at some point we may again need two versions, one actual and some artificial one; for app dedicated for testing it's ok to make artificial versions, but not so for actually used app like kube-state-metrics-app.

rossf7:
Yes its true we need 2 working versions to test the upgrade. So maybe a fake app is better. But anyway we need to move off sample catalog and the repo is already archived.

sslavic:
When publishing those two releases, I temporarily unarchived sample catalog, and archived it again after the two releases got published there.

PR #184, opened 2020-01-02
Prevent reconciliation getting blocked by broken tiller in tenant cluster

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/8190

Uses the backoff setting in https://github.com/giantswarm/helmclient/pull/146 and reduces the max wait for installing tiller to 30 seconds.

Also now the tcnamespace and tiller resources only execute for chart-operator app CRs. So if a tenant cluster is unavailable we only try to install tiller once.


Gopkg.toml
```diff
@@ -58,7 +58,7 @@ required = [
 
 [[constraint]]
   name = "github.com/giantswarm/e2e-harness"
-  branch = "master"
+  branch = "reduce-backoff-time"
```

Gopkg.toml
```diff
@@ -78,7 +78,7 @@ required = [
 
 [[constraint]]
   name = "github.com/giantswarm/helmclient"
-  branch = "master"
+  branch = "reduce-backoff-time"
```

service/controller/app/v1/resource/clients/resource.go
```diff
@@ -125,9 +126,10 @@ func (r *Resource) addClientsToContext(ctx context.Context, cr v1alpha1.App) err
 			K8sClient: cc.K8sClient,
 			Logger:    r.logger,
 
-			RestConfig:          restConfig,
-			TillerImageRegistry: r.imageRegistry,
-			TillerNamespace:     r.tillerNamespace,
+			EnsureTillerInstalledMaxWait: 30 * time.Second,
```
rossf7: Reduces the max wait from 2 mins which is the default. This is needed for integration tests but is far too long here.

service/controller/app/v1/resource/tcnamespace/create.go
```diff
@@ -32,6 +32,14 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return nil
 	}
 
+	// Resource is used to bootstrap chart-operator in tenant clusters.
+	// So for other apps we can skip this step.
```
rossf7: The `giantswarm` namespace only needs to be created for the chart-operator app. As this is part of the bootstrap process.

service/controller/app/v1/resource/tiller/create.go
```diff
@@ -22,7 +22,15 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 	if key.InCluster(cr) {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q in %#q uses InCluster kubeconfig no need to install tiller", cr.Name, cr.Namespace))
-		r.logger.LogCtx(ctx, "level", "debug", "message", "cancelling the resource")
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+		return nil
+	}
+
+	// Resource is used to bootstrap chart-operator in tenant clusters.
+	// So for other apps we can skip this step.
```
rossf7: This is the actual fix. If tiller is broken we only wait for the chart-operator app.

rossf7:
Tested on `ghost`.

rossf7:
> So it still gets blocked for a little bit if chart-operator is not deployed 

Yes with this we still wait 30 seconds per tenant cluster. But it's a lot better than now which is 2 mins per default app. 😬 

We can iterate further if needed but this mess goes away with Helm 3.

PR #185, opened 2020-01-03
Update e2e tests for k8s 1.16 and KIND changes

rossf7:
The test-app uses the new 0.7.0 and 0.7.1 releases prepared by @sslavic that work with K8s 1.16.

I'd like us to also stop using the sample catalog but this gets the tests passing.

### DIFF
--- a/integration/test/app/kubeconfig/kubeconfig_test.go
+++ b/integration/test/app/kubeconfig/kubeconfig_test.go
@@ -93,7 +93,7 @@ func TestAppLifecycleUsingKubeconfig(t *testing.T) {
 		}
 
 		// Normally KIND assign 127.0.0.1 as server address, that should change into kubernetes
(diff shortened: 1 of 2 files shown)


integration/test/app/kubeconfig/kubeconfig_test.go
```diff
@@ -93,7 +93,7 @@ func TestAppLifecycleUsingKubeconfig(t *testing.T) {
 		}
 
 		// Normally KIND assign 127.0.0.1 as server address, that should change into kubernetes
-		c.Clusters["kind"].Server = "https://kubernetes.default.svc.cluster.local"
+		c.Clusters["kind-kind"].Server = "https://kubernetes.default.svc.cluster.local"
```
rossf7: The kubeconfig format changed in the latest KIND release.

PR #188, opened 2020-01-07
Copy annotations into chart CR with chart-operator prefix

tomahawk28:
During looking into https://github.com/giantswarm/giantswarm/issues/8082, I found out we do not compare annotations between desired and current status. 

We would copy annotations from app CR only with `chart-operator.giantswarm.io` prefix and compare the value as well.

### DIFF
--- a/pkg/annotation/annotation.go
+++ b/pkg/annotation/annotation.go
@@ -3,6 +3,9 @@
 package annotation
 
 const (
+	// ChartOperatorPrefix is the prefix of the annotation that controls the logic inside chart-operator
+	ChartOperatorPrefix = "chart-operator.giantswarm.io"
+
 	// CordonReason is the name of the annotation that indicates
 	// the reason of why app-operator should not apply any update to this app CR.
 	CordonReason = "app-operator.giantswarm.io/cordon-reason"
@@ -10,8 +13,4 @@ const (
 	// CordonUntil is the name of the annotation that indicates
 	// the expiration date for this cordon rule.
 	CordonUntil = "app-operator.giantswarm.io/cordon-until"
-
-	// ForceHelmUpgrade is the name of the annotation that controls whether force
-	// is used when upgrading the Helm release.
-	ForceHelmUpgrade = "chart-operator.giantswarm.io/force-helm-upgrade"
 )
--- a/pkg/annotation/annotation.go
+++ b/pkg/annotation/annotation.go
@@ -3,7 +3,7 @@
 package annotation
 
 const (
-	// ChartOperatorPrefix is the prefix of the annotation that controls the logic inside chart-operator
+	// ChartOperatorPrefix is the prefix for annotations that control logic inside chart-operator.
 	ChartOperatorPrefix = "chart-operator.giantswarm.io"
 
 	// CordonReason is the name of the annotation that indicates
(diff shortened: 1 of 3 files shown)


tomahawk28:
Tested on `gauss`

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -73,11 +74,11 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
 	annotations := map[string]string{}
 
-	// ForceHelmUpgrade has been set for this app CR so this needs to be passed
-	// on to the chart CR.
-	val, ok := input[annotation.ForceHelmUpgrade]
-	if ok {
-		annotations[annotation.ForceHelmUpgrade] = val
+	for k, val := range input {
+		// Copy all annotations which has a prefix with chart-operator.giantswarm.io.
+		if strings.HasPrefix(k, annotation.ChartOperatorPrefix) {
```
tomahawk28: Better to copy all annotation related to `chart-operator`

pkg/annotation/annotation.go
```diff
@@ -3,15 +3,14 @@
 package annotation
 
 const (
+	// ChartOperatorPrefix is the prefix of the annotation that controls the logic inside chart-operator
```
rossf7:
```suggestion
	// ChartOperatorPrefix is the prefix for annotations that control logic inside chart-operator.
```

service/controller/app/v1/resource/chart/desired.go
```diff
@@ -73,11 +74,11 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 func generateAnnotations(input map[string]string) map[string]string {
 	annotations := map[string]string{}
 
-	// ForceHelmUpgrade has been set for this app CR so this needs to be passed
-	// on to the chart CR.
-	val, ok := input[annotation.ForceHelmUpgrade]
-	if ok {
-		annotations[annotation.ForceHelmUpgrade] = val
+	for k, val := range input {
```
rossf7:
```suggestion
	for k, v := range input {
```

Let's try and be consistent with var naming. This is shorter but I'd also be fine with `key` and `val`.

service/controller/app/v1/resource/chart/resource.go
```diff
@@ -76,6 +79,19 @@ func equals(current, desired *v1alpha1.Chart) bool {
 	if !reflect.DeepEqual(current.Labels, desired.Labels) {
 		return false
 	}
+	for k, desiredValue := range desired.Annotations {
```
rossf7:
```suggestion

	for k, desiredValue := range desired.Annotations {
```

service/controller/app/v1/resource/chart/resource.go
```diff
@@ -76,6 +79,19 @@ func equals(current, desired *v1alpha1.Chart) bool {
...
+		if !strings.HasPrefix(k, annotation.ChartOperatorPrefix) {
+			continue
+		}
+
+		if currentValue, ok := current.Annotations[k]; !ok {
+			return false
+		} else {
```
rossf7: We return above. So do we need the nested if here?

service/controller/app/v1/resource/chart/desired.go
tomahawk28: Agreed, shorther version I would like.

PR #189, opened 2020-01-07
Only install chart-operator for its own app CR

rossf7:
Follow up to https://github.com/giantswarm/app-operator/pull/184. We only need to install chart-operator for its own app CR.

### DIFF
--- a/service/controller/app/v1/resource/chartoperator/create.go
+++ b/service/controller/app/v1/resource/chartoperator/create.go
@@ -24,9 +24,16 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	}
 
 	if key.InCluster(cr) {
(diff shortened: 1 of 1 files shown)


service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -24,12 +24,19 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	}
 
 	if key.InCluster(cr) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q in %#q uses InCluster kubeconfig no need to install chart operator", cr.Name, cr.Namespace))
-
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q uses InCluster kubeconfig no need to install chart operator", key.AppName(cr)))
```
rossf7: Just aligning the log messages.

PR #190, opened 2020-01-17
Only install chart-operator for its own app CR

rossf7:
For the tcnamespace and tiller resources we cancel the resource unless the app CR is for chart-operator. But the chartoperator namespace got missed.

This resource is only used for bootstrapping chart-operator so this is more efficient. But a better design might be to move the bootstrap process to a new operator. I proposed this in https://github.com/giantswarm/giantswarm/issues/8448 after discussions with Tim during Operator Week.

### DIFF
--- a/service/controller/app/v1/resource/chartoperator/create.go
+++ b/service/controller/app/v1/resource/chartoperator/create.go
@@ -24,9 +24,16 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	}
 
 	if key.InCluster(cr) {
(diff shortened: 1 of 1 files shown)


service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -24,12 +24,19 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	}
 
 	if key.InCluster(cr) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q in %#q uses InCluster kubeconfig no need to install chart operator", cr.Name, cr.Namespace))
-
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q uses InCluster kubeconfig no need to install chart operator", key.AppName(cr)))
```
tomahawk28: <3

PR #195, opened 2020-01-29
vendor: updated helmclient

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/8479

Needs https://github.com/giantswarm/helmclient/pull/158 and uses context.WithTimeout so we only wait for 10 secs when finding out if tiller is installed.


rossf7:
Closing in favor of #196.

PR #196, opened 2020-01-30
Don't try to create tunnel when tiller is outdated

rossf7:
Towards giantswarm/giantswarm#8479

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -76,9 +76,9 @@ required = [
   name = "github.com/giantswarm/exporterkit"
   branch = "master"
 
-[[constraint]]
+[[override]]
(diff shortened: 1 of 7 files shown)


Gopkg.toml
```diff
@@ -76,9 +76,9 @@ required = [
   name = "github.com/giantswarm/exporterkit"
   branch = "master"
 
-[[constraint]]
+[[override]]
```

Gopkg.toml
```diff
@@ -76,9 +76,9 @@ required = [
...
   name = "github.com/giantswarm/helmclient"
-  branch = "master"
+  branch = "outdated-tiller"
```

rossf7:
One integration test failed due to error creating kind cluster. Already tested manually. Going with this.

PR #197, opened 2020-02-05
Use Circle Orb for integration tests

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/4606

~~Adds an inline orb that implements a new integration-test command. I'd like to get feedback here first. Once approved I'll integrate it into architect-orb.~~

**EDIT:** I'll open a new PR in architect-orb. If you have any feedback on the approach you can leave it here.

This is based on what Pawel created in https://github.com/giantswarm/release-operator/pull/133.

## Basic test

```
      - inline-orb/integration-test:
          name: basic-integration-test
          test-dir: "./integration/test/app/basic"
          requires:
            - build
```

## Test with custom config

```
      - inline-orb/integration-test:
          name: kubeconfig-integration-test
          kind-config: "./integration/config/kind-config.yaml"
          setup-script: "./integration/config/setup.sh"
          test-dir: "./integration/test/app/kubeconfig"
          requires:
            - build
```

Tests can be configured by passing a kind config file or a setup script. The setup script is executed before the test is executed.


.circleci/config.yml
```diff
@@ -1,70 +1,168 @@
...
+      create-cluster:
+        parameters:
+          kind-config:
+            type: string
+        steps:
+          - run:
+              name: "Create cluster"
+              command: |
+                cd ~/project
+
```
rossf7: We can pass a kind config file to the test. Otherwise a minimal config is generated.

.circleci/config.yml
```diff
@@ -1,70 +1,168 @@
...
+      test-setup:
+        parameters:
+          setup-script:
+            type: string
+        steps:
+          - run:
+              name: "Run test setup commands"
+              command: |
+                if [ -e << parameters.setup-script >> ]; then
+                  echo << parameters.setup-script >>
```
MarcelMue: Why echo the script? Just for debug purposes?

integration/config/setup.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+kubectl label namespace default foo=bar
```
MarcelMue: Leftover?

.circleci/config.yml
rossf7: Yes its to show which commands are run. Otherwise you just see the output.

integration/config/setup.sh
rossf7: It's just some dummy commands to test that the setup script param works.

MarcelMue: For me it would make sense to move this to the `architect` orb then and use it from there. Or do you want to merge it here?

rossf7: Yes sure I'll move it.

It can't be merged here. It's just the inline orb is a faster feedback loop if we wanted to make changes.

.circleci/config.yml
```diff
@@ -1,70 +1,168 @@
...
+                  sh << parameters.setup-script >>
+                fi
```
kopiczko: For this I'd like to use `when:` and then simply run the script if it's set. We can also print the script name in the `name:`.

.circleci/config.yml
```diff
@@ -1,70 +1,168 @@
...
+
+      go-test:
```
kopiczko: Command names has to be unique in the orb so it will have to be renamed while moving.

.circleci/config.yml
```diff
@@ -1,70 +1,168 @@
...
+        parameters:
+          env:
+            type: string
+          test-dir:
+            type: string
+        steps:
+          - run:
+              name: "Create .env file"
+              command: |
+                echo '<< parameters.env >>' | sed -e 's/^\(.\)/export \1/' > .env
```
kopiczko: I recently realised using `tee` for those things is very useful for debugging.

```suggestion
                echo '<< parameters.env >>' | sed -e 's/^\(.\)/export \1/' | tee .env
```

.circleci/config.yml
```diff
@@ -1,70 +1,168 @@
...
+              # TODO remove this part after switch to Go modules.
+              name: "Move code to $GOPATH"
+              command: |
+                mkdir -p $GOPATH/src/github.com/${CIRCLE_PROJECT_USERNAME} && mv ~/project $GOPATH/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
+          - run:
+              name: "Run test"
+              command: |
+                cd $GOPATH/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
+                source .env
+                CGO_ENABLED=0 E2E_KUBECONFIG=~/.kube/config go test -tags k8srequired -timeout 20m -v << parameters.test-dir >>
```
kopiczko: I'd like all steps to basically be oneliners. That's how the orb is designed and that makes it easy to debug things (don't care about the `cd` part because it goes away anyway). 

```suggestion
                # TODO remove `cd` line after switch to Go modules.
                cd $GOPATH/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
                source .env && CGO_ENABLED=0 E2E_KUBECONFIG=~/.kube/config go test -tags k8srequired -timeout 20m -v << parameters.test-dir >>
```

.circleci/config.yml
```diff
@@ -1,70 +1,168 @@
...
+          KUBERNETES_VERSION: v1.16.3
+        parameters:
+          env:
+            default: ""
+            description: |
+              Environment variables configuration. Must be in following format:
+
+                  VAR_A="VALUE_A"
+                  VAR_B="VALUE_B"
+            type: string
```
kopiczko: Do you think it makes more sense to keep those env vars here in the circle configuration or rather specifying it as a path to a file?

.circleci/config.yml
rossf7: Sweet! I didn't know that was possible. This can be used in a few places.

.circleci/config.yml
rossf7: Cool I'll get that sorted.

.circleci/config.yml
rossf7: Yes I think a file is better and matches the other params.

rossf7:
Stale.

PR #198, opened 2020-02-06
Use context.WithCancel in resources that connect to tenant clusters

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/8591

We are seeing delays in processing optional apps in control planes with a lot of tenant clusters. Some tenant clusters may be unavailable due to new clusters, upgrades or broken clusters.

- This uses context.WithCancel to fail after 3 secs rather than waiting 30 secs for the tenant.IsAPINotAvailable error.
- If the configmap resource fails then the loop is cancelled. As following resources are likely to also fail.


service/controller/app/v1/resource/configmap/current.go
```diff
@@ -36,23 +39,36 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
 	if apierrors.IsNotFound(err) {
 		// Return early as configmap does not exist.
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find configmap %#q in namespace %#q", name, r.chartNamespace))
 		return nil, nil
 	} else if tenant.IsAPINotAvailable(err) {
-		// We should not hammer tenant API if it is not available, the tenant cluster
-		// might be initializing. We will retry on next reconciliation loop.
+		// We should not hammer tenant API if it is not available. We cancel
+		// the reconciliation because its likely following resources will also
+		// fail.
```
rossf7: The configmap resource is followed by the secret, chart and status resources that connect to tenant clusters. So we cancel the reconciliation since they are likely to also fail.

service/controller/app/v1/resource/secret/current.go
```diff
@@ -36,7 +38,10 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding secret %#q in namespace %#q", name, r.chartNamespace))
 
-	chart, err := cc.K8sClient.CoreV1().Secrets(r.chartNamespace).Get(name, metav1.GetOptions{})
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
+	defer cancel()
+
+	secret, err := cc.K8sClient.CoreV1().Secrets(r.chartNamespace).Get(name, metav1.GetOptions{})
```
rossf7: Fixing the copy pasta.

service/controller/app/v1/resource/status/create.go
```diff
@@ -2,12 +2,14 @@ package status
...
 	"context"
+	"errors"
 	"fmt"
+	"time"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/errors/tenant"
 	"github.com/giantswarm/microerror"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
rossf7: Fixing the alias.

rossf7:
Tested on `ghost`.

service/controller/app/v1/resource/chart/current.go
```diff
@@ -50,6 +55,12 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	} else if err != nil {
 		return nil, microerror.Mask(err)
 	}
+	if errors.Is(ctx.Err(), context.DeadlineExceeded) {
```
tomahawk28: Does it really work? doesn't we have to wrap it with `select` statement so it would return timeout errors by a channel?
Example in golang docs; 
```go
package main

import (
	"context"
	"fmt"
	"time"
)

func main() {
	// Pass a context with a timeout to tell a blocking function that it
	// should abandon its work after the timeout elapses.
	ctx, cancel := context.WithTimeout(context.Background(), 50*time.Millisecond)
	defer cancel()

	select {
	case <- time.After(1 * time.Second):
		fmt.Println("overslept")
	case <-ctx.Done():
		fmt.Println(ctx.Err()) // prints "context deadline exceeded"
	}

}
```

I did a similar thing as this code suggests here. https://play.golang.org/p/A9X7iYHCqTW
But it doesn't work.

rossf7: It worked when I tested the migration resources in cluster-operator. But I'll do some more testing tomorrow.

rossf7: @tomahawk28 Your example works for me.

This was tested in cluster-operator but doing another test to be sure. I'd prefer to keep the code as it is because its simpler.

tomahawk28: The question to me is, it would fall into context timeout error if it takes more than 30 seconds. 
The current structure suggests it would only know that context timeout happened after 30 seconds. 
Can we maybe test that situation?

rossf7: The context timeout is 3 seconds not 30 seconds. We need to use a short timeout because we need to make a lot of API calls to fetch the current state.

I'm going to test it in ginger with a working TC and a broken TC. I think that's the best way to simulate the current problems.

tomahawk28: Sorry, my knowledge around here maybe not enough. 
What I mean above, should we expect timeout (3 seconds) to happen with the `select` statement? 
```
	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
	defer cancel()
        c := make(chan int)
        go func() { 
        // Working with tenant cluster       
  //...
       // Done     
       <- interface{}
       } () 

	select {
	case <- c 
                // Done normally, let's move on. 
	case <-ctx.Done():
		// timeout happened first, canceled the reconciliation. 
	}
```

rossf7: @tomahawk28 Thanks exactly right. <3 This needed the channel and signal handling. 🤦‍♂ 

PTAL Tested on `ginger`.

service/controller/app/v1/resource/configmap/current.go
```diff
@@ -34,25 +38,82 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
 		// Return early as configmap does not exist.
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find configmap %#q in namespace %#q", name, r.chartNamespace))
 		return nil, nil
-	} else if tenant.IsAPINotAvailable(err) {
-		// We should not hammer tenant API if it is not available, the tenant cluster
-		// might be initializing. We will retry on next reconciliation loop.
+	} else if tenant.IsAPINotAvailable(res.Error) {
+		// Set status so we don't try to connect to the tenant cluster
+		// again in this reconciliation loop.
+		cc.Status.TenantCluster.IsUnavailable = true
```
tomahawk28: If we cancel the reconciliation, why we even need to set `cc.Status.TenantCluster.IsUnavailable` since it would eventually stop whole steps?

rossf7: Good point it should just cancel the resource not the whole loop. Fixed.

The following resources check the context and cancel the resource. To give us more control.

service/controller/app/v1/resource/tcnamespace/create.go
```diff
@@ -53,10 +55,40 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+			// Set status so we don't try to connect to the tenant cluster
+			// again in this reconciliation loop.
+			cc.Status.TenantCluster.IsUnavailable = true
+
+			r.logger.LogCtx(ctx, "level", "debug", "message", "timeout creating namespace")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+			return nil
+		}
+	default:
```
tomahawk28: I don't think we need `default` in here since in any case `ch` is not yet ready, then it would fall into here.

rossf7: Thanks! Fixed.

rossf7:
Final test on ginger. Going with this.

PR #199, opened 2020-02-07
Use k8sclient library in controller context

rossf7:
Some refactoring to use the `k8sclient` package when creating clients for tenant clusters. This simplifies the code and prepares for later improvements.

- I also dropped the current state tests that make heavy use of the fake k8s clients.
- We have integration tests and we should improve these. Instead of maintaining these complex unit tests.


service/controller/app/v1/controllercontext/context.go
```diff
@@ -16,9 +15,8 @@ const controllerKey contextKey = "controller"
 
 type Context struct {
 	AppCatalog v1alpha1.AppCatalog
-	G8sClient  versioned.Interface
 	HelmClient helmclient.Interface
-	K8sClient  kubernetes.Interface
+	K8sClient  k8sclient.Interface
```
rossf7: Context gets simpler.

service/controller/app/v1/resource/clients/resource.go
```diff
@@ -100,30 +101,37 @@ func (r *Resource) addClientsToContext(ctx context.Context, cr v1alpha1.App) err
 		}
 	}
 
-	restConfig, err := kubeConfig.NewRESTConfigForApp(ctx, cr)
-	if err != nil {
-		return microerror.Mask(err)
-	}
-
+	var restConfig *rest.Config
```
rossf7: Main change is to the clients resource.

PR #203, opened 2020-02-11
Use time.After to handle unresponsive tenant clusters

rossf7:
Towards giantswarm/giantswarm#8591

Replacing `context.WithTimeout` with `time.After` since client-go calls don't take a context.

See https://github.com/giantswarm/cluster-operator/pull/919#discussion_r377641875.

### DIFF
--- a/service/controller/app/v1/resource/configmap/current.go
+++ b/service/controller/app/v1/resource/configmap/current.go
@@ -2,7 +2,6 @@ package configmap
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


service/controller/app/v1/resource/configmap/current.go
```diff
@@ -46,43 +45,34 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
-			// Set status so we don't try to connect to the tenant cluster
-			// again in this reconciliation loop.
-			cc.Status.TenantCluster.IsUnavailable = true
-
-			r.logger.LogCtx(ctx, "level", "debug", "message", "timeout getting configmap")
-			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
-			resourcecanceledcontext.SetCanceled(ctx)
-			return nil, nil
-		}
+	case <-time.After(3 * time.Second):
```
MarcelMue: Seems a bit random to wait for 3 seconds - any reason for this specifically?

tomahawk28: To get TenantClusterUnavailable error in helmclient, it usually takes more than 30 seconds. By this change, we would only need 3 seconds to cancel it.

rossf7: Yes 30 seconds is too long because the operator needs to process other CRs. We can iterate on this if we need to reduce it further.

MarcelMue: My question would be why we wait at all instead of doing it in the next reconciliation period - not certain if that is better though.

I'll think about this.

PR #204, opened 2020-02-21
Structure clients in the controller context

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/8863

Prep for #205. I want to add more info to the controller context. So using the k8sclient library which means g8sClient can be removed.


rossf7:
Using k8sclient doesn't work because it makes and API call and the tenant cluster may not be up yet.

Updated to use a Clients struct which can also be used for the helm client.

rossf7:
Tested on `ginger`.

PR #205, opened 2020-02-21
Add annotations for configmap and secret resource versions

rossf7:
Towards giantswarm/giantswarm#8863

Adds annotations for `configmap-version` and `secret-version` to the configmaps and secrets generated by app-operator and consumed by chart-operator.

This reduces the time taken to reconcile config changes. As a chart CR update event will be generated which will recalculate the desired state.


rossf7:
Tested on `ghost`.

rossf7:
Thanks @tomahawk28 its been tested but I'll do some more checks after the problems we had on Friday.

The helm releases will still be updated because the CRs are configured to use the draughtsman values. The update will happen sooner due to the annotations but it would happen after the resync period anyway.

rossf7:
Stale. Will redo once operator is flat.

PR #206, opened 2020-02-24
Add notes annotation to configmaps and secrets to not edit

rossf7:
Towards giantswarm/giantswarm#8863

The `*-chart-values` and `*-chart-secrets` resources are managed by app-operator and consumed by chart-operator.

They should not be edited directly so this adds a notes annotation to make this clearer. 

```yaml
kind: ConfigMap
metadata:
  annotations:
    giantswarm.io/notes: DO NOT EDIT
  creationTimestamp: "2019-12-18T13:59:10Z"
  labels:
  name: cluster-operator-2.0.0-chart-values
  namespace: giantswarm
```


rossf7:
Tested on `ghost`.

service/controller/app/v1/resource/configmap/desired.go
```diff
@@ -55,6 +56,9 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      key.ChartConfigMapName(cr),
 			Namespace: r.chartNamespace,
+			Annotations: map[string]string{
+				annotation.Notes: "DO NOT EDIT",
```
xh3b4sd: I feel like this should be more specific. I am not sure what should not be edited now. The annotation? Something else in there? The whole CR? And then, why though?

rossf7: Yes I see what you mean. Although I got feedback from @puja108 to make the message less specific. https://github.com/giantswarm/giantswarm/issues/8863#issuecomment-589657132

The reason is because this configmap is generated by app-operator and consumed by chart-operator. So the user should edit the user configmap for the app CR.

What about changing it to  `giantswarm.io/notes: DO NOT EDIT. Values managed by app-operator.`?

xh3b4sd: I am not big about the words in particular. I just think it should transport a certain message without raising more questions. /cc @giantswarm/sig-ux

rossf7: Got feedback from SIG UX and extended the message to have more context.

@xh3b4sd PTAL

PR #215, opened 2020-03-20
dep: Use released versions of libraries

rossf7:


rossf7:
Some stray commits got in here. Closing in favor of https://github.com/giantswarm/app-operator/pull/216

PR #216, opened 2020-03-20
 dep: Use released versions of libraries 

rossf7:
Same as https://github.com/giantswarm/chart-operator/pull/383.

PR #217, opened 2020-03-23
Revert " dep: Use released versions of libraries  (#216)"

rossf7:
This reverts commit 910d178759b6a1f8fb01104191eda92dba1f2a50.

This doesn't work in anubis because its on 1.14.


rossf7:
Going with a ping.

PR #218, opened 2020-03-23
helm3: For dep use released versions of libraries

rossf7:
Continuing in helm3 branch since this is blocked in master by k8s 1.14 control planes.


rossf7:
Going with a ping.

PR #219, opened 2020-03-23
helm3: Add chartcrd resource to improve bootstrapping

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/8555

This adds a chartcrd resource and means we on longer wait in the chart-operator resource. This means we no longer need to wait for a running chart-operator pod.

So the chart-operator bootstrap process is faster and doesn't block processing other app CRs.


service/controller/app/v1/resource/chartcrd/create.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
+	"github.com/giantswarm/backoff"
+	"github.com/giantswarm/microerror"
+)
+
+// EnsureCreated ensures the chart CRD exists in the tenant cluster so the
+// chart resource can create chart CRs.
```
rossf7: This is the new resource. It runs before the chart resource so the chart CRD is always present in the tenant cluster.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -135,41 +134,35 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
...
+		// We update the helm release but with a short timeout so we don't
+		// block reconciling other CRs. This gives time to make the port
+		// forwarding connection to the Tiller API.
+		//
+		// If we do timeout the install will continue in the background.
+		// We will check the progress in the next reconciliation loop.
+		go func() {
+			err = cc.Clients.Helm.InstallReleaseFromTarball(ctx, tarballPath, namespace, helm.ReleaseName(release), helm.ValueOverrides(chartOperatorValues))
+			close(ch)
+		}()
```
rossf7: Uses the same approach as chart-operator with a short timeout. So we don't block.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -135,41 +134,35 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
...
 			return nil
 		}
 
-		// Wait for chart-operator to be deployed. If it takes longer than
-		// the timeout the chartconfig CRs will be created during the next
-		// reconciliation loop.
-		b := backoff.NewConstant(20*time.Second, 5*time.Second)
-		n := func(err error, delay time.Duration) {
-			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("%#q deployment is not ready retrying in %s", release, delay), "stack", fmt.Sprintf("%#v", err))
-		}
```
rossf7: Since the chartcrd resource creates the CRD we don't need to wait for chart-operator.

service/controller/app/v1/resource/chartcrd/create.go
```diff
@@ -0,0 +1,51 @@
+package chartcrd
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+	"github.com/giantswarm/app-operator/service/controller/app/v1/key"
```
tomahawk28: Ordering 😉

rossf7: Fixed.

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -221,35 +214,35 @@ func (r Resource) updateChartOperator(ctx context.Context, cr v1alpha1.App) erro
...
-
-		err = backoff.RetryNotify(o, b, n)
 		if err != nil {
 			return microerror.Mask(err)
 		}
 
-		r.logger.LogCtx(ctx, "level", "debug", "message", "chart-operator deployment is ready")
+		r.logger.LogCtx(ctx, "level", "error", "message", "updated chart-operator release")
```
xh3b4sd: Why is this an error log?

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -135,41 +134,35 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
...
-			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("%#q deployment is not ready retrying in %s", release, delay), "stack", fmt.Sprintf("%#v", err))
-		}
-
-		err = backoff.RetryNotify(o, b, n)
 		if err != nil {
 			return microerror.Mask(err)
 		}
 
-		r.logger.LogCtx(ctx, "level", "debug", "message", "chart-operator deployment is ready")
+		r.logger.LogCtx(ctx, "level", "error", "message", "created chart-operator release")
```
xh3b4sd: Why is this an error log?

service/controller/app/v1/resource/chartoperator/resource.go
```diff
@@ -135,41 +134,35 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
...
-		// chart CRD is installed. This allows the chart
-		// resource to create CRs in the same reconcilation loop.
-		r.logger.LogCtx(ctx, "level", "debug", "message", "waiting for ready chart-operator deployment")
-
-		o := func() error {
-			err := r.checkDeploymentReady(ctx, cc.Clients.K8s.K8sClient())
-			if err != nil {
-				return microerror.Mask(err)
-			}
+		r.logger.LogCtx(ctx, "level", "error", "message", "creating chart-operator release")
```
xh3b4sd: Why is this an error log?

rossf7: Not sure. I'll fix.

service/controller/app/v1/resource/chartoperator/resource.go
rossf7: Same. Will fix.

xh3b4sd:
We should rather delay cluster creation instead of introducing a really odd design decision. When we do this now we will have to deal with it the next 2 years and who knows which implications the CRD versioning weirdness has. We just started out with it. Nobody can tell. And it is not so much about the CRD apiversion itself but rather about the OpenAPISchema validation versions, which defines the validation for the different CRs. So when anything in the CP defines the version scheme, which can only be a single one, then ALL Tenant Clusters have to comply with this, meaning we can't even upgrade schemes in TCs. So this thing here doesn't fly at all IMO.

rossf7:
> We should rather delay cluster creation instead of introducing a really odd design decision.

chart-operator is the only operator we install in tenants so I don't agree the design is odd.

> who knows which implications the CRD versioning weirdness has.

This I do agree with. There is too much uncertainly about the CRD versioning. 

Going to close this. I'll do a fresh PR with the changes to not wait for chart-operator.

xh3b4sd:
> chart-operator is the only operator we install in tenants so I don't agree the design is odd.

I meant the design of operator-CRD-lifecycle relationship.

PR #220, opened 2020-03-23
helm3: Remove e2e-harness

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/9519

This is prep for adding Helm 3 in #223 .

e2e-harness uses helmclient so this gets rid of the dependency. It also improves the integration testing situation since we want to retire e2e-harness.


Gopkg.toml
```diff
@@ -56,18 +56,10 @@ required = [
   name = "github.com/giantswarm/backoff"
   version = "~0.1.0"
 
-[[constraint]]
-  name = "github.com/giantswarm/e2e-harness"
-  version = "~0.1.0"
```
rossf7: 🔥

integration/key/key.go
```diff
@@ -0,0 +1,36 @@
+// +build k8srequired
+
+package key
```
rossf7: Consolidating the key funcs into a single file.

integration/release/error.go
```diff
@@ -0,0 +1,41 @@
+package release
```
rossf7: release package is extracted from e2e-harness with just what we need.

integration/setup/setup.go
```diff
@@ -61,10 +62,43 @@ func installResources(ctx context.Context, config Config) error {
 	}
 
 	{
-		err = config.Release.InstallOperator(ctx, key.AppOperatorReleaseName(), release.NewVersion(env.CircleSHA()), templates.AppOperatorValues, v1alpha1.NewAppCRD())
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("installing app-operator"))
```
rossf7: We install the operator here instead of using the release package.

integration/test/app/basic/simple_test.go
```diff
@@ -7,16 +7,15 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
```
rossf7: Sorry diff is a mess. Probably easiest to just check the new file.

integration/test/app/kubeconfig/kubeconfig_test.go
```diff
@@ -9,92 +9,54 @@ import (
...
 	"github.com/giantswarm/e2esetup/chart/env"
-	"github.com/giantswarm/e2etemplates/pkg/chartvalues"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
 
 	"github.com/giantswarm/app-operator/integration/key"
 	"github.com/giantswarm/app-operator/integration/templates"
+	"github.com/giantswarm/app-operator/pkg/label"
```
rossf7: Same here for the kubeconfig test.

integration/release/release.go
```diff
@@ -63,7 +63,7 @@ func (r *Release) WaitForAppCatalogCRD(ctx context.Context) error {
 		r.logger.Log("level", "debug", "message", fmt.Sprintf("failed to list appcatalogs: retrying in %s", t), "stack", fmt.Sprintf("%v", err))
 	}
 
-	b := backoff.NewExponential(2*time.Minute, 60*time.Second)
+	b := backoff.NewExponential(10*time.Minute, 60*time.Second)
```
rossf7: These backoffs were too optimistic. 🤦‍♂ 

chart-operator needs to create the chart CRD. So the kubeconfig test tasks 2 loops.

integration/test/app/kubeconfig/kubeconfig_test.go
```diff
@@ -23,73 +21,25 @@ import (
 )
 
 const (
-	clusterName = "kind-kind"
-	namespace   = "giantswarm"
+	catalogConfigMapName = "default-catalog-configmap"
+	chartOperatorName    = "chart-operator"
+	clusterName          = "kind-kind"
+	namespace            = "giantswarm"
```
rossf7: Same diff is a mess. Probably best to just check the new file.

integration/setup/setup.go
```diff
@@ -61,10 +62,43 @@ func installResources(ctx context.Context, config Config) error {
...
+
+		operatorVersion := fmt.Sprintf("1.0.0-%s", env.CircleSHA())
```
fiunchinho: Are you using 1.0.0 specifically on purpose?

integration/test/app/basic/simple_test.go
```diff
@@ -118,43 +72,76 @@ func TestAppLifecycle(t *testing.T) {
...
-
-		err = config.Release.WaitForStatus(ctx, fmt.Sprintf("%s-%s", namespace, key.CustomResourceReleaseName()), "DEPLOYED")
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating %#q app cr", key.TestAppReleaseName()))
+
+		appCR := &v1alpha1.App{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      key.TestAppReleaseName(),
+				Namespace: namespace,
+				Labels: map[string]string{
+					label.AppOperatorVersion: "1.0.0",
```
fiunchinho: I'd extract `1.0.0` to a constant that we can refer from different places.

integration/setup/setup.go
rossf7: Yes this is because its deployed by draughtsman. I added a TODO comment and this will be cleaned up when the operator is flattened.

integration/test/app/basic/simple_test.go
rossf7: Yes good point. I added it to the key package because its used by both tests.

PR #223, opened 2020-03-24
helm3: Upgrade helmclient to Helm 3

rossf7:
Towards https://github.com/giantswarm/roadmap/issues/30

This updates the helm3 branch to use the new helm 3 version of helmclient.


integration/test/app/basic/simple_test.go
```diff
@@ -50,6 +49,11 @@ func TestAppLifecycle(t *testing.T) {
 			if err != nil {
 				t.Fatalf("expected %#v got %#v", nil, err)
 			}
+
+			// TODO: Removing hardcoding once there is a chart-operator release
+			// with Helm 3 support in the default catalog.
+			tarballURL = "https://giantswarm.github.io/default-test-catalog/chart-operator-0.12.1-13521d4e2cb5378dbff26995e094d1c23a15e121.tgz"
+
```
rossf7, line 56: Like the TODO says this hardcoding is needed until there is a release of chart-operator with helm 3 support in the catalog.

go.mod
```diff
@@ -7,10 +7,10 @@ require (
 	github.com/giantswarm/appcatalog v0.1.11
 	github.com/giantswarm/apprclient v0.2.0
 	github.com/giantswarm/backoff v0.2.0
-	github.com/giantswarm/e2esetup v0.2.0
+	github.com/giantswarm/e2esetup v0.2.1-0.20200401192455-acfdbff0dabe
 	github.com/giantswarm/errors v0.2.0
 	github.com/giantswarm/exporterkit v0.2.0
-	github.com/giantswarm/helmclient v0.2.0
+	github.com/giantswarm/helmclient v0.0.0-20200331110847-d697c5c4a0fb
```
xh3b4sd, line 13: Can you try updating the deps again with our tags? The time/hash prefix should not be necessary with out libraries.

rossf7: This is because it's using the helm3 branch of helmclient. Before this gets merged to master we'll merge helmclient into master and create a tag.

tomahawk28: This is because we are using `helm3` branch in helmclient for this branch.

rossf7:
> I think we should not install chart-operator when it's there in helm v2 release. But it can be done by different PRs.

I wanted to test with the helm 3 version of chart-operator. But yes we'll need to remove the hardcoded chart-operator versions before merging to master.

rossf7:
> we should not install chart-operator when it's there in helm v2 release.

Ah thanks yes if chart-operator is already installed we should update the chart-operator app CR. Not install chart-operator.

I'll do a follow up PR to check for this.

PR #224, opened 2020-03-26
migrate from dep to Go modules

tomahawk28:

PR #226, opened 2020-03-27
Revert "Revert "migrate from dep to Go modules (#224)" (#225)"

tomahawk28:
This reverts commit 5f528eba293f31764d265a1060827711406551dc.

Re-apply go modules.


tomahawk28:
going with a ping.

PR #227, opened 2020-03-27
Implement delete patch on update for configmaps and secrets

rossf7:
```
 D 03/27 15:57:21 found 1 orphan secrets giantswarm.releases-kvm-unique-chart-secrets | chart-operator/service/collector/orphan_secret.go:94
```

Fix for orphan chart configmaps and secrets. If there was no resource then we cleared the app CR spec but didn't delete the configmap and secret.

In this case its because the release apps moved from the control-plane to releases catalog. But there are other edge cases that this will fix.


rossf7:
Tested on `gorgoth`.

PR #242, opened 2020-04-06
Moving Releasemigration into app-operator

tomahawk28:
## Checklist

Copy `releasemigration` from chart-operator.


main.go
```diff
@@ -103,6 +103,7 @@ func mainWithError() (err error) {
 	daemonCommand := newCommand.DaemonCommand().CobraCommand()
 
 	daemonCommand.PersistentFlags().String(f.Service.Chart.Namespace, "giantswarm", "The namespace where chart CRs are located.")
+	daemonCommand.PersistentFlags().String(f.Service.Helm.TillerNamespace, "giantswarm", "Namespace for the Tiller pod.")
```
tomahawk28, line 106: It wasn't introduced until now.

rossf7: Cool makes sense. So far only tiller in tenant clusters was managed and its always in the giantswarm namespace.

service/controller/app/v1/resource/releasemigration/error.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+var releasesNotDeletedError = &microerror.Error{
+	Kind: "releasesNotDeletedError",
+}
+
+// releasesNotDeletedErrorMatching asserts releasesNotDeletedError
+func releasesNotDeletedErrorMatching(err error) bool {
+	return microerror.Cause(err) == releasesNotDeletedError
+}
```
rossf7, line 14: Do we actually match on this error? If not can we use `executionFailed`?

PR #255, opened 2020-04-14
Bootstrap chart-operator in control plane clusters

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7896

This is the final PR for bootstrapping chart-operator in tenant clusters.

It can be merged before https://github.com/giantswarm/chart-operator/pull/421 because it won't take affect until chart-operator is added to the app collections.

### DIFF
--- a/helm/app-operator-chart/templates/rbac.yaml
+++ b/helm/app-operator-chart/templates/rbac.yaml
@@ -72,11 +72,12 @@ rules:
   verbs:
   - "create"
 - apiGroups:
-  - extensions
+  - apps
   resources:
   - deployments
   verbs:
-  - 'create'
+  - get
+  - create
 - apiGroups:
   - ""
   resources:
--- a/helm/app-operator-chart/templates/rbac.yaml
+++ b/helm/app-operator-chart/templates/rbac.yaml
@@ -76,8 +76,8 @@ rules:
   resources:
   - deployments
   verbs:
-  - get
   - create
+  - get
 - apiGroups:
   - ""
   resources:
--- a/service/controller/app/v1/resource/chartoperator/create.go
+++ b/service/controller/app/v1/resource/chartoperator/create.go
@@ -39,7 +39,7 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 	// Check whether tenant cluster has a chart-operator helm release yet.
 	{
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding chart-operator release %#q in tenant cluster", release))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding chart-operator release %#q", release))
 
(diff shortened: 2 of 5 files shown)


helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -72,11 +72,12 @@ rules:
   verbs:
   - "create"
 - apiGroups:
-  - extensions
+  - apps
```
rossf7, line 75: Fixing the API group.

helm/app-operator-chart/templates/rbac.yaml
```diff
@@ -72,11 +72,12 @@ rules:
...
   resources:
   - deployments
   verbs:
-  - 'create'
+  - create
+  - get
```
rossf7, line 80: For checking if the deployment is ready.

service/controller/app/v1/resource/chartoperator/create.go
```diff
@@ -23,29 +23,23 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
-	// Resource is used to bootstrap chart-operator in tenant clusters.
-	// So for other apps we can skip this step.
+	// Resource is used to bootstrap chart-operator. So for other apps we can
+	// skip this step.
```
rossf7, line 27: Just removing tenant cluster from comments. There are a few of these.

service/controller/app/v1/resource/clients/resource.go
```diff
@@ -54,18 +54,14 @@ func New(config Config) (*Resource, error) {
 	if config.ImageRegistry == "" {
 		return nil, microerror.Maskf(invalidConfigError, "%T.ImageRegistry must not be empty", config)
 	}
-	if config.TillerNamespace == "" {
-		config.TillerNamespace = "giantswarm"
-	}
```
rossf7, line 59: We now use the in cluster flag to determine which tiller namespace should be used.

service/controller/app/v1/resource/clients/resource.go
```diff
@@ -130,6 +126,16 @@ func (r *Resource) addClientsToContext(ctx context.Context, cr v1alpha1.App) err
...
 	}
 
+	var tillerNamespace string
+	{
+		// When InCluster is used we use the Tiller deployed in kube-system.
+		if key.InCluster(cr) {
+			tillerNamespace = metav1.NamespaceSystem
+		} else {
+			tillerNamespace = "giantswarm"
+		}
```
rossf7, line 136: This is a pain but the in cluster flag is the best way to determine this. Helm 3 helps a bit but we still need to know the namespace for the migration logic.

rossf7:
Tested on `geckon`.

rossf7:
Going with this.

It won't take effect until we create app CRs for chart-operator in the control planes. https://github.com/giantswarm/chart-operator/pull/421

PR #258, opened 2020-04-16
Process current CR for chart-operator bootstrapping

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7896

This is cleanup that got missed from a previous change. We only execute the chartoperator resource for its CRs. So we can process the current CR instead of fetching it again.

https://github.com/giantswarm/app-operator/blob/5fe0a86c388282d6b35f1d05c938f65bfeb1d6b6/service/controller/app/v1/resource/chartoperator/create.go#L26-L32

### DIFF
--- a/integration/templates/chart_operator_values.go
+++ b/integration/templates/chart_operator_values.go
@@ -5,7 +5,7 @@ package templates
 // ChartOperatorValues values required by chart-operator-chart.
 const ChartOperatorValues = `
 clusterDNSIP: 10.96.0.10
-e2e: true
+e2e: false
 externalDNSIP: 8.8.8.8
 image:
   registry: "quay.io"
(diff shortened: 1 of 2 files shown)


integration/templates/chart_operator_values.go
```diff
@@ -5,7 +5,7 @@ package templates
 // ChartOperatorValues values required by chart-operator-chart.
 const ChartOperatorValues = `
 clusterDNSIP: 10.96.0.10
-e2e: true
+e2e: false
```
rossf7, line 8: This is for the giantswarm-critical priority class. Since this is still on Helm 2 its created by EnsureTillerInstalled.

rossf7:
Tested on `geckon`.

PR #259, opened 2020-04-16
Remove hardcoded release name from chartoperator resource

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7896

This removed the hardcoded release name and namespace. This is needed because in control planes the chart-operator CR will be called `chart-operator-unique`.

PR #262, opened 2020-04-17
helm3: Sync chartoperator resource with master

rossf7:
Syncs helm3 branch with changes in https://github.com/giantswarm/app-operator/pull/258 and https://github.com/giantswarm/app-operator/pull/259


rossf7:
Going with a ping. One more of these to do.

PR #265, opened 2020-04-20
Remove v1 resourceset

tomahawk28:
Towards giantswarm/giantswarm#7896

Flattening both app, appcatalog resourceset.


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA      = "n/a"
 	name        = "app-operator"
 	source      = "https://github.com/giantswarm/app-operator"
-	version     = "n/a"
+	version     = "1.0.0"
```
tomahawk28, line 8: adding `1.0.0` and compare it with version labels in CR.

service/controller/app/resource/appnamespace/resource.go
```diff
@@ -9,12 +9,12 @@ import (
...
 	"k8s.io/client-go/kubernetes"
 
-	"github.com/giantswarm/app-operator/service/controller/app/v1/controllercontext"
+	"github.com/giantswarm/app-operator/service/controller/app/controllercontext"
 )
 
 const (
 	// Name is the identifier of the resource.
-	Name = "appnamespacev1"
+	Name = "appnamespace"
```
tomahawk28, line 17: Fixing all resource name

pkg/project/project.go
rossf7: This will only work for the first release. As soon as we need to create `1.0.1` it doesn't match any more.

I added a ChartVersion func to chart-operator. I think we need AppVersion here.

https://github.com/giantswarm/chart-operator/blob/ae3cb2f4c26821f66aa47199bdbc64855580d402/pkg/project/project.go#L11-L16

tomahawk28: Got it, leave it as it 1.0.0 but would introduce AppVersion in this file.

kopiczko, line 8: I'd suggest to go with `1.0.0-dev` before the first flat release.

kopiczko: App version and chart version are equal for production releases.

kopiczko: And one more thing. The orb will fail if value of `version` in `project.go` and the value in `appVersion` in `Chart.yaml` are not equal.

kopiczko: Example https://circleci.com/gh/giantswarm/release-operator/2311

pkg/project/project.go
```diff
@@ -5,9 +5,15 @@ var (
 	gitSHA      = "n/a"
 	name        = "app-operator"
 	source      = "https://github.com/giantswarm/app-operator"
-	version     = "n/a"
+	version     = "1.0.0-dev"
 )
 
+// AppVersion is fixed for app CRs. Its version is not linked to a release.
+// release. We may revisit this in future.
```
rossf7, line 12:
```suggestion
// We may revisit this in future.
```

tomahawk28:
Tested in ginger

PR #268, opened 2020-04-22
Add to app collection

tomahawk28:
Toward giantswarm/giantswarm#7895

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  architect: giantswarm/architect@v0.8.12
+  architect: giantswarm/architect@0.8.12
 
 jobs:
   build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  architect: giantswarm/architect@0.8.11
+  architect: giantswarm/architect@v0.8.12
 
 jobs:
   build:
@@ -86,6 +86,7 @@ workflows:
           name: push-app-operator-to-aws-app-collection
           app_name: "app-operator"
           app_collection_repo: "aws-app-collection"
+          disable_force_upgrade: "true"
           unique: "true"
           requires:
             - push-app-operator-to-control-plane-app-catalog
@@ -100,6 +101,7 @@ workflows:
           name: push-app-operator-to-azure-app-collection
           app_name: "app-operator"
           app_collection_repo: "azure-app-collection"
+          disable_force_upgrade: "true"
           unique: "true"
           requires:
             - push-app-operator-to-control-plane-app-catalog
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -80,3 +80,48 @@ workflows:
           filters:
             tags:
               only: /^v.*/
+
+      # deploy to aws installations (only tags)
+      - architect/push-to-app-collection:
```
kopiczko, line 85:
```suggestion
      - architect/push-to-app-collection:
          context: "architect"
```

.circleci/config.yml
```diff
@@ -80,3 +80,48 @@ workflows:
...
+          requires:
+            - push-app-operator-to-control-plane-app-catalog
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v.*/
+
+      # deploy to azure installations (only tags)
+      - architect/push-to-app-collection:
```
kopiczko, line 100:
```suggestion
      - architect/push-to-app-collection:
          context: "architect"
```

.circleci/config.yml
```diff
@@ -80,3 +80,48 @@ workflows:
...
+          requires:
+            - push-app-operator-to-control-plane-app-catalog
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v.*/
+
+      # deploy to kvm installations (only tags)
+      - architect/push-to-app-collection:
```
kopiczko, line 115:
```suggestion
      - architect/push-to-app-collection:
          context: "architect"
```

tomahawk28:
@rossf7 I created below script, WDYT? There are no configmap & secret but it would work. 
```
#!/bin/bash
l=$(opsctl list installations | sed '1d' | awk '{ print $1 }')
for t in $l; do
	echo ${t}
cat <<EOF | kubectl create -f -
apiVersion: application.giantswarm.io/v1alpha1
kind: Chart
metadata:
  annotations:
    chart-operator.giantswarm.io/force-helm-upgrade: "true"
  finalizers:
  - operatorkit.giantswarm.io/chart-operator-chart
  generation: 1
  labels:
    chart-operator.giantswarm.io/version: 1.0.0
  name: app-operator-unique
  namespace: giantswarm
spec:
  name: app-operator-unique
  namespace: giantswarm
  tarballURL: https://giantswarm.github.io/control-plane-catalog/app-operator-1.0.0.tgz
  version: 1.0.0
EOF
done
```

rossf7:
@tomahawk28 Yes the script looks fine. I don't think the YAML needs the generation field but it does no harm.

I have a concern about China as it will pull the image from Quay. But once app-operator processes its own CR it will use Aliyun again.

tomahawk28:
I think it would pull from Aliyun if CP have a right value for it. https://github.com/giantswarm/app-operator/blob/1850d95ad79588c3882f46ddcde11c1dd68386b4/helm/app-operator/templates/deployment.yaml#L36

PR #272, opened 2020-04-23
fix workflow

tomahawk28:
Fixing workflow, tagging repo is not triggering catalog creation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,10 @@ workflows:
           password_envar: "QUAY_PASSWORD"
           requires:
             - build
+          filters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -54,18 +58,12 @@ workflows:
...
-          test-dir: "integration/test/app/basic"
-          requires:
-            - build
-
-      - architect/integration-test:
-          name: kubeconfig-integration-test
-          test-dir: "integration/test/app/kubeconfig"
-          requires:
-            - build
+          filters:
```
tomahawk28, line 61: part of push-to-aliyun

PR #276, opened 2020-04-28
Sync helm3 branch to the latest master branch

tomahawk28:
We created helm3 branch from almost one and a half month ago, now we occasionally sync the changes from master into this branch, I hope this is the last step before we merge helm3 branch. 

If something we need to fix in this PR, mostly it coming from master branch.


pkg/project/project.go
```diff
@@ -5,9 +5,15 @@ var (
 	gitSHA      = "n/a"
 	name        = "app-operator"
 	source      = "https://github.com/giantswarm/app-operator"
-	version     = "n/a"
+	version     = "1.0.1"
```
tomahawk28, line 8: Please note this is a different kind of version from the below `AppVersion`

rossf7: This is the version of the operator. It's different because all app CRs currently have a version of 1.0.0. This is explained in the comment below.

I think we need to revisit this after Helm 3. So we run multiple app-operator instances linked to a release.

PR #283, opened 2020-05-05
avoid affecting CPs

tomahawk28:
We should avoid affecting CPs when we are testing TCs.


service/controller/app/resource/releasemigration/create.go
```diff
@@ -30,6 +30,13 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	//TODO: Remove this statement when we need to test on CP
```
rossf7, line 33:
```suggestion
	//TODO: Remove this statement when we need to test on control planes.
```

PR #285, opened 2020-05-06
migrating only in case of chart-operator

tomahawk28:
`releasemigration` works in case of below; 
1. When its chart-operator app CR reconciliation
2. if the `appVersion` is greater or equal than 1
3. When the latest chart-operator is up and running in TC.

### DIFF
--- a/service/controller/app/resource/releasemigration/create.go
+++ b/service/controller/app/resource/releasemigration/create.go
@@ -62,12 +62,12 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return nil
 	}
 
-	content, err := cc.Clients.Helm.GetReleaseContent(ctx, key.Namespace(cr), key.ReleaseName(cr))
+	chart, err := cc.Clients.K8s.G8sClient().ApplicationV1alpha1().Charts(r.chartNamespace).Get(cr.GetName(), metav1.GetOptions{})
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
-	if content.Version != key.Version(cr) || content.Status != "DEPLOYED" {
+	if chart.Status.Version != key.Version(cr) || chart.Status.Release.Status != "DEPLOYED" {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("%#q is not deployed yet", key.AppName(cr)))
 		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
 		return nil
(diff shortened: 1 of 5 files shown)


service/controller/app/resource/releasemigration/create.go
```diff
@@ -43,6 +44,35 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return nil
 	}
 
+	// Resource is used to migrating all charts. it's too noisy if it works on every reconciliation.
```
rossf7, line 47: This is the same case as the chartoperator resource. We only need it to execute for the chart-operator app CR.

Could you try and make the comment and logging similar? 

https://github.com/giantswarm/app-operator/blob/5d02ba5453a6dd119f0e39f8c8cf5698a286d501/service/controller/app/resource/chartoperator/create.go#L28-L34

service/controller/app/resource/releasemigration/create.go
```diff
@@ -43,6 +44,35 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+		return nil
+	}
+
+	v, err := semver.NewVersion(key.Version(cr))
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if v.Major() < 1 {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("chart %#q's current version %#q is not greater than 1.0.0.", key.AppName(cr), key.Version(cr)))
```
rossf7, line 60: I think the log message needs to explain we don't trigger the migration because the tenant cluster uses Helm 2.

For others outside our team they may not be aware 1.X is Helm 3 and 0.X is Helm 2.

service/controller/app/resource/releasemigration/create.go
```diff
@@ -52,7 +52,13 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return nil
 	}
 
-	v, err := semver.NewVersion(key.Version(cr))
+	if cr.Status.AppVersion == "" {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("chart %#q is not installed yet", key.AppName(cr)))
```
rossf7, line 56:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q is not installed yet", key.AppName(cr)))
```

tomahawk28:
```D 05/06 17:30:59 /apis/application.giantswarm.io/v1alpha1/namespaces/0imrb/apps/chart-operator releasemigrationv1 chart `chart-operator`, version `0.13.0` is using Helm 2. we don't need to trigger Helm 3 migration. | app-operator/service/controller/app/resource/releasemigration/create.go:67 | controller=app-operator | event=update | loop=1 | version=149033782```

PR #289, opened 2020-05-07
Add team label to app alerts

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/9705

- Adds a team label to the app info metric for routing alerts.
- Also drops the `service_test.go` unit test. This is hard to maintain and is mainly to prevent crashing but we have integration test coverage for that.

For now the mapping of apps to teams is in the configmap. Once the Representation of Catalog Entries story https://github.com/giantswarm/roadmap/issues/26 is implemented it will come from the appcatalogentry CR. 

```
curl http://localhost:8000/metrics | grep app_operator_app_info | grep azure-operator

app_operator_app_info{name="azure-operator-3.0.4-2",namespace="giantswarm",status="DEPLOYED",team="celestial",version="3.0.4-2"} 1
app_operator_app_info{name="azure-operator-3.0.5-2",namespace="giantswarm",status="DEPLOYED",team="celestial",version="3.0.5-2"} 1
app_operator_app_info{name="azure-operator-3.0.6-1",namespace="giantswarm",status="DEPLOYED",team="celestial",version="3.0.6-1"} 1
app_operator_app_info{name="azure-operator-3.0.7-1",namespace="giantswarm",status="DEPLOYED",team="celestial",version="3.0.7-1"} 1
app_operator_app_info{name="azure-operator-4.0.0",namespace="giantswarm",status="DEPLOYED",team="celestial",version="4.0.0"} 1
app_operator_app_info{name="azure-operator-nginx
```

## Checklist


helm/app-operator/templates/configmap.yaml
```diff
@@ -14,6 +14,14 @@ data:
       listen:
         address: 'http://0.0.0.0:{{ .Values.port }}'
     service:
+      collector:
+        apps:
+          defaultTeam: "batman"
```
rossf7, line 19: Batman is the default team if there is no mapping.

rossf7:
Tested on `ghost`.

service/service.go
```diff
@@ -124,11 +126,27 @@ func New(config Config) (*Service, error) {
 		}
 	}
 
+	// TODO: Remove once appcatalogentry CRs exist and have the name of the
+	// team responsible for the app.
+	//
+	//	https://github.com/giantswarm/roadmap/issues/26
```
tomahawk28, line 132: The issue is closed already. Maybe you are referring to others?

rossf7: Yes I know. @piontec when can you create the implementation issue?

I also don't really get why we can't use the same issue for spec and implementation.

piontec: Wait, what do you mean by implementation issue :) ?

piontec:
Here's my realted ticket that would need team labels to route alerts about violating basic SLOs

service/service.go
rossf7: Well the spec issue https://github.com/giantswarm/roadmap/issues/26 is closed.

But it isn't implemented yet. Where do we track that?

PR #297, opened 2020-05-09
check status version with spec version before move on

tomahawk28:
Checking `.spec.version` with `.status.version` before we move on to release migration.

PR #298, opened 2020-05-10
helm3: Upgrading control planes to Helm 3

rossf7:
Fixes to support the Helm 3 migration in control planes as well as tenant clusters.


helm/app-operator/templates/rbac.yaml
```diff
@@ -6,17 +6,17 @@ metadata:
     {{- include "labels.common" . | nindent 4 }}
 rules:
 - apiGroups:
-  - apiextensions.k8s.io
+    - apiextensions.k8s.io
```
rossf7, line 9: Needs #299 first to clean up the diff.

service/controller/app/resource/chartoperator/create.go
```diff
@@ -42,7 +42,7 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	{
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding %#q deployment", cr.Name))
 
-		_, err = cc.Clients.K8s.K8sClient().AppsV1().Deployments(key.Namespace(cr)).Get(key.AppName(cr), metav1.GetOptions{})
+		_, err = cc.Clients.K8s.K8sClient().AppsV1().Deployments(key.Namespace(cr)).Get(cr.Name, metav1.GetOptions{})
```
rossf7, line 45: In the control plane the deployment is chart-operator-unique not chart-operator. So we need to use the CR name.

service/controller/app/resource/releasemigration/create.go
```diff
@@ -53,6 +53,12 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return nil
 	}
 
+	if key.Version(cr) != cr.Status.Version {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q is not reconciled to the latest desired status yet", key.AppName(cr)))
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+		return nil
+	}
```
rossf7, line 60: From #297

service/controller/app/resource/releasemigration/create.go
```diff
@@ -117,7 +123,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		}
 
 		// install helm-2to3-migration app
-		err = r.ensureReleasesMigrated(ctx, cc.Clients.K8s.K8sClient(), cc.Clients.Helm, tillerNamespace)
+		err = r.ensureReleasesMigrated(ctx, cc.Clients.K8s, cc.Clients.Helm, tillerNamespace)
```
rossf7, line 126: We now need both k8s and g8s clients. So its better to pass the collection.

service/controller/app/resource/releasemigration/resource.go
```diff
@@ -61,20 +61,38 @@ func (r *Resource) Name() string {
...
 
-func (r *Resource) findHelmV2Releases(k8sClient kubernetes.Interface, tillerNamespace string) ([]string, error) {
+func (r *Resource) findHelmV2Releases(k8sClient k8sclient.Interface, tillerNamespace string) ([]string, error) {
+	charts := make(map[string]bool)
+
+	// Get list of chart CRs as not all helm 2 releases will have a chart CR.
+	list, err := k8sClient.G8sClient().ApplicationV1alpha1().Charts("giantswarm").List(metav1.ListOptions{})
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
```
rossf7, line 71: In the control plane there are also Helm 2 releases managed by `draughtsman` and `opsctl ensure appcatalogs`.

So we should only migrate helm releases that have a chart CR.

rossf7:
Tested on `ghost`.

helm/app-operator/templates/rbac.yaml
```diff
@@ -51,30 +51,58 @@ rules:
     - create
     - patch
     - update
+- apiGroups:
```
rossf7, line 54: Additional rules are for creating the helm-2to3-migration job and related resources.

https://github.com/giantswarm/helm-2to3-migration/tree/bdac5f952e479ffc582d408818101a6309fda44a/helm/helm-2to3-migration/templates

tomahawk28: is it necessary? We created it by helm anyway, and myself never encountered with an RBAC error related to this chart.

rossf7: If you install the chart with tiller it has cluster-admin. Using helm 3 version of helmclient there is no tiller so app-operator needs the permissions.

service/controller/app/resource/releasemigration/resource.go
```diff
@@ -61,20 +61,38 @@ func (r *Resource) Name() string {
 	return Name
 }
 
-func (r *Resource) findHelmV2Releases(k8sClient kubernetes.Interface, tillerNamespace string) ([]string, error) {
+func (r *Resource) findHelmV2Releases(k8sClient k8sclient.Interface, tillerNamespace string) ([]string, error) {
+	charts := make(map[string]bool)
+
+	// Get list of chart CRs as not all helm 2 releases will have a chart CR.
+	list, err := k8sClient.G8sClient().ApplicationV1alpha1().Charts("giantswarm").List(metav1.ListOptions{})
```
tomahawk28, line 68:
```suggestion
	list, err := k8sClient.G8sClient().ApplicationV1alpha1().Charts(r.chartNamespace).List(metav1.ListOptions{})
```
What do you think utilise r.chartNamespace?

rossf7: Good point, thx. It was hardcoded in a few placed I've updated.

PR #299, opened 2020-05-10
helm3: Tidy up RBAC rules

rossf7:
Prep for #298 

Since there is no tiller we need extra permissions to create the helm-2to3 migration job.

But the current cluster role is messy. This tidies it up so the diff for the new permissions is cleaner.


helm/app-operator/templates/rbac.yaml
```diff
@@ -6,17 +6,17 @@ metadata:
     {{- include "labels.common" . | nindent 4 }}
 rules:
 - apiGroups:
-  - apiextensions.k8s.io
+    - apiextensions.k8s.io
```
rossf7, line 9: Fixing indentation.

helm/app-operator/templates/rbac.yaml
```diff
@@ -52,37 +52,37 @@ rules:
     - patch
     - update
 - apiGroups:
-  - "rbac.authorization.k8s.io"
+    - rbac.authorization.k8s.io
```
rossf7, line 55: Removing unnecessary quotes.

helm/app-operator/templates/rbac.yaml
```diff
@@ -52,37 +52,37 @@ rules:
...
-  - services
-  - serviceaccounts
+    - services
+    - serviceaccounts
   verbs:
-  - "create"
+    - create
 - apiGroups:
-  - extensions
+    - apps
```
rossf7, line 74: This should be apps now we're on K8s 1.16.

rossf7:
I don't think this is controversial. Going with a ping.

PR #303, opened 2020-05-11
helm3: Update go modules

rossf7:


rossf7:
Same as https://github.com/giantswarm/app-operator/pull/302 for helm3 branch. Going with a ping.

PR #308, opened 2020-05-12
Deleting unnecessary resources for thiccc 

tomahawk28:
toward https://github.com/giantswarm/giantswarm/issues/10798
Deleting unnecessary resources for `thiccc` branch. 
circleCI setting changes would follow.


service/service.go
```diff
@@ -95,19 +93,6 @@ func New(config Config) (*Service, error) {
 		}
 	}
 
-	var appCatalogController *appcatalog.AppCatalog
```
tomahawk28, line 98: Another app-operator in unique would handle appCatalog CRD as well, so we don't need it on thiccc branch.

.circleci/config.yml
```diff
@@ -38,78 +38,3 @@ workflows:
             tags:
               only: /^v.*/
 
-      # build and push chart to app catalog
```
tomahawk28, line 41: Deleting integration testing & app catalog at the moment.

service/controller/app/key/key.go
```diff
@@ -137,6 +138,18 @@ func KubecConfigSecretNamespace(customResource v1alpha1.App) string {
 	return customResource.Spec.KubeConfig.Secret.Namespace
 }
 
+func LabelSelectorService() labels.Selector {
+	// Selector to check if "app=chart-operator" and if helm-major-version is not belonging to 3.
+	s := fmt.Sprintf("%s=%s,%s notin (%d)", label.App, "chart-operator", label.HelmMajorVersion, 3)
```
tomahawk28, line 143: It will only get chart-operator app CR with no match for helm 3 version labels.

rossf7: This would be easier to review if all the deletions were in one PR and have a separate for the label selector.

service/controller/app/key/key.go
```diff
@@ -137,6 +138,18 @@ func KubecConfigSecretNamespace(customResource v1alpha1.App) string {
 	return customResource.Spec.KubeConfig.Secret.Namespace
 }
 
+func LabelSelectorService() labels.Selector {
```
rossf7, line 141:
```suggestion
func AppLabelSelector() labels.Selector {
```

Do we need service in the name? What about this since its selecting apps?

service/controller/app/key/key.go
tomahawk28: Didn't notice that was including the part of all together, sorry!

service/controller/app/key/key.go
tomahawk28: You're right, it's clear w/o service in the name.

PR #309, opened 2020-05-12
circleCI settings for thiccc branch

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/10798

New circleCI setting for thiccc branch only.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
       - checkout
       - run: |
-          wget -q $(curl -sS -H "Authorization: token $RELEASE_TOKEN" https://api.github.com/repos/giantswarm/architect/releases/latest | grep browser_download_url | head -n 1 | cut -d '"' -f 4)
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -18,23 +43,23 @@ workflows:
...
           password_envar: "QUAY_PASSWORD"
           requires:
-            - go-build
-          filters:
-            # Trigger the job also on git tag.
-            tags:
-              only: /^v.*/
+            - build
 
       - architect/push-to-docker:
```
rossf7, line 48: I think we need to use push-to-docker-legacy so the image tag is correct for draughtsman.

See https://github.com/giantswarm/aws-operator/blob/ca543815a0672225ca14396649ce51f9fe7502d2/.circleci/config.yml#L258

rossf7:
For draughtsman the tag is just the SHA without a version prefix. If you check the build step you can see it pushes the image to Quay.

https://circleci.com/gh/giantswarm/app-operator/4647

So we don't need the `push-app-operator-to-quay` job. But it doesn't push to Aliyun. So I think we need push-to-docker-legacy for that.

tomahawk28:
I think we need to revert this in installations, right? https://github.com/giantswarm/installations/pull/1181

rossf7:
> I think we need to revert this in installations, right?

Yes but there is also the project list in architect. I would check the flattening docs to make sure we don't miss anywhere.

PR #315, opened 2020-05-13
Helm3 major version label

rossf7:
We will add a new label `app-operator.giantswarm.io/helm-major-version="3"` to chart-operator app CRs in cluster-operator if chart-operator is version 1.X.

In the thiccc pod we use a label selector to exclude these CRs. The thiccc pod is only used to bootstrap chart-operator.

For Helm 3 we only execute the tcnamespace and chartoperator resources if this label is present.

### DIFF
--- a/integration/test/app/basic/simple_test.go
+++ b/integration/test/app/basic/simple_test.go
@@ -102,6 +102,8 @@ func TestAppLifecycle(t *testing.T) {
 				Name: key.DefaultCatalogName(),
 				Labels: map[string]string{
 					label.AppOperatorVersion: key.AppOperatorVersion(),
+					// Helm major version is 3 for the master branch.
+					label.HelmMajorVersion: "3",
 				},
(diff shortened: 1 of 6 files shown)


tomahawk28:
6 resources also on thiccc branch, should we block on that too?
- appnamespace
- appcatalog
- clients
- tcnamespace
- tiller
- chartoperator

rossf7:
The first 3 resources add information to the controller context which we need.

Good point I forgot about the tiller resource. 🤦‍♂️ 

I think the design still works. The configmap, secret and chart resources we removed as they are not part of the bootstrapping.

tomahawk28:
> Good point I forgot about the tiller resource.

I was thinking about it but we don't have tiller in helm3, so should be okay.

rossf7:
Tested on `ghost` with Helm 2 and CI green for Helm 3.

PR #325, opened 2020-05-18
Apply helm3 changes

rossf7:
Toward giantswarm/giantswarm#10973

## Checklist


go.mod
```diff
@@ -3,15 +3,14 @@ module github.com/giantswarm/app-operator
...
 require (
+	github.com/Masterminds/semver/v3 v3.0.3
 	github.com/giantswarm/apiextensions v0.2.2
 	github.com/giantswarm/appcatalog v0.2.3
 	github.com/giantswarm/backoff v0.2.0
-	github.com/giantswarm/e2e-harness v0.2.0
-	github.com/giantswarm/e2esetup v0.2.0
-	github.com/giantswarm/e2etemplates v0.2.0
-	github.com/giantswarm/errors v0.2.3
+	github.com/giantswarm/e2esetup v0.2.1-0.20200401192455-acfdbff0dabe
```
tomahawk28, line 10: Do we have a problem with e2esetup 0.2.0? I saw chart-operator is using 0.1.0 currently.

rossf7: Yes good catch. It's because it's still using Helm 2. I think we need to create a 0.3.0 release with helmclient 1.0.0.

tomahawk28: I will follow up after we finished CP testing.

PR #330, opened 2020-05-18
Set HTTP client timeout for helmclient and improve log message

rossf7:
Fixes problems with the migration job found in testing on giraffe.

Also needs fix for image templating in https://github.com/giantswarm/helm-2to3-migration/pull/14

### DIFF
--- a/helm/app-operator/templates/configmap.yaml
+++ b/helm/app-operator/templates/configmap.yaml
@@ -22,6 +22,9 @@ data:
             firecracker: "aws-operator,cluster-operator,kiam,releases-aws"
             ludacris: "bridge-operator,cert-exporter,cert-operator,coredns-app,flannel-operator,ignition-operator,kvm-operator,kube-state-metrics-app,metrics-server-app,net-exporter,node-operator,release-operator,releases-kvm"
             magic: "cluster-service,dex-app,g8s-cert-manager,g8s-efk,g8s-grafana,g8s-oauth2-proxy,g8s-prometheus,gatekeeper-app,gatekeeper-control-plane-rules,kubernetesd,rbac-operator,tokend,userd,vault-app"
+      helm:
+        http:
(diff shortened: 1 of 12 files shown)


helm/app-operator/templates/configmap.yaml
```diff
@@ -22,6 +22,9 @@ data:
             firecracker: "aws-operator,cluster-operator,kiam,releases-aws"
             ludacris: "bridge-operator,cert-exporter,cert-operator,coredns-app,flannel-operator,ignition-operator,kvm-operator,kube-state-metrics-app,metrics-server-app,net-exporter,node-operator,release-operator,releases-kvm"
             magic: "cluster-service,dex-app,g8s-cert-manager,g8s-efk,g8s-grafana,g8s-oauth2-proxy,g8s-prometheus,gatekeeper-app,gatekeeper-control-plane-rules,kubernetesd,rbac-operator,tokend,userd,vault-app"
+      helm:
+        http:
+          clientTimeout: '{{ .Values.Installation.V1.Helm.HTTP.ClientTimeout }}'
```
rossf7, line 27: For setting the timeout to 30 secs in China.

service/controller/app/resource/releasemigration/resource.go
```diff
@@ -181,7 +181,7 @@ func (r *Resource) ensureReleasesMigrated(ctx context.Context, k8sClient k8sclie
 	}
 
 	n := func(err error, t time.Duration) {
-		r.logger.Log("level", "debug", "message", "failed to migrate all helm v2 releases")
+		r.logger.Log("level", "debug", "message", "migration not complete")
```
rossf7, line 184: This retry is waiting for the migration job to complete. So I think this is clearer as nothing is failed. WDYT?

tomahawk28: I agreed, much better to understand.

PR #337, opened 2020-05-21
Update deps

rossf7:
Sorting updates found by dependabot. 

Most important is errors update as checking for unavailable tenant clusters is not working on godsmack.

## Checklist

-  [x] Update changelog in CHANGELOG.md.


rossf7:
Deps only changes. Going with a ping.

PR #339, opened 2020-05-21
Release v1.1.2

rossf7:
## Checklist


rossf7:
> I think we need to revert helm-2to3-migration.

It should work as image.registry is the same. But yes good to remove that. I'll do it now.

PR #343, opened 2020-05-25
Update apiextensions as 0.4.3

tomahawk28:
Toward giantswarm/giantswarm#10644

Update apiextension as 0.4.3 so we update app CRD.


tomahawk28:
Tested in ghost
```
kg get crd apps.application.giantswarm.io -o yaml
...
                description: Release is the status of the Helm release for the deployed
                  app.
                properties:
                  lastDeployed:
                    description: LastDeployed is the time when the app was last deployed.
                    format: date-time
                    type: string
                  reason:
                    description: Reason is the description of the last status of helm
                      release when the app is not installed successfully, e.g. deploy
                      resource already exists.
                    type: string
                  status:
                    description: Status is the status of the deployed app, e.g. DEPLOYED.
                    type: string
                required:
                - status
                type: object
```

PR #349, opened 2020-05-28
Checking deployment image & status rather than helm status

tomahawk28:
When chart-operator 1.0.x deployed into clusters, it could not report its status back to chart-operator chart CRs since it could only look into the helm 3 release. At that point, the migration from helm 2 is not even starting. So before we start the migration, we need to check the version chart-operator containers and its deployment status (ReadyReplica), rather than helm status.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## Unreleased
 
+### Changed
+
+- Checking chart-operator deployment status before initiate helm 3 migration.
+
 ## [v1.1.3] 2020-05-26
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Changed
 
-- Checking chart-operator deployment status before initiate helm 3 migration.
+- Check chart-operator deployment status before initiating helm 3 migration.
 
 ## [v1.1.3] 2020-05-26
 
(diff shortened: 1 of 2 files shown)


service/controller/app/resource/releasemigration/create.go
```diff
@@ -69,7 +82,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return nil
 	}
 
-	if strings.ToLower(cr.Status.Release.Status) != helmclient.StatusDeployed {
+	if deploy.Status.ReadyReplicas > 0 {
```
tomahawk28, line 89: this is the point we are using chart-operator deployment status.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## Unreleased
 
+### Changed
+
+- Checking chart-operator deployment status before initiate helm 3 migration.
```
rossf7, line 12:
```suggestion
- Check chart-operator deployment status before initiating helm 3 migration.
```

service/controller/app/resource/releasemigration/create.go
```diff
@@ -54,11 +53,29 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
 
-	v, err := semver.NewVersion(cr.Status.AppVersion)
+	deploy, err := cc.Clients.K8s.K8sClient().AppsV1().Deployments(key.Namespace(cr)).Get(cr.Name, metav1.GetOptions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// extract spec container image
+	image := deploy.Spec.Template.Spec.Containers[0].Image
+	tag := strings.Split(image, ":")[1]
```
rossf7, line 76: In newer releases we also have a label on the deployed `app.kubernetes.io/version: 0.13.0`.

But we might need to upgrade from a version that misses this label. So I think using the image is safest. 👍

service/controller/app/resource/releasemigration/create.go
rossf7: Nice and good catch on finding this.

PR #350, opened 2020-05-29
readyReplica should be same as zero when it's not ready

tomahawk28:
Wrong if statement checking. 🤦 

I should check `readyReplicas` same as 0 to cancel migration.


tomahawk28:
going with a ping.

PR #359, opened 2020-06-03
Process app CRs with both version 0.0.0 and 1.0.0

rossf7:
This is so we don't need to revert the opsctl and architect changes. It will be removed in 1.0.6 when we run separate app-operator instances again.

## Checklist


rossf7:
Tested on `gorgoth`.

PR #362, opened 2020-06-05
Label selector for unique

tomahawk28:
With lableSelector in app-operator controller, We don't need to check the labels from resource_set anymore, and that's good for logging and reconciliation timing. 

logs > 
```
D 06/05 09:01:32 /apis/application.giantswarm.io/v1alpha1/namespaces/54naz/apps/node-exporter canceling reconciliation | operatorkit/controller/controller.go:464 | controller=app-operator | event=update | loop=289 | version=403269248
D 06/05 09:01:32 /apis/application.giantswarm.io/v1alpha1/namespaces/54naz/apps/node-exporter reconciled object | operatorkit/controller/controller.go:446 | controller=app-operator | event=update | loop=289 | version=403269248
D 06/05 09:01:32 /apis/application.giantswarm.io/v1alpha1/namespaces/w8g2o/apps/node-exporter reconciling object | operatorkit/controller/controller.go:444 | controller=app-operator | event=update | loop=290 | version=403269254
D 06/05 09:01:32 /apis/application.giantswarm.io/v1alpha1/namespaces/w8g2o/apps/node-exporter finding resource set | operatorkit/controller/controller.go:459 | controller=app-operator | event=update | loop=290 | version=403269254
D 06/05 09:01:32 /apis/application.giantswarm.io/v1alpha1/namespaces/w8g2o/apps/node-exporter did not find resource set | operatorkit/controller/controller.go:463 | controller=app-operator | event=update | loop=290 | version=403269254
D 06/05 09:01:32 /apis/application.giantswarm.io/v1alpha1/namespaces/w8g2o/apps/node-exporter canceling reconciliation | operatorkit/controller/controller.go:464 | controller=app-operator | event=update | loop=290 | version=403269254
D 06/05 09:01:32 /apis/application.giantswarm.io/v1alpha1/namespaces/w8g2o/apps/node-exporter reconciled object | operatorkit/controller/controller.go:446 | controller=app-operator | event=update | loop=290 | version=403269254
```

## Checklist

- [ ] Update changelog in CHANGELOG.md.

PR #363, opened 2020-06-05
helm 2 supporting; Label selector for unique

tomahawk28:
Same as https://github.com/giantswarm/app-operator/pull/362 but adding for helm 2 supported operator.

### DIFF
--- a/service/controller/app/app.go
+++ b/service/controller/app/app.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/giantswarm/app-operator/pkg/project"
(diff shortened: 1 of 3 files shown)


tomahawk28:
Tested in `gauss`, both app-operator-unique, app-operator-1-0-4 (how we rename the latter one?)

rossf7:
> both app-operator-unique, app-operator-1-0-4 (how we rename the latter one?)

Good question. I think we should rename app-operator-1-0-4 to app-operator-1-0-5.

While we have the hardcoded `1.0.0` version label we only want to run 1 TC pod. WDYT?

tomahawk28:
@rossf7 PTAL, I would release new version and test app-collection on gauss.

PR #364, opened 2020-06-10
Checking YAML merge error and show it to user as status. 

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/11469

## Checklist

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -235,6 +235,8 @@ github.com/giantswarm/helmclient v1.0.0 h1:VukapA4dpSEQLr2skPbNjFuHfT/huhaGqL10h
 github.com/giantswarm/helmclient v1.0.0/go.mod h1:2KJ2cg7UTbVnYwpar92Qnm6x60gA6RGIoGATBvAaf/g=
 github.com/giantswarm/helmclient v1.0.2 h1:dlGoHxv81T+mZXZaqpwbQTP+QRf9GpzWq+yHYemFuRw=
 github.com/giantswarm/helmclient v1.0.2/go.mod h1:5jfYwrjmyGq8u7Q51qsq9RzsT/Zs27sTcA34c8dkUf4=
+github.com/giantswarm/helmclient v1.0.3-0.20200609125658-aee859ab1afa h1:XdCQM4Rjdy4Gs5JF1FZ3bfmL+G6RrTbcSvVCGbUlfiU=
+github.com/giantswarm/helmclient v1.0.3-0.20200609125658-aee859ab1afa/go.mod h1:5jfYwrjmyGq8u7Q51qsq9RzsT/Zs27sTcA34c8dkUf4=
 github.com/giantswarm/k8sclient v0.2.0 h1:Az+Knfs2UJrNFoqfOsfkudg+rt3fn5EM1aKrnLSg3fU=
 github.com/giantswarm/k8sclient v0.2.0/go.mod h1:tUgUynwKgMnjnN2PldGmxJW6+8eZn2p7tT+tEhXDwOg=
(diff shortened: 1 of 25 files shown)


go.sum
```diff
@@ -235,6 +235,8 @@ github.com/giantswarm/helmclient v1.0.0 h1:VukapA4dpSEQLr2skPbNjFuHfT/huhaGqL10h
 github.com/giantswarm/helmclient v1.0.0/go.mod h1:2KJ2cg7UTbVnYwpar92Qnm6x60gA6RGIoGATBvAaf/g=
 github.com/giantswarm/helmclient v1.0.2 h1:dlGoHxv81T+mZXZaqpwbQTP+QRf9GpzWq+yHYemFuRw=
 github.com/giantswarm/helmclient v1.0.2/go.mod h1:5jfYwrjmyGq8u7Q51qsq9RzsT/Zs27sTcA34c8dkUf4=
+github.com/giantswarm/helmclient v1.0.3-0.20200609125658-aee859ab1afa h1:XdCQM4Rjdy4Gs5JF1FZ3bfmL+G6RrTbcSvVCGbUlfiU=
```
tomahawk28, line 238: TODOTODO: Revert this to regular version.

tomahawk28:
When user-provided configmap is not correct yaml format, we show them as below in app CR.
```  version: 1.0.0
status:
  appVersion: ""
  release:
    lastDeployed: null
    reason: 'parsing error: failed to parse user configmap, logs from merging: parsing
      src failed error: yaml: line 22: did not find expected key'
    status: configmap-merge-failed
  version: ""
```
if its secret,
```status:
  appVersion: ""
  release:
    lastDeployed: null
    reason: 'parsing error: failed to parse user secret'
    status: secret-merge-failed
  version: ""
```

both results in my local env, I'm going to create 1.0.6 patch which could support helm2.

tomahawk28:
Same as https://github.com/giantswarm/app-operator/pull/365

service/controller/app/resource/chart/current.go
```diff
@@ -27,14 +27,21 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
-	if cc.Status.TenantCluster.IsDeleting {
+	if cc.Status.ChartStatus.Status != "" {
```
rossf7, line 30: How do we know that is the problem? What if we add another type of status message?

service/controller/app/resource/configmap/resource.go
```diff
@@ -7,12 +7,17 @@ import (
...
 	"github.com/giantswarm/app-operator/service/controller/app/values"
 )
 
 const (
 	// Name is the identifier of the resource.
 	Name = "configmap"
+
+	// configmapMergeFailedStatus is set in the CR status when there is an failure during
+	// merge configmaps.
+	configmapMergeFailedStatus = "configmap-merge-failed"
```
rossf7, line 20: See my other comment. I think it would be better to make this public and only check for these statuses. WDYT?

service/controller/app/resource/chart/current.go
tomahawk28: In most cases, we would add status only for failure, but checking status would do no harm. I agree.

service/controller/app/resource/secret/resource.go
```diff
@@ -7,12 +7,17 @@ import (
...
+	"github.com/giantswarm/app-operator/service/controller/app/controllercontext"
 	"github.com/giantswarm/app-operator/service/controller/app/values"
 )
 
 const (
 	// Name is the identifier of the resource.
 	Name = "secret"
+
+	// SecretMergeFailedStatus is set in the CR status when there is an failure during
+	// merge secrets.
```
rossf7, line 19: Comment should be removed.

PR #371, opened 2020-06-16
Use apiextensions with updated openapi validation

rossf7:
Towards giantswarm/giantswarm#11556

Same as https://github.com/giantswarm/app-operator/pull/379 but for Helm 3.

## Checklist


tomahawk28:
Hi, I think you are going to create another PR for helm2? 
If then, I will create another release for it after that merged.

rossf7:
> Hi, I think you are going to create another PR for helm2?

Please check https://github.com/giantswarm/apiextensions/pull/496 the only approval is from Marian.

I need to test it more and then will merge and create the helm2 release. 

The problem is I keep on getting interrupted 😞  but I will block some focus time. 💪

PR #379, opened 2020-06-19
helm2: Make fields optional for App and AppCatalog CRDs

rossf7:
Towards giantswarm/giantswarm#11556

## Checklist


rossf7:
Tested on geckon but we've had problems with optional fields before.

So I won't be deployed till next week but I'd like to get a review.

CHANGELOG.md
```diff
@@ -7,6 +7,23 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## Unreleased
 
+### Changed
+
+- Fix problems with openapi valdidation rules for app and appcatalog CRDs.
+- Make optional fields nullable for app and appcatalog CRDs.
+
```
rossf7, line 14: Good we have the helm2 branch but I want to fix the changelog.

service/controller/app/key/key_test.go
```diff
@@ -11,44 +12,82 @@ import (
...
-			Catalog: "giant-swarm-catalog-name",
-			Config: v1alpha1.AppSpecConfig{
-				ConfigMap: v1alpha1.AppSpecConfigConfigMap{
-					Name:      "giant-swarm-configmap-name",
-					Namespace: "giant-swarm-configmap-namespace",
+	testCases := []struct {
+		name          string
+		obj           v1alpha1.App
+		expectedValue string
+	}{
```
rossf7, line 19: I checked https://github.com/giantswarm/apiextensions/pull/285 and it's pretty old. So we were not using kubebuilder and the optional fields were not set to nullable.

So no changes are required but I extended the unit test coverage. As if we change apiextensions in future and break this we will have a lot of problems.

PR #380, opened 2020-06-19
Using selectors inside collector

tomahawk28:
Two app-operators are sending same metrics over its collectors. it only needs to send metrics for app CRs that are responsible. 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Delete helm-2to3-migration job after migration is finished.
 - Sending metrics with app CR's version in the spec.
-- Only send metrics from app CRs that reconciled from the controller. 
+- Only emit metrics for app CRs reconciled by this instance of the operator.
 
 ## [v1.1.5] 2020-06-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Delete helm-2to3-migration job after migration is finished.
 - Sending metrics with app CR's version in the spec.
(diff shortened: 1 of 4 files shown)


service/collector/app_resource.go
```diff
@@ -115,7 +118,11 @@ func (c *AppResource) Describe(ch chan<- *prometheus.Desc) error {
 }
 
 func (c *AppResource) collectAppStatus(ctx context.Context, ch chan<- prometheus.Metric) error {
-	r, err := c.g8sClient.ApplicationV1alpha1().Apps("").List(metav1.ListOptions{})
+	options := metav1.ListOptions{
+		LabelSelector: key.AppVersionSelector(c.uniqueApp).String(),
```
tomahawk28, line 122: I'm searching for how to use selector directly inside k8sclient, but looks like there's no way. I think it's rather simple.

tomahawk28:
If it's okay, I will implement the same things into helm2.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Delete helm-2to3-migration job after migration is finished.
 - Sending metrics with app CR's version in the spec.
+- Only send metrics from app CRs that reconciled from the controller.
```
rossf7, line 14:
```suggestion
- Only emit metrics for app CRs reconciled by this instance of the operator.
```

service/collector/app_resource.go
rossf7: Yes I don't know of a better way to do this. Thanks for researching it but lets go with this.

rossf7:
> If it's okay, I will implement the same things into helm2.

I missed replying here. Yes please also implement for helm 2. Let's include it in the app-operator 1.0.8 release I'm preparing.

PR #381, opened 2020-06-22
Helm2; using selectors inside collector

tomahawk28:
Same as https://github.com/giantswarm/app-operator/pull/380 but only for helm2

## Checklist


rossf7:
@tomahawk28 Changes look fine. But CI is not happy.

Once this is merged I think we can release 1.0.8? I don't think there is anything else missing.

tomahawk28:
> But CI is not happy.

Yes, I canceled the last two CI tests which were triggered by my mistake. 

If we think that we would have this `helm2` branch for long, I could provide a fix before release 1.0.8, WDYT? it would take not that long.

rossf7:
> But CI is not happy.

@tomahawk28 I just mentioned it because the build step was red. I was expecting the integration tests to be red.

I'd like us to implement the `VERSION` file like @fiunchinho described here. I think this would help with long lived branches like `helm2. WDYT?

https://gigantic.slack.com/archives/C04TGHDEF/p1591257038015600

But since we're short on dev resource we shouldn't spend a lot of time on it at the moment.

PR #425, opened 2020-07-24
Fix changelog links

rossf7:
3 dots are needed for the release automation.


rossf7:
> I found this also in helm2 branch.

Thx and yes its a bit hard to debug. So good both branches are fixed now.

PR #434, opened 2020-07-29
App operator metric

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/12358

## Checklist


service/collector/app_operator.go
```diff
@@ -0,0 +1,173 @@
...
+		return microerror.Mask(err)
+	}
+
+	for version := range appVersions {
+		var ready int32
+
+		if version == project.AppTenantVersion() {
+			// There should be a single app-operator instance with major version
+			// 1 for Helm 2 tenant clusters.
+			ready, err = helm2AppOperatorExists(operatorVersions)
```
rossf7: This part is a bit tricky. For Helm 2 there should only be a single instance with major version 1.

service/collector/app_operator.go
```diff
@@ -0,0 +1,173 @@
...
+		)
+	}
+
+	return nil
+}
+
+func (a *AppOperator) collectAppVersions(ctx context.Context) (map[string]bool, error) {
+	appVersions := map[string]bool{}
+
+	l, err := a.k8sClient.G8sClient().ApplicationV1alpha1().Apps("").List(metav1.ListOptions{})
```
rossf7, line 126: Gets all app CRs in the cluster.

rossf7:
Tested on `ginger`.

I'd like to get a review for this but Jihyuk is AFK this week. @xh3b4sd @mig4 @njuettner If you have time to check this that would help a lot. Thx :)

pkg/label/label.go
```diff
@@ -3,6 +3,13 @@
 package label
 
 const (
+	// App label is used to identify Kubernetes resources.
+	App = "app"
+
+	// AppKubernetesVersion label is used to identify the version of Kubernetes
+	// resources.
+	AppKubernetesVersion = "app.kubernetes.io/version"
```
xh3b4sd, line 11: We can start putting labels and annotations into `apiextensions`. See https://github.com/giantswarm/apiextensions/tree/master/pkg/label.

rossf7:
> When app-operator runs multiple times and emit metrics for specific relationships, isn't this setup messing with the numbers? 

For the existing app metrics app-operator only emits for the CRs its processing. This new metric is for all app CRs but its only emitted by the unique instance that handles CP app CRs.

So I think we're covered. But a dedicated collector may well be a better design. I'll keep it in mind.

pkg/label/label.go
rossf7: Cool but I'll do this as a follow up.

PR #438, opened 2020-07-31
WIP: Support Helm 3 in tenant clusters

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/11748

## Checklist


tomahawk28:
No problem with merging this but do you have an idea how we release new versions after this PR merge, app-operator releasing after this PR would only reconcile the app CR with 2.0.0 appVersion.

rossf7:
Thanks @tomahawk28 that's great news the testing was OK. 

I'd like to break this down and merge the chartcrd resource separately. So the commit history is cleaner. I can do a PR for that.

> but do you have an idea how we release new versions after this PR merge, app-operator releasing after this PR would only reconcile the app CR with 2.0.0 appVersion.

I think we should create a `2.1.0` release once all these changes are merged. We can update the app-operator unique app CR to use that and it reconciles CP app CRs with version `0.0.0`.

When we create platform releases we can use 2.1.0 as the app-operator component and use release-operator to manage this app CR. WDYT?

rossf7:
PR is ready. I still need to test it. https://github.com/giantswarm/app-operator/pull/453

I think we can merge it to master. As the chartcrd resource only executes for tenant cluster app CRs. So we can still do a patch release if we need to fix something. Is that OK?

Maybe one more PR is enough. If not then a temp branch is a good idea. 👍

tomahawk28:
> So we can still do a patch release if we need to fix something. Is that OK? 

Sure, But I'm more interested in a long-term view, what if TC with helm 2 has a problem with the app-operator, so we need to patch it? 
some upstream repo keep the release version as a branch, so why don't we keep `2.0.x-version` branch as so we can update the app-operator for helm2?

rossf7:
> Sure, But I'm more interested in a long-term view, what if TC with helm 2 has a problem with the app-operator, so we need to patch it?

All the app CRs using Helm 2 will keep the version label `1.0.0`. We have the helm2 branch in app-operator so we can make fixes there. To deploy them we can update the app CRs in the app collections. e.g. https://github.com/giantswarm/aws-app-collection/blob/master/helm/aws-app-collection-chart/templates/app-operator-unique-1.0.9.yaml

> some upstream repo keep the release version as a branch, so why don't we keep 2.0.x-version branch as so we can update the app-operator for helm2?

I agree release branches can be useful and at some point there will be app-operator v3. But I'd prefer to wait until we need a release branch.

tomahawk28:
Ah, you're right, we have `helm2` branch to fix it there, Sorry I didn't realize that.

rossf7:
Closing in favor of https://github.com/giantswarm/app-operator/pull/457

PR #446, opened 2020-08-07
Update app alert routing for Atlas and Biscuit

rossf7:
Follow up to https://github.com/giantswarm/g8s-prometheus/pull/989.

I would like to keep the per team mapping. But it should be routed based on the team label on the app metrics. Having app names in the alerts doesn't scale.

This mapping in the configmap is dumb and gets out of date. It is temporary until the Representation of Apps story which is next for Batman after Helm 3. 

With that the maintainers field in Chart.yaml will become the source of truth and team will be a field in the new appcatalogentry CRs.

## Checklist

### DIFF
--- a/helm/app-operator/templates/configmap.yaml
+++ b/helm/app-operator/templates/configmap.yaml
@@ -21,7 +21,7 @@ data:
           defaultTeam: "batman"
           teams: |
             atlas: "g8s-efk,g8s-grafana,g8s-oauth2-proxy,g8s-prometheus,gatekeeper-app,gatekeeper-control-plane-rules"
-            biscuit: "cluster-service,dex-app,g8s-cert-manager,kubernetesd,organization-operator,rbac-operator,tokend,userd,vault-app,vault-exporter"
+            biscuit: "api,cluster-service,dex-app,g8s-cert-manager,kubernetesd,organization-operator,rbac-operator,tokend,userd,vault-app,vault-exporter"
             celestial: "azure-collector,azure-operator,etcd-backup-operator,releases-azure"
             firecracker: "aws-operator,cluster-operator,kiam,releases-aws"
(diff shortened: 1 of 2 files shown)


rossf7:
PTAL and checking the mappings are correct. 🙏 

I took them from the team ownership page.

helm/app-operator/templates/configmap.yaml
```diff
@@ -20,10 +20,11 @@ data:
         apps:
           defaultTeam: "batman"
           teams: |
+            atlas: "g8s-efk,g8s-grafana,g8s-oauth2-proxy,g8s-prometheus,gatekeeper-app,gatekeeper-control-plane-rules"
+            biscuit: "cluster-service,dex-app,g8s-cert-manager,kubernetesd,organization-operator,rbac-operator,tokend,userd,vault-app"
```
kopiczko, line 24: Do we own cluster-service and kuberetesd? I think vault-exporter is missing.

rossf7: Thanks I'll check against the apps dashboard and see if I've missed any.

cluster-service and kubernetesd are both in deprecated components but no team. So not sure. If I remove them it will default to batman.

https://intranet.giantswarm.io/docs/product/team-ownership/#deprecated-components

rossf7: I added vault-exporter and removed route53-manager. I'll do a separate PR for KAAS as some other apps are missing.

kopiczko: `api` is also missing. Should it also be added to biscuit? @JosephSalisbury do you have opinion here?

JosephSalisbury: I'd be fine with it if we need a decision now, but i'll see if we can find a proper ownership - see https://github.com/orgs/giantswarm/projects/10#card-43276660 for product board tracking

kopiczko: Cool, let's go with this then.

rossf7: > I'd be fine with it if we need a decision now, but i'll see if we can find a proper ownership

@JosephSalisbury @kopiczko Thanks, much appreciated. If this changes let us know and we'll update.

PR #447, opened 2020-08-07
Add apps mapping for Firecracker and Celestial

rossf7:
We need to update the apps to team mappings as its out of date. This is used to route app failed alerts.

This mapping in the configmap is dumb and gets out of date. It's temporary until the Representation of Apps story which is next for Batman after Helm 3.

With that the maintainers field in Chart.yaml will become the source of truth and team will be a field in the new appcatalogentry CRs.


rossf7:
Don't think this is controversial. Going with a ping.

PR #454, opened 2020-08-14
Don't wait for chart-operator pod

rossf7:
Towards giantswarm/giantswarm#11748

Since the chart CRD is now created by the chartcrd resource. We don't need to wait for a ready chart-operator pod.

We added this so that chart CRs can be created in the same reconciliation loop. So that pre-installed apps are installed faster in new tenant clusters.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Add chartcrd resource for creating chart CRD in tenant clusters.
 
+### Removed
+
+- Don't wait for chart-operator pod since chart CRD is created by the operator.
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Add chartcrd resource for creating chart CRD in tenant clusters.
 
+### Removed
+
+- Don't wait for chart-operator pod since chart CRD is created by the operator.
```
tomahawk28, line 16:
```suggestion
- Don't wait for chart-operator pod since chart CRD is created by the `chartcrd` resource. 
```
WDYT?

rossf7: Yes that's better. Thx!

PR #455, opened 2020-08-14
Use labels from apiextensions and remove Helm version label

rossf7:
## Checklist


rossf7:
Don't think this is controversial as we didn't use the Helm major version label. Going with a ping.

PR #457, opened 2020-08-17
Remove hardcoded version in app CR version label

rossf7:
Towards giantswarm/giantswarm#11748

## Checklist


rossf7:
Tested on `godsmack`.

integration/test/app/kubeconfig/kubeconfig_test.go
```diff
@@ -189,7 +190,7 @@ func TestAppWithKubeconfig(t *testing.T) {
 				Name:      chartOperatorName,
 				Namespace: namespace,
 				Labels: map[string]string{
-					label.AppOperatorVersion: key.AppOperatorVersion(),
```
tomahawk28, line 192: operator doesn't have any usage for`AppOperatorVersion`, can we delete this function?

rossf7: Thanks, good catch. Done.

PR #466, opened 2020-08-26
Install chart-operator first in integration test

tomahawk28:
I found some integration tests are failed since it didn't install a chart-operator first.


tomahawk28:
all integration tests passed. going with a ping.

PR #469, opened 2020-08-27
Add monitoring label

omegas27:
## Checklist

### DIFF
--- a/helm/app-operator/templates/service.yaml
+++ b/helm/app-operator/templates/service.yaml
@@ -5,6 +5,7 @@ metadata:
   namespace: {{ include "resource.default.namespace" . }}
   labels:
     {{- include "labels.common" . | nindent 4 }}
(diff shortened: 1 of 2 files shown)


helm/app-operator/templates/service.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   namespace: {{ include "resource.default.namespace" . }}
   labels:
     {{- include "labels.common" . | nindent 4 }}
+    giantswarm.io/monitoring: "true"
```
tomahawk28, line 8: I think this label is for tenant cluster, doesn't it?

omegas27: We decided to use the same labels for control-plane components

omegas27: That way, we can use only one service monitor

mig4 (approved):

PR #472, opened 2020-09-01
Adding validation resource

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/13045

## Checklist

### DIFF
--- a/service/controller/app/resource/status/create.go
+++ b/service/controller/app/resource/status/create.go
@@ -31,12 +31,6 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return nil
 	}
 
-	if cc.Status.ClusterStatus.IsUnavailable {
-		r.logger.LogCtx(ctx, "level", "debug", "message", "tenant cluster is unavailable")
-		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
-		return nil
-	}
-
 	var desiredStatus v1alpha1.AppStatus
 
 	if cc.Status.ChartStatus.Status != "" {
@@ -47,6 +41,12 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			},
 		}
 	} else {
+		if cc.Status.ClusterStatus.IsUnavailable {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "tenant cluster is unavailable")
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+			return nil
+		}
+
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding status for chart %#q in namespace %#q", cr.Name, r.chartNamespace))
 
 		chart, err := cc.Clients.K8s.G8sClient().ApplicationV1alpha1().Charts(r.chartNamespace).Get(ctx, cr.Name, metav1.GetOptions{})
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 ### Added
 
 - Add monitoring label
+- Add validation resource that checks if references to other resources exist in
+app CRs. A message is added to the app CR status for the user.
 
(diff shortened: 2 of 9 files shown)


tomahawk28:
1. configmap not found
```
status:
  appVersion: ""
  release:
    lastDeployed: null
    reason: 'not found error: configmap `dd` in namespace `dd` not found'
    status: configmap-merge-failed
  version: ""
```
2. secret not found
```
status:
  appVersion: ""
  release:
    lastDeployed: null
    reason: 'not found error: secret `dd` in namespace `giantswarm` not found'
    status: secret-merge-failed
  version: ""
```

tomahawk28:
When no namespace on the resource, 
```
status:
  appVersion: ""
  release:
    lastDeployed: null
    reason: namespace is not specified for secret `d`
    status: resource-not-found
```

tomahawk28:
Changed as below format; 

```
status:
  appVersion: ""
  release:
    lastDeployed: null
    reason: secrets "x" not found
    status: resource-not-found
```

service/controller/app/resource/status/create.go
```diff
@@ -47,6 +41,12 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			},
 		}
 	} else {
+		if cc.Status.ClusterStatus.IsUnavailable {
```
tomahawk28, line 44: I moved these statements here since we need to update the app CR status even when the tenant cluster is not available.

rossf7 (changes requested):
I really appreciate the efforts here and the desire to quickly improve the developer experience for customers. But this is going to need some rework and IMO we shouldn't be implementing validation logic like this without test coverage.

It would have been better to outline the design in https://github.com/giantswarm/giantswarm/issues/13045 once we decided to change it and add the new validation resource.

Soon we will want to create an app-admission-controller that also has this logic. So I think it would be better to put the validation logic in `pkg/validation` so it can be extracted later. Something like 

```go
func ValidateApp(ctx context.Context, cr *v1alpha1.App) (string, error) {
	...
	return reason, nil
}
```

I think the validation resource should execute first. If it doesn't pass then we set the CR status and cancel the reconciliation. Passing the status via the controller context is complex and is deprecated functionality in operatorkit.

For testing the ideal for me would be unit tests for `ValidateApp` with each case and the error we want to display to users. Plus a single step in the basic integration test to make sure the CR status is set OK. WDYT?

Also we'd talked about releasing this tomorrow for the Thursday deploys thing. But there is no rush here and this is my bad for not communicating that clearly.

IMO its much better to spend the time here to improve the design and make sure we have test coverage.

tomahawk28:
> I think the validation resource should execute first. If it doesn't pass then we set the CR status and cancel the reconciliation. Passing the status via the controller context is complex and is deprecated functionality in operatorkit.

That's right, we are deprecating controller context so I think it makes sense that not to use that as you suggested. 

One thing I would like to point out, that `ValidateApp` spec may could be optimized as below, 

```golang
func ValidateApp(ctx context.Context, cr *v1alpha1.App) error {
	...
	return reason // or return nil
}
```

rather than using string as first parameter, we could utilize error variable returned by this function, WDYT?

tomahawk28:
> Soon we will want to create an app-admission-controller that also has this logic. So I think it would be better to put the validation logic in pkg/validation so it can be extracted later.

Even if we put this function in `resource.go` in `validation`, extraction would work as same. So why not in the `resource.go` file?

tomahawk28:
tested in `ghost`
Logs:
```
W 09/03 18:15:45 /apis/application.giantswarm.io/v1alpha1/namespaces/giantswarm/apps/chart-operator-unique validation validation error validation error: configmap `d` in namespace `dd` not found | app-operator/v2/service/controller/app/resource/validation/create.go:24 | controller=app-operator | event=update | loop=279 | version=128656135
```

yaml:

```
status:
  appVersion: ""
  release:
    lastDeployed: null
    reason: 'validation error: configmap `d` in namespace `dd` not found'
    status: resource-not-found
  version: ""
```

rossf7 (approved):
LGTM ❤️ 

@tomahawk28 Thanks this works great and will improve the developer experience. As well as hopefully reducing support requests.🤞

PR #474, opened 2020-09-04
Execute validation resource first

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/13045

## Checklist

### DIFF
--- a/service/controller/app/resources.go
+++ b/service/controller/app/resources.go
@@ -262,13 +262,15 @@ func newAppResources(config appResourcesConfig) ([]resource.Interface, error) {
 	}
 
 	resources := []resource.Interface{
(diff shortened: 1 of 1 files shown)


service/controller/app/resources.go
```diff
@@ -262,13 +262,15 @@ func newAppResources(config appResourcesConfig) ([]resource.Interface, error) {
 	}
 
 	resources := []resource.Interface{
+		// validationResource checks CRs for validation errors and sets the CR status.
+		validationResource,
```
rossf7, line 266: Moving the validation first so its before the clients resource. As the clients resource needs to use the kubeconfig secret.

rossf7:
Tested on `gauss`.

tomahawk28 (approved):

PR #487, opened 2020-09-07
helm2: Use v1.0.10 not v1.1.0

rossf7:
This has to be a patch bump because we already released v1.1.0 and it has Helm 3 support.

## Checklist


rossf7 (commented):
Going with a ping.

PR #489, opened 2020-09-14
Adding resource version on chart CRs

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/13196

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
+- Put resourceVersion on the desired chart status. 
+
 ### Added
 
 - Add monitoring label
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,9 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
-### Changed
-
-- Put resourceVersion on the desired chart status. 
-
 ### Added
 
+- Add resource version for chart configmaps and secrets to the chart CR to reduce latency of update events.
 - Add monitoring label
 - Add validation resource that checks if references to other resources exist in
 app CRs. A message is added to the app CR status for the user.
--- a/service/controller/app/resource/chart/desired.go
+++ b/service/controller/app/resource/chart/desired.go
@@ -10,6 +10,7 @@ import (
 	"github.com/giantswarm/appcatalog"
 	"github.com/giantswarm/microerror"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/giantswarm/app-operator/v2/pkg/annotation"
 	"github.com/giantswarm/app-operator/v2/pkg/project"
@@ -39,7 +40,7 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		return chartCR, nil
 	}
 
-	config, err := r.generateConfig(ctx, cr, cc.AppCatalog, r.chartNamespace)
+	config, err := generateConfig(ctx, cc.Clients.K8s.K8sClient(), cr, cc.AppCatalog, r.chartNamespace)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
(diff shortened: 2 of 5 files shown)


tomahawk28:
Tested in `ghost`.

service/controller/app/resource/chart/desired.go
```diff
@@ -85,28 +89,42 @@ func generateAnnotations(input map[string]string) map[string]string {
 	return annotations
 }
 
-func generateConfig(cr v1alpha1.App, appCatalog v1alpha1.AppCatalog, chartNamespace string) v1alpha1.ChartSpecConfig {
+func (r *Resource) generateConfig(ctx context.Context, cr v1alpha1.App, appCatalog v1alpha1.AppCatalog, chartNamespace string) (v1alpha1.ChartSpecConfig, error) {
```
tomahawk28, line 92: This function needs k8sclient from now since it has to fetch `resourceVersion` from configmap/secret objects.

service/controller/app/resource/chart/desired.go
```diff
@@ -85,28 +89,42 @@ func generateAnnotations(input map[string]string) map[string]string {
...
+		}
+
+		secretConfig := v1alpha1.ChartSpecConfigSecret{
+			Name:            secretName,
+			Namespace:       chartNamespace,
+			ResourceVersion: secret.GetResourceVersion(),
 		}
 
-		config.Secret = secret
+		config.Secret = secretConfig
```
tomahawk28, line 124: Since `secret` already exists, change the name to `secretConfig`.

service/controller/app/resource/chart/desired.go
```diff
@@ -85,28 +89,42 @@ func generateAnnotations(input map[string]string) map[string]string {
 	return annotations
 }
 
-func generateConfig(cr v1alpha1.App, appCatalog v1alpha1.AppCatalog, chartNamespace string) v1alpha1.ChartSpecConfig {
+func (r *Resource) generateConfig(ctx context.Context, cr v1alpha1.App, appCatalog v1alpha1.AppCatalog, chartNamespace string) (v1alpha1.ChartSpecConfig, error) {
 	config := v1alpha1.ChartSpecConfig{}
 
 	if hasConfigMap(cr, appCatalog) {
+		configMapName := key.ChartConfigMapName(cr)
+		cm, err := r.k8sClient.CoreV1().ConfigMaps(chartNamespace).Get(ctx, configMapName, metav1.GetOptions{})
```
rossf7, line 97: The CM is in the tenant cluster so I think this needs to check for `tenant.IsAPINotAvailable(err)`. Also if the configmap doesn't exist I think we should just not set the resource version. 

Same for getting the secret. WDYT?

service/controller/app/resource/chart/desired.go
rossf7: For TC app CRs the k8s client is for the tenant. So we need to use the client from the controller context.

rossf7 (commented):
Direction is good but needs a couple of changes.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
+- Put resourceVersion on the desired chart status.
```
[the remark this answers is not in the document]
tomahawk28: > Changed is only for breaking changes and requires a major bump.

Didn't know that earlier. I will adjust the changelog as you suggest. Thanks!

service/controller/app/resource/chart/desired.go
tomahawk28: Good call, I will extend the unittest for both cases.

tomahawk28: I found two points we can discuss further. 

> The CM is in the tenant cluster so I think this needs to check for tenant.IsAPINotAvailable(err)

the operator checks this in `current.go` and we cancel the resource when the tenant cluster is unavailable. 
https://github.com/giantswarm/app-operator/blob/e89b6ff14e49ed4834b3734b9adb183e7c098423/service/controller/app/resource/chart/current.go#L65
So we don't need the same error checking code for `generateConfig` function. 

> Also if the configmap doesn't exist I think we should just not set the resource version.

Actually I think the app-operator has to leave the config as blank when it could not find the relevant resource. Otherwise, chart-operator would try to fetch that resource on their own reconciliation. @rossf7 WDYT?

rossf7: > the operator checks this in current.go and we cancel the resource when the tenant cluster is unavailable.

OK cool I just wanted to be sure we handle the TC being unavailable. But like you said since this is done in current.go I think this is fine. 👍 

> Actually I think the app-operator has to leave the config as blank when it could not find the relevant resource.

Yes we just need to set the resource version to speed things up. If it doesn't exist I think the behavior should be the same as now.

service/controller/app/resource/chart/resource.go
```diff
@@ -24,8 +23,7 @@ const (
 // Config represents the configuration used to create a new chart resource.
 type Config struct {
 	// Dependencies.
-	G8sClient versioned.Interface
-	Logger    micrologger.Logger
+	Logger micrologger.Logger
```
rossf7, line 26: ❤️

rossf7 (approved):
Nice! Great to be starting to speed things up. 🚀 

LGTM when tested

tomahawk28:
Tested in `ghost`

PR #491, opened 2020-09-16
helm2: Remove collector as it moved to app-exporter

rossf7:
Collector has been moved to app-exporter. Already removed from master. We can remove it from here too.

## Checklist


tomahawk28 (approved):
LGTM

PR #497, opened 2020-09-18
Check configmap/secret struct for updating

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/13309

## Checklist

### DIFF
--- a/service/controller/app/resource/configmap/resource.go
+++ b/service/controller/app/resource/configmap/resource.go
@@ -3,6 +3,7 @@ package configmap
 import (
 	"reflect"
 
+	"github.com/ghodss/yaml"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


service/controller/app/resource/configmap/resource.go
```diff
@@ -86,7 +87,24 @@ func equals(a, b *corev1.ConfigMap) bool {
...
+			return false
+		}
+
+		err = yaml.Unmarshal([]byte(b.Data["values"]), &dest)
+		if err != nil {
+			return false
+		}
+	}
+
+	if !reflect.DeepEqual(source, dest) {
```
tomahawk28, line 107: Comparing two values as a struct, not YAML strings would not trigger the constant updates.

tomahawk28:
Tested in `geckon`

rossf7 (approved):
Nice, good catch. Is there anywhere else we need this? Like in the values service.

LGTM when tested manually and tests are green.

tomahawk28:
Values service requires k8sclient, but we don't need that for map comparison. Let's put those changes into each resource.

tomahawk28:
Tested in `geckon`, no further changes in configmaps, secrets.

PR #517, opened 2020-10-05
Use apptestctl to setup integration tests

rossf7:
Simplifying the test setup by using `apptestctl` but  with `--install-operators` flag set to false. As we need to control the version of the operators in the test setup.


rossf7:
Will do this later.

PR #535, opened 2020-10-16
Helm2; Release v1.0.13

tomahawk28:
Release chart-operator install patch fix.


tomahawk28:
There was a problem with devctl, so I'm going to release this by the old method.

tomahawk28:
going with a ping

PR #546, opened 2020-10-20
Move key pkg

rossf7:
Prep for adding new appcatalogentry resource.

Currently we have separate key packages for the app and appcatalog controller. We should have a single key package under service to match other operators like aws-operator.


tomahawk28 (approved):
LGTM

PR #547, opened 2020-10-21
Add appcatalogentry resource with delete support

rossf7:
Towards giantswarm/giantswarm#10930

Replaces the empty resource with the new resource for the appcatalog controller.

This PR only implements delete to make it easier to review.


service/controller/appcatalog/resources.go
```diff
@@ -24,13 +29,23 @@ func newAppCatalogResources(config appCatalogResourcesConfig) ([]resource.Interf
...
+
+		appCatalogEntryResource, err = appcatalogentry.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
 	}
 
 	resources := []resource.Interface{
-		emptyResource,
+		appCatalogEntryResource,
```
rossf7, line 48: The empty resource gets replaced.

rossf7:
Tested on `ghost`.

service/controller/appcatalog/resource/appcatalogentry/resource.go
```diff
@@ -0,0 +1,79 @@
...
+	entries, err := r.k8sClient.G8sClient().ApplicationV1alpha1().AppCatalogEntries(metav1.NamespaceDefault).List(ctx, lo)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	for _, entry := range entries.Items {
+		currentEntryCRs[entry.Name] = entry.DeepCopy()
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("got %d appcatalogentries for appcatalog %#q", len(currentEntryCRs), cr.Name))
```
tomahawk28, line 76: Is that something we can print out in `EnsureDeleted`?

tomahawk28 (approved):
Just a one comment LGTM

service/controller/appcatalog/resource/appcatalogentry/resource.go
rossf7: Good point. I've changed it to `getting current appcatalogentries` and `got x current appcatalogentries`.

There is also logs for deleting the CRs so it should be clear now.

PR #548, opened 2020-10-21
Refactor key package for app CRs

rossf7:

### DIFF
--- a/service/controller/app/resource/appcatalog/create.go
+++ b/service/controller/app/resource/appcatalog/create.go
@@ -10,7 +10,7 @@ import (
 
 // EnsureCreated gets the appCatalog CR specified in the provided app CR.
 func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
-	cr, err := key.ToCustomResource(obj)
+	cr, err := key.ToApp(obj)
 	if err != nil {
 		return microerror.Mask(err)
 	}
--- a/service/controller/key/app.go
+++ b/service/controller/key/app.go
@@ -73,6 +73,26 @@ func AppStatus(customResource v1alpha1.App) v1alpha1.AppStatus {
 	return customResource.Status
 }
 
+func AppVersionSelector(unique bool) labels.Selector {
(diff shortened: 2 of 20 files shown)


service/controller/app/resource/appcatalog/create.go
```diff
@@ -10,7 +10,7 @@ import (
 
 // EnsureCreated gets the appCatalog CR specified in the provided app CR.
 func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
-	cr, err := key.ToCustomResource(obj)
+	cr, err := key.ToApp(obj)
```
rossf7, line 13: Package got flattened to support app and appcatalog CRs. `ToAppCatalog` already exists so aligning the naming.

service/controller/key/app.go
```diff
@@ -41,14 +41,10 @@ func AppCatalogSecretNamespace(customResource v1alpha1.AppCatalog) string {
 	return customResource.Spec.Config.Secret.Namespace
 }
 
-// AppConfigMapName returns the name of the configmap that stores app level
-// config for the provided app CR.
```
rossf7, line 45: Some of the key funcs have godoc and others don't. I'd like this to be consistent.

The names are clear so I think we should just drop the godoc.

service/controller/key/app.go
```diff
@@ -73,6 +65,26 @@ func AppStatus(customResource v1alpha1.App) v1alpha1.AppStatus {
...
+	}
+	s := fmt.Sprintf("%s=%s", label.AppOperatorVersion, version)
+
+	selector, err := labels.Parse(s)
+	if err != nil {
+		panic(fmt.Sprintf("failed to parse selector %#q with error %#q", s, err))
+	}
+
+	return selector
+}
```
rossf7, line 86: Last change just fixes the ordering.

rossf7 (commented):
Tested on `ghost`.

tomahawk28 (approved):

PR #550, opened 2020-10-23
Add integration test for appcatalogentry resource

rossf7:
Towards giantswarm/giantswarm#10930

## Checklist

### DIFF
--- a/integration/setup/setup.go
+++ b/integration/setup/setup.go
@@ -56,11 +56,12 @@ func installResources(ctx context.Context, config Config) error {
 		}
 	}
 
-	// Create AppCatalog and App CRDs. The Chart CRD is created by the operator
+	// Create CRDs. The Chart CRD is created by the operator
 	// for the kubeconfig test that bootstraps chart-operator.
 	crds := []string{
-		"AppCatalog",
 		"App",
+		"AppCatalog",
+		"AppCatalogEntry",
 	}
 
 	{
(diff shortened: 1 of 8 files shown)


integration/setup/setup.go
```diff
@@ -117,10 +118,11 @@ func installResources(ctx context.Context, config Config) error {
 				},
 			},
 		}
+		// Release is named app-operator-unique as some functionality is only
+		// implemented for the unique instance.
 		opts := helmclient.InstallOptions{
-			ReleaseName: project.Name(),
+			ReleaseName: fmt.Sprintf("%s-unique", project.Name()),
```
rossf7, line 124: Since the appcatalogentry CRs are only generated by the unique instance.

integration/test/appcatalog/appcatalogentry/app_catalog_entry_test.go
```diff
@@ -0,0 +1,159 @@
...
+			}
+
+			return nil
+		}
+
+		n := func(err error, t time.Duration) {
+			config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("appcatalogentry CRs still exist: retrying in %s", t), "stack", fmt.Sprintf("%v", err))
+		}
+
+		b := backoff.NewConstant(5*time.Minute, 15*time.Second)
```
tomahawk28, line 153: Can we change it to `MaxRetry` if it would be reconciled by the event?

tomahawk28 (commented):
Just one comment :)

integration/test/appcatalog/appcatalogentry/app_catalog_entry_test.go
rossf7: Cool yes its triggered by the event so using max retry is probably faster.

Updated. PTAL

tomahawk28 (approved):
LGTM, let's merge after CI sees green.

PR #563, opened 2020-10-28
Status endpoint in app-operator

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/pull/13134

the new endpoint would be like this; `PATCH http://app-operator.giantswarm:8000/status/giantswarm/userd-unique/`

I will create separate PR for URL annotation into chart CRs. 

## Checklist

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,12 +3,15 @@ package main
 import (
 	"context"
 	"fmt"
-
+	applicationv1alpha1 "github.com/giantswarm/apiextensions/v2/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/k8sclient/v4/pkg/k8sclient"
+	"github.com/giantswarm/k8sclient/v4/pkg/k8srestconfig"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/microkit/command"
 	microserver "github.com/giantswarm/microkit/server"
 	"github.com/giantswarm/micrologger"
 	"github.com/spf13/viper"
+	"k8s.io/client-go/rest"
 
 	"github.com/giantswarm/app-operator/v2/flag"
 	"github.com/giantswarm/app-operator/v2/pkg/project"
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -44,14 +47,52 @@ func mainWithError() (err error) {
...
+				},
+			}
+
+			restConfig, err = k8srestconfig.New(c)
+			if err != nil {
+				panic(err)
+			}
+		}
+
+		var k8sClient k8sclient.Interface
```
tomahawk28, line 71: Since both `server` and `service` need k8sclient object, creating a k8sclient object from `main.go` makes sense.

server/endpoint/status/endpoint.go
```diff
@@ -0,0 +1,158 @@
...
+
+		defer r.Body.Close()
+		err := json.NewDecoder(r.Body).Decode(&request)
+		if err != nil {
+			return nil, microerror.Maskf(decodeFailedError, "%v", err.Error())
+		}
+
+		namespace, _ := ctx.Value("app_namespace").(string)
+		name, _ := ctx.Value("app_name").(string)
+		request.AppNamespace = namespace
```
tomahawk28, line 65: This is not fancy like `api` repository does, but I would like to keep it as simple.

server/endpoint/status/endpoint.go
```diff
@@ -0,0 +1,158 @@
...
+		currentCR.Status = desiredStatus
+
+		_, err = e.k8sClient.G8sClient().ApplicationV1alpha1().Apps(request.AppNamespace).UpdateStatus(ctx, currentCR, metav1.UpdateOptions{})
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		e.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("status set for app %#q in namespace %#q", request.AppName, request.AppNamespace))
+
+		return nil, nil
```
tomahawk28, line 118: Return nothing but 200 response in case of no error, is that okay?

server/endpoint/status/endpoint.go
rossf7: Yes api has middlewares and other stuff we don't need. Good call to keep it simple.

server/endpoint/status/endpoint.go
rossf7: Yes 200 makes sense since it's updating the status. I thought maybe 202 Accepted but the processing is not async.

server/endpoint/status/request.go
```diff
@@ -0,0 +1,13 @@
...
+import v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+type Request struct {
+	AppName      string  `json:"app_name"`
+	AppNamespace string  `json:"app_namespace"`
+	AppVersion   string  `json:"app_version"`
+	Version      string  `json:"version"`
+	LastDeployed v1.Time `json:"last_deployed"`
+	Status       string  `json:"status"`
+	Reason       string  `json:"reason"`
```
rossf7, line 12: Should we order the fields?

server/server.go
```diff
@@ -41,8 +45,9 @@ func New(config Config) (microserver.Server, error) {
 	var endpointCollection *endpoint.Endpoint
 	{
 		c := endpoint.Config{
-			Logger:  config.Logger,
-			Service: config.Service,
+			Logger:    config.Logger,
+			Service:   config.Service,
+			K8sClient: config.K8sClient,
```
rossf7, line 50: Let's order these too. To keep things tidy.

rossf7 (commented):
Direction is nice here. I left some comments but just minor stuff.

I had thought that the endpoint would just add an annotation to the app CR to trigger the status resource.

I like this approach instead. The operator will check after the resync period but should see no change. One downside I see is if we change the status we need to update the resource and the endpoint.

For now we'll only use the internal cluster IP service. For tenant clusters we'll need TLS to secure the traffic.

server/server.go
```diff
@@ -20,17 +20,20 @@ import (
...
+	Service   *service.Service
+
+	Viper *viper.Viper
 }
 
 // New creates a new server object with given configuration.
 func New(config Config) (microserver.Server, error) {
 	var err error
 
+	if config.K8sClient == nil {
```
tomahawk28, line 34: Also found out some config checkpoints are missing. Added in here.

rossf7 (approved):
LGTM

PR #568, opened 2020-10-30
Creating Ingress to connect status endpoint from TCs

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/13201

Ingress object would be created only if this is not `-unique` release which means app-operator for TCs. 

## Checklist

### DIFF
--- a/helm/app-operator/templates/configmap.yaml
+++ b/helm/app-operator/templates/configmap.yaml
@@ -17,7 +17,11 @@ data:
       app:
         unique: {{ include "resource.app.unique" . }}
       chart:
+        {{- if hasSuffix "-unique" .Release.Name }}
         webhookBaseURL: 'http://{{ include "resource.default.name" . }}.{{ include "resource.default.namespace" . }}.svc.cluster.local:{{ .Values.port }}'
+        {{- else }}
+        webhookBaseURL: 'http://{{ include "resource.default.name" . }}.{{ .Values.Installation.V1.Kubernetes.API.Address }}'
+        {{- end }}
       helm:
         http:
           clientTimeout: '{{ .Values.Installation.V1.Helm.HTTP.ClientTimeout }}'
--- a/helm/app-operator/templates/ingress.yaml
+++ b/helm/app-operator/templates/ingress.yaml
@@ -15,7 +15,7 @@ spec:
       - {{ include "resource.default.name" . }}.{{ .Values.Installation.V1.Kubernetes.API.Address }}
       secretName: {{ include "resource.default.name" . }}-ingress
   rules:
-    - host: {{ .Values.Installation.V1.Monitoring.Prometheus.Host }}
+    - host: {{ include "resource.default.name" . }}.{{ .Values.Installation.V1.Kubernetes.API.Address }}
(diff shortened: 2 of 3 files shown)


tomahawk28:
Tested from `gauss`

```
# creating curl pod from TCs. 
kubectl run curl-jhbok --image=radial/busyboxplus:curl -i --tty --rm

[ root@curl-jhbok:/ ]$ curl -sSL -D - https://app-operator-dev.g8s.gauss.eu-west-1.aws.gigantic.io
HTTP/1.1 200 OK
Server: nginx/1.15.6
Date: Fri, 30 Oct 2020 14:35:03 GMT
Content-Type: application/json; charset=utf-8
Content-Length: 466
Connection: keep-alive
Vary: Accept-Encoding
Strict-Transport-Security: max-age=15724800; includeSubDomains

{"description":"The app-operator manages apps in Kubernetes clusters.","git_commit":"334fcb890955785d101b0626184547be9fc5d3be","go_version":"go1.14.1","name":"app-operator","os_arch":"linux/amd64","source":"https://github.com/giantswarm/app-operator","version":"2.6.1-dev","version_bundles":[{"changelogs":[{"component":"app-operator","description":"Added initial version.","kind":"added","urls":null}],"components":[],"name":"app-operator","version":"2.6.1-dev"}]}
```

helm/app-operator/templates/configmap.yaml
```diff
@@ -17,7 +17,11 @@ data:
       app:
         unique: {{ include "resource.app.unique" . }}
       chart:
+        {{- if hasSuffix "-unique" .Release.Name }}
         webhookBaseURL: 'http://{{ include "resource.default.name" . }}.{{ include "resource.default.namespace" . }}.svc.cluster.local:{{ .Values.port }}'
+        {{- else }}
+        webhookBaseURL: 'http://{{ include "resource.default.name" . }}.{{ .Values.Installation.V1.Kubernetes.API.Address }}'
```
tomahawk28, line 23: If the release name is `app-operator-dev` in `gauss` installation, this would be `app-operator-dev.g8s.gauss.eu-west-1.aws.gigantic.io`

rossf7: So we use the service for the instance, like `app-operator-2-3-5`?  I think that's the right approach. 

The TLS handling looks good but can we also use IP whitelisting? So only TCs can connect?

tomahawk28: Do you mean annotation `nginx.ingress.kubernetes.io/whitelist-source-range:`? but how do we set the IP range only for TCs?

tomahawk28: Okay, @corest confirmed that this is not possible since tenant clusters in AWS are using NAT gateway, which is dynamic, so we can't set the whitelist as upfront.

rossf7: Yes whitelist isn't going to work. Maybe we can use an auth token? Let's look into that next week.

rossf7 (approved):
LGTM

PR #570, opened 2020-11-03
Using app CR's resourceVersion as webhook token. 

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/13201

We need an authentication token for app-operator's webhook endpoint. 
`app` CR's resourceVersion is perfect for that matter since we can also reduce the duplicate status updating jobs. 

## Checklist

### DIFF
--- a/pkg/annotation/annotation.go
+++ b/pkg/annotation/annotation.go
@@ -30,4 +30,7 @@ const (
 
 	// WebhookURL is the URL that chart-operator reports chart updates.
 	WebhookURL = "webhook-url"
+
+	// WebhookToken is the string that chart-operator send together when updating status.
+	WebhookToken = "webhook-token"
 )
--- a/server/endpoint/status/endpoint.go
+++ b/server/endpoint/status/endpoint.go
@@ -13,6 +13,8 @@ import (
 	kitendpoint "github.com/go-kit/kit/endpoint"
 	kithttp "github.com/go-kit/kit/transport/http"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/app-operator/v2/service/controller/key"
 )
 
 const (
@@ -82,6 +84,19 @@ func (e Endpoint) Endpoint() kitendpoint.Endpoint {
 		var err error
 
 		request := r.(Request)
+
+		cr, err := e.k8sClient.G8sClient().ApplicationV1alpha1().Apps(request.AppNamespace).Get(ctx, request.AppName, metav1.GetOptions{})
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		if !key.InCluster(*cr) {
+			if request.Token == "" {
+				return nil, microerror.Maskf(wrongTokenError, "token empty")
+			} else if cr.GetResourceVersion() != request.Token {
+				return nil, microerror.Maskf(wrongTokenError, "incorrect token %#q ", request.Token)
+			}
+		}
 
 		desiredStatus := v1alpha1.AppStatus{
 			AppVersion: request.AppVersion,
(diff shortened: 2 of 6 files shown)


service/controller/app/resource/chart/desired.go
```diff
@@ -74,16 +74,19 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 
 	annotations := generateAnnotations(cr.GetAnnotations())
 
-	if key.InCluster(cr) {
```
tomahawk28, line 77: Previously we only put webhook annotation if it's an in-cluster object. Not anymore.

tomahawk28:
Tested in `gauss`

1. empty token
```
[ root@curl-jhbok:/ ]$ curl --request PATCH https://app-operator-dev.g8s.gauss.eu-
west-1.aws.gigantic.io/status/9y0a9/chart-operator --data '{"name": "new name"}' -
H "content-type: application/json"
{"code":"INTERNAL_ERROR","error":"wrong token error: token empty","from":"app-operator"}
```

2. wrong token
```
curl --request PATCH https://app-operator-dev.g8s.gauss.eu-
west-1.aws.gigantic.io/status/9y0a9/chart-operator --data '{"status": "deployed",
"token": "1"}' -H "content-type: application/json"
{"code":"INTERNAL_ERROR","error":"wrong token error: incorrect token `1` ","from":"app-operator"}
```

3. correct token
```
curl --request PATCH https://app-operator-dev.g8s.gauss.eu-
west-1.aws.gigantic.io/status/9y0a9/chart-operator --data '{"status": "deployed",
"token": "48994325"}' -H "content-type: application/json"
null
```

pkg/annotation/annotation.go
```diff
@@ -30,4 +30,7 @@ const (
 
 	// WebhookURL is the URL that chart-operator reports chart updates.
 	WebhookURL = "webhook-url"
+
+	// WebhookToken is the string that chart-operator send together when updating status.
+	WebhookToken = "webhook-token"
```
rossf7, line 35: Lets keep the annotations ordered.

server/endpoint/status/endpoint.go
```diff
@@ -83,6 +85,19 @@ func (e Endpoint) Endpoint() kitendpoint.Endpoint {
...
+		cr, err := e.k8sClient.G8sClient().ApplicationV1alpha1().Apps(request.AppNamespace).Get(ctx, request.AppName, metav1.GetOptions{})
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		if !key.InCluster(*cr) {
+			if request.Token == "" {
+				return nil, microerror.Maskf(wrongTokenError, "token empty")
+			} else if cr.GetResourceVersion() != request.Token {
+				return nil, microerror.Maskf(wrongTokenError, "incorrect token %#q ", request.Token)
```
rossf7, line 97: Do we need to worry about race conditions here?

What happens if the resource version changes because of another update to the app CR? Should we just reject the request and log instead or erroring?

Something else may have updated the app CR and changed

rossf7 (commented):
Just a couple of comments.

server/endpoint/status/endpoint.go
tomahawk28: You're right about that. Any changes to metadata lead to the rejection of status update. 
What do you think about using `generation` instead of `resourceVersion`in metadata, since we only change the generation when the spec is changed.

rossf7: Generation is a lot easier to brute force so that is quite a big drawback.

This is a tricky issue. Let's take a step back and see what is the best way to secure the endpoint.

tomahawk28: I thought about the process, and any `resourceVersion` changes before the status update are very unlikely because whenever app CR has been changed, it will update chart CRs with the latest resourceVersion from app CRs. 

The only thing that I don't like is that this would trigger cascading reconciliation into the app-operator and chart-operator *until* chart-operator sees no difference in chart CRs' status.

But this may be the reasonable cost that we have to pay to update the status immediately. WDYT?

tomahawk28:
Closing this in favor of https://github.com/giantswarm/app-operator/pull/571

PR #571, opened 2020-11-04
Utilise Webhook auth token for validation and configmap resource for TC

tomahawk28:
## Checklist

### DIFF
--- a/main.go
+++ b/main.go
@@ -111,7 +111,8 @@ func mainWithError() (err error) {
 				Logger:    newLogger,
 				Service:   newService,
 
-				Viper: v,
+				Viper:            v,
+				WebhookAuthToken: v.GetString(f.Service.Chart.WebhookAuthToken),
 			}
 
 			newServer, err = server.New(c)
--- a/main.go
+++ b/main.go
@@ -148,7 +148,7 @@ func mainWithError() (err error) {
 
 	daemonCommand.PersistentFlags().Bool(f.Service.App.Unique, false, "Whether the operator is deployed as a unique app.")
 	daemonCommand.PersistentFlags().String(f.Service.Chart.Namespace, "giantswarm", "The namespace where chart CRs are located.")
-	daemonCommand.PersistentFlags().String(f.Service.Chart.WebhookAuthToken, "", "The webhook token value to get authentication from webhook URL.")
+	daemonCommand.PersistentFlags().String(f.Service.Chart.WebhookAuthToken, "", "The auth token value for requests to the webhook.")
 	daemonCommand.PersistentFlags().String(f.Service.Chart.WebhookBaseURL, "", "The webhook base URL where chart-operator reports chart updates.")
 	daemonCommand.PersistentFlags().String(f.Service.Helm.HTTP.ClientTimeout, "5s", "HTTP timeout for pulling chart tarballs.")
 	daemonCommand.PersistentFlags().String(f.Service.Image.Registry, "quay.io", "The container registry for pulling Tiller images.")
--- a/main.go
+++ b/main.go
@@ -147,6 +147,7 @@ func mainWithError() (err error) {
 
 	daemonCommand.PersistentFlags().Bool(f.Service.App.Unique, false, "Whether the operator is deployed as a unique app.")
 	daemonCommand.PersistentFlags().String(f.Service.Chart.Namespace, "giantswarm", "The namespace where chart CRs are located.")
+	daemonCommand.PersistentFlags().String(f.Service.Chart.WebhookAuthToken, "", "The webhook token value to get authentication from webhook URL.")
 	daemonCommand.PersistentFlags().String(f.Service.Chart.WebhookBaseURL, "", "The webhook base URL where chart-operator reports chart updates.")
 	daemonCommand.PersistentFlags().String(f.Service.Helm.HTTP.ClientTimeout, "5s", "HTTP timeout for pulling chart tarballs.")
 	daemonCommand.PersistentFlags().String(f.Service.Image.Registry, "quay.io", "The container registry for pulling Tiller images.")
--- a/server/endpoint/status/endpoint.go
+++ b/server/endpoint/status/endpoint.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/giantswarm/app-operator/v2/service/controller/key"
 	"net/http"
+	"strings"
 
 	"github.com/giantswarm/apiextensions/v3/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/k8sclient/v5/pkg/k8sclient"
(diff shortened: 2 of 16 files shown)


server/endpoint/status/endpoint.go
```diff
@@ -99,6 +109,12 @@ func (e Endpoint) Endpoint() kitendpoint.Endpoint {
 			return nil, microerror.Mask(err)
 		}
 
+		if !key.InCluster(*currentCR) {
```
tomahawk28, line 112: Only validate the auth token for TC apps.

tomahawk28:
Tested in `gauss` after creating 12.5.1 TC. 

```
 ~  kubectl run curl-jhbok --image=radial/busyboxplus:curl -i --tty --rm
If you don't see a command prompt, try pressing enter.

# wrong token error
[ root@curl-jhbok:/ ]$ curl --request PATCH -sSL https://app-operator-dev.g8s.gauss.eu-west-1.aws.gigantic.io/statu
s/2i54u/chart-operator --data '{"status": ""}' -H "content-type: application/json" -H "Authorization: 123455"
{"code":"INTERNAL_ERROR","error":"wrong token error: got wrong auth token \"123455\" for `chart-operator` app `2i54u` namespace","from":"app-operator"}


# successful case
[ root@curl-jhbok:/ ]$ curl --request PATCH -sSL https://app-operator-dev.g8s.gauss.eu-west-1.aws.gigantic.io/statu
s/2i54u/chart-operator --data '{"status": ""}' -H "content-type: application/json" -H "Authorization: ***********"
null

**Comment by tomahawk28** [comment]
auth secret had been created in TC.
```
kg get secret auth-token  -o yaml
apiVersion: v1
data:
  token: ODNXMk80QlJ6VmJ3cCozS2pWcmo4JjFNejY4JUImdkE=
kind: Secret
metadata:
  annotations:
    giantswarm.io/notes: DO NOT EDIT. Values managed by app-operator.
  creationTimestamp: "2020-11-04T14:51:48Z"
  labels:
  name: auth-token
  namespace: giantswarm
  resourceVersion: "3041"
  selfLink: /api/v1/namespaces/giantswarm/secrets/auth-token
  uid: 2a54daaf-0546-4c10-8935-403bb6772497
type: Opaque
```

**Comment by tomahawk28** [comment]
I didn't use CRUS for auth secret since there are more than one app-operators could reconcile TC apps at some points. deleting auth token would make a race condition.

File: service/controller/app/resource/authtoken/create.go [hunk 2]
```diff
@@ -0,0 +1,104 @@
...
+		return microerror.Mask(err)
+	}
+
+	if key.InCluster(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q in %#q uses InCluster kubeconfig no need to install webhook auth token", cr.Name, cr.Namespace))
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+		return nil
+	}
+
+	if key.AppName(cr) != key.ChartOperatorAppName {
```
tomahawk28, line 32: Only for chart-operator resource.

tomahawk28:
I will re-write the chart-operator PR https://github.com/giantswarm/chart-operator/pull/629 after merging this.

main.go
```diff
@@ -147,6 +148,7 @@ func mainWithError() (err error) {
 
 	daemonCommand.PersistentFlags().Bool(f.Service.App.Unique, false, "Whether the operator is deployed as a unique app.")
 	daemonCommand.PersistentFlags().String(f.Service.Chart.Namespace, "giantswarm", "The namespace where chart CRs are located.")
+	daemonCommand.PersistentFlags().String(f.Service.Chart.WebhookAuthToken, "", "The webhook token value to get authentication from webhook URL.")
```
rossf7, line 151:
```suggestion
	daemonCommand.PersistentFlags().String(f.Service.Chart.WebhookAuthToken, "", "The auth token value for requests to the webhook.")
```

Small change. WDYT?

server/endpoint/status/endpoint.go
```diff
@@ -99,6 +109,12 @@ func (e Endpoint) Endpoint() kitendpoint.Endpoint {
...
+			if request.AuthToken != e.webhookAuthToken {
+				return nil, microerror.Maskf(wrongTokenError, "got wrong auth token %q for %#q app %#q namespace", request.AuthToken, request.AppName, request.AppNamespace)
```
rossf7, line 114:
```suggestion
				return nil, microerror.Maskf(wrongTokenError, "incorrect auth token for %#q app %#q in namespace", request.AppName, request.AppNamespace)
```

I'm being paranoid as usual :smile: but I'd prefer not to log the token. If there is a misconfiguration then we need to reset the token as it can be in logs.

I think this is still clear and OK for debugging. WDYT?

service/controller/app/resource/authtoken/create.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToApp(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if key.InCluster(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q in %#q uses InCluster kubeconfig no need to install webhook auth token", cr.Name, cr.Namespace))
```
rossf7, line 27:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("app %#q in %#q uses InCluster kubeconfig no need for webhook auth token", cr.Name, cr.Namespace))
```

Just because it's shorter.

service/controller/app/resource/authtoken/create.go
rossf7: Good we need this. Not for now but I think maybe we should drop the logs for non chart-operator CRs in the resources.

They add a lot of noise but don't add much value IMO.

service/controller/app/resource/authtoken/create.go
```diff
@@ -0,0 +1,104 @@
...
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+		return nil
+	}
+
+	desiredSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      authTokenName,
+			Namespace: namespace,
+			Annotations: map[string]string{
+				annotation.Notes: fmt.Sprintf("DO NOT EDIT. Values managed by %s.", project.Name()),
```
rossf7, line 61: Nice. 👍

rossf7 (commented):
This is nice. Just a few comments and minor things.

main.go
tomahawk28: Much better, Thanks!

server/endpoint/status/endpoint.go
tomahawk28: You don't want to print the wrong token, too? I see It's not giving any value I think.

service/controller/app/resource/authtoken/create.go
tomahawk28: Agreed, I can jump into that tomorrow morning since we are in still the cool-down phase.

rossf7: Thanks that's a good cool-down task.

rossf7 (approved):
LGTM 🚀

PR #576, opened 2020-11-09
Use annotation package from app library

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/11656

This is prep for using the validation package from the app library. This uses the extracted annotation package from app.

Next PR will use the extracted key package.


rossf7:
Tested on `ginger`.

tomahawk28 (approved):

PR #581, opened 2020-11-09
Use key package from app library

rossf7:
This is prep for also using the validation package from the app library.


rossf7:
Tested on `ginger`.

tomahawk28 (approved):

PR #605, opened 2020-12-10
Use pause annotation to stop reconciling App CRs

kubasobon:
Towards https://github.com/giantswarm/giantswarm/issues/14821

## Checklist


main.go
```diff
@@ -146,6 +146,7 @@ func mainWithError() (err error) {
 
 	daemonCommand := newCommand.DaemonCommand().CobraCommand()
 
+	daemonCommand.PersistentFlags().String(f.Service.App.PauseAnnotation, "", "Annotation used to pause App CR reoncilliation.")
```
tomahawk28, line 149:
```suggestion
	daemonCommand.PersistentFlags().String(f.Service.App.PauseAnnotations, "", "Annotations used to pause App CR reoncilliation.")
```

Looks like it's apparently plural, right?

integration/setup/setup.go
```diff
@@ -115,6 +115,11 @@ func installResources(ctx context.Context, config Config) error {
 					"Registry": map[string]interface{}{
 						"Domain": "quay.io",
 					},
+					"GiantSwarm": map[string]interface{}{
+						"AppOperator": map[string]interface{}{
+							"PauseAnnotation": "app-operator.giantswarm.io/paused",
```
kopiczko, line 120: The name looks so generic now that I'm thinking we can put it as a default in `values.yaml` and skip `installations`. I think this pause annotation is a sane feature to have and having it defined in `values.yaml` would make it easily discoverable for external the user.

kopiczko (approved):

main.go
kubasobon: I wanted to have just 1 annotation (excluding `operatorkit` defaults) which can pause app-operator, hence singular.

If you think it would be beneficial to extend this and allow for multiple annotations, let me know. That would be a separate PR though.

tomahawk28 (approved):
LGTM if it has been tested!

main.go
JosephSalisbury, line 149:
```suggestion
	daemonCommand.PersistentFlags().String(f.Service.App.PauseAnnotation, "", "Annotation used to pause App CR reconciliation.")
```

rossf7 (approved):
LGTM

kopiczko:
As discussed [here](https://github.com/giantswarm/giantswarm/issues/14821#issuecomment-748000367) we'll use baked-in `app-operator.giantswarm.io/paused` instead.

PR #612, opened 2020-12-14
Reuse clients in clients resource when app CR uses inCluster

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/14755

This is to help with the performance problems we're seeing in ghost due to a high volume of requests to the K8s API server.

Currently we create a new k8sclient per app CR. This includes the dynamic API which results in requests to the discovery API.

For in cluster app CRs we can just reuse the existing clients. For tenant app CRs we can use the controller-runtime client. I'll do that in the next PR.

## Checklist

### DIFF
--- a/service/controller/app/resource/clients/resource.go
+++ b/service/controller/app/resource/clients/resource.go
@@ -13,7 +13,6 @@ import (
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	"github.com/spf13/afero"
-	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
 	"github.com/giantswarm/app-operator/v2/service/controller/app/controllercontext"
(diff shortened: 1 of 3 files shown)


service/controller/app/resource/clients/resource.go
```diff
@@ -91,10 +96,20 @@ func (r *Resource) addClientsToContext(ctx context.Context, cr v1alpha1.App) err
 		return nil
 	}
 
+	// App CR uses inCluster so we can reuse the existing clients.
+	if key.InCluster(cr) {
+		cc.Clients = controllercontext.Clients{
+			K8s:  r.k8sClient,
+			Helm: r.helmClient,
+		}
```
rossf7, line 104: If we're using inCluster we can just reuse the existing clients.

rossf7:
Tested on `godsmack`.

tomahawk28 (approved):
Perfect! Thanks!

PR #625, opened 2021-01-04
Use pause annotation to cancel reconciliation

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/14821


kopiczko:
Tested on gollum

go.mod
```diff
@@ -6,7 +6,8 @@ require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/ghodss/yaml v1.0.0
-	github.com/giantswarm/apiextensions/v3 v3.13.0
+	github.com/giantswarm/apiextensions/v3 v3.13.1-0.20210104133707-4fe232b93488
```
tomahawk28, line 9: I think this should be reverted?

rossf7 (approved):
LGTM once apiextensions is released

service/controller/app/resource/pause/resource.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if len(cr.GetAnnotations()) == 0 {
+		return nil
+	}
+
+	v := cr.GetAnnotations()[annotation.AppOperatorPaused]
+	if v == "true" {
```
tomahawk28, line 54: if the annotation has a key named `annotation.AppOperatorPaused`, do we still need to see the value whether it is `true` or not?

go.sum
```diff
@@ -34,6 +34,7 @@ github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
 github.com/CloudyKit/fastprinter v0.0.0-20170127035650-74b38d55f37a/go.mod h1:EFZQ978U7x8IRnstaskI3IysnWY5Ao3QgZUKOXlsAdw=
 github.com/CloudyKit/jet v2.1.3-0.20180809161101-62edd43e4f88+incompatible/go.mod h1:HPYO+50pSWkPoj9Q/eq0aRGByCL6ScRlUmiEX5Zgm+w=
+github.com/DATA-DOG/go-sqlmock v1.4.1/go.mod h1:f/Ixk793poVmq4qj/V1dPUg2JEAKC73Q5eFN3EC/SaM=
```
tomahawk28, line 37: Why suddenly so many go modules updated?

kopiczko: `go mod tidy` in the CI doesn't complain. I don't know why is that. Maybe something in apiextensions changed in the meantime.

go.mod
kopiczko: I was testing before releasing apiextensions.

service/controller/app/resource/pause/resource.go
kopiczko: This is how it's actually done in cluster-api: https://github.com/kubernetes-sigs/cluster-api/blob/dadd7bd6233edb1e422e7af282e923abff52f830/util/annotations/helpers.go#L35 so let's do it this way.

kopiczko:
Replaced with https://github.com/giantswarm/app-operator/pull/626

PR #635, opened 2021-01-13
Use key funcs in integration tests

rossf7:
This got a bit big but it makes improvements to the integration tests.

Prior to  renaming the kubeconfig test to workloadcluster and using it to test the app-operator per cluster changes.

- Replaces constants with key funcs.
- Installs app-operator and chart-operator as unique so the status webhook works.
- Uses chart-operator 2.5.1 which is the latest release without VPA.

Next PR will use apptest library to create the app CRs and will release chart-operator with the VPA fix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  architect: giantswarm/architect@0.10.1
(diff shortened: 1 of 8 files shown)


tomahawk28 (approved):
<3

PR #636, opened 2021-01-13
Add support for VPA

rossf7:
## Checklist


rossf7:
Need to test with per cluster app-operator to find reduced requests and limits.

Also need to set a memory limit for app-operator-unique. Will come back to this.

PR #640, opened 2021-01-14
Watch current namespace unless its the unique instance

rossf7:

## Checklist

### DIFF
--- a/helm/app-operator/templates/deployment.yaml
+++ b/helm/app-operator/templates/deployment.yaml
@@ -44,6 +44,11 @@ spec:
         - daemon
         - --config.dirs=/var/run/{{ include "name" . }}/configmap/
         - --config.files=config
+        env:
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
         livenessProbe:
           httpGet:
             path: /healthz
--- a/pkg/label/label.go
+++ b/pkg/label/label.go
@@ -21,24 +21,18 @@ const (
 )
 
 func AppVersionSelector(unique bool) labels.Selector {
-	version := GetProjectVersion(unique)
-	s := fmt.Sprintf("%s=%s", label.AppOperatorVersion, version)
+	var selector string
 
-	selector, err := labels.Parse(s)
+	if unique {
+		selector = fmt.Sprintf("%s=%s", label.AppOperatorVersion, project.ManagementClusterAppVersion())
+	} else {
+		selector = fmt.Sprintf("%s!=%s", label.AppOperatorVersion, project.ManagementClusterAppVersion())
+	}
+
+	s, err := labels.Parse(selector)
 	if err != nil {
 		panic(fmt.Sprintf("failed to parse selector %#q with error %#q", s, err))
 	}
 
-	return selector
-}
-
-func GetProjectVersion(unique bool) string {
-	if unique {
-		// When app-operator is deployed as a unique app it only processes
-		// management cluster app CRs. These CRs always have the version label
-		// app-operator.giantswarm.io/version: 0.0.0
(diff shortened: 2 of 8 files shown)


helm/app-operator/templates/deployment.yaml
```diff
@@ -44,6 +44,11 @@ spec:
         - daemon
         - --config.dirs=/var/run/{{ include "name" . }}/configmap/
         - --config.files=config
+        env:
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
```
rossf7, line 51: This env var is used to get the namespace using the Downward API.

https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables

pkg/label/label.go
```diff
@@ -20,25 +20,25 @@ const (
...
 func AppVersionSelector(unique bool) labels.Selector {
-	version := GetProjectVersion(unique)
-	s := fmt.Sprintf("%s=%s", label.AppOperatorVersion, version)
+	var selector string
 
-	selector, err := labels.Parse(s)
+	if unique {
+		// Unique instance watches all namespaces for app CRs with the unique
+		// app version (0.0.0).
+		selector = fmt.Sprintf("%s=%s", label.AppOperatorVersion, project.ManagementClusterAppVersion())
```
rossf7, line 31: app-operator-unique works the same as now.

pkg/label/label.go
```diff
@@ -20,25 +20,25 @@ const (
...
+	} else {
+		// Other instances watch the namespace they are running in but exclude
+		// unique app CRs.
+		selector = fmt.Sprintf("%s!=%s", label.AppOperatorVersion, project.ManagementClusterAppVersion())
```
rossf7, line 35: For workload clusters the namespace will include app-operator's own app CR. 

We exclude that using the selector because its a unique app CR.

pkg/project/project.go
```diff
@@ -33,9 +33,3 @@ func Source() string {
 func Version() string {
 	return version
 }
-
-// WorkloadClusterAppVersion is always 1.0.0 for workload cluster app CRs using Helm 2.
-// For app CRs using Helm 3 we use project.Version().
-func WorkloadClusterAppVersion() string {
-	return "1.0.0"
-}
```
rossf7, line 41: Not needed anymore. 🎉

service/controller/app/app.go
```diff
@@ -113,6 +117,12 @@ func NewApp(config Config) (*App, error) {
 			Name: "app",
 		}
 
+		if !config.UniqueApp {
+			// Only watch app CRs in the current namespace. The label selector
+			// excludes the operator's own app CR which has the unique version.
+			c.Namespace = config.PodNamespace
```
rossf7, line 123: Sets the watch namespace for the app controller.

For the unique instance we continue to watch all namespaces.

service/watcher/appvalue.go
```diff
@@ -36,10 +38,23 @@ func NewAppValueWatcher(config AppValueWatcherConfig) (*AppValueWatcher, error)
...
+	if config.PodNamespace == "" {
+		return nil, microerror.Maskf(invalidConfigError, "%T.PodNamespace must not be empty", config)
+	}
+
+	var appNamespace string
+
+	if !config.UniqueApp {
+		// Only watch app CRs in the current namespace. The label selector
+		// excludes the operator's own app CR which has the unique version.
+		appNamespace = config.PodNamespace
```
rossf7, line 51: Same change for the watcher which is used when building the cache.

rossf7:
Will come back to this.

PR #643, opened 2021-01-15
Tidy up service wiring

rossf7:
Just some minor stuff. Prep for https://github.com/giantswarm/app-operator/pull/640/


rossf7 (commented):
Don't think this is controversial. Will go with a ping.

PR #647, opened 2021-01-26
Use apptest in integration tests

rossf7:

Updates the integration tests to use the apptest library for creating app CRs. This simplifies the test code a lot and is prep for extending the tests for running an app-operator per workload cluster.

### DIFF
--- a/integration/setup/setup.go
+++ b/integration/setup/setup.go
@@ -121,8 +121,8 @@ func installResources(ctx context.Context, config Config) error {
 		// Release is named app-operator-unique as some functionality is only
 		// implemented for the unique instance.
 		opts := helmclient.InstallOptions{
-			ReleaseName:   key.AppOperatorUniqueName(),
-			WaitForDeploy: true,
+			ReleaseName: key.AppOperatorUniqueName(),
+			Wait:        true,
 		}
 		err = config.HelmClient.InstallReleaseFromTarball(ctx,
 			operatorTarballPath,
--- a/integration/setup/setup.go
+++ b/integration/setup/setup.go
@@ -121,7 +121,8 @@ func installResources(ctx context.Context, config Config) error {
 		// Release is named app-operator-unique as some functionality is only
 		// implemented for the unique instance.
 		opts := helmclient.InstallOptions{
-			ReleaseName: key.AppOperatorUniqueName(),
+			ReleaseName:   key.AppOperatorUniqueName(),
+			WaitForDeploy: true,
 		}
 		err = config.HelmClient.InstallReleaseFromTarball(ctx,
 			operatorTarballPath,
--- a/integration/test/app/basic/basic_test.go
+++ b/integration/test/app/basic/basic_test.go
@@ -7,10 +7,10 @@ import (
 	"testing"
 	"time"
 
-	"github.com/giantswarm/apiextensions/v3/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/apiextensions/v3/pkg/crd"
 	"github.com/giantswarm/apiextensions/v3/pkg/label"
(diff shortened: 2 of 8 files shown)


integration/setup/setup.go
```diff
@@ -122,6 +122,7 @@ func installResources(ctx context.Context, config Config) error {
 		// implemented for the unique instance.
 		opts := helmclient.InstallOptions{
 			ReleaseName: key.AppOperatorUniqueName(),
+			Wait:        true,
```
rossf7, line 125: The tests were failing due to the status webhook because coredns was not ready yet. 

Waiting for both operators to be ready makes the tests more reliable.

Maybe we should add wait support to app CRs? As it could be useful in some situations. I'll add a note to our board to discuss.

integration/test/app/basic/basic_test.go
```diff
@@ -39,10 +39,22 @@ func TestAppLifecycle(t *testing.T) {
...
+		{
+			crdName := "Chart"
+			config.Logger.Debugf(ctx, "ensuring %#q CRD exists", crdName)
+
+			err := config.K8sClients.CRDClient().EnsureCreated(ctx, crd.LoadV1("application.giantswarm.io", crdName), backoff.NewMaxRetries(7, 1*time.Second))
+			if err != nil {
+				t.Fatalf("expected %#v got %#v", nil, err)
+			}
+
+			config.Logger.Debugf(ctx, "ensured %#q CRD exists", crdName)
```
rossf7, line 51: Creating the chart CRD needs to happen first or the readiness probe in chart-operator will fail.

integration/test/app/basic/basic_test.go
```diff
@@ -72,92 +84,31 @@ func TestAppLifecycle(t *testing.T) {
...
-				Namespace: key.Namespace(),
-				Version:   "0.1.0",
+		apps := []apptest.App{
+			{
+				// Install test app.
+				CatalogName:   key.DefaultCatalogName(),
+				Name:          key.TestAppName(),
+				Namespace:     key.Namespace(),
+				Version:       "0.1.0",
+				WaitForDeploy: true,
```
rossf7, line 105: This is the main change for the basic test.

integration/test/app/kubeconfig/kubeconfig_test.go
```diff
@@ -59,173 +56,39 @@ func TestAppWithKubeconfig(t *testing.T) {
...
+		apps := []apptest.App{
+			{
+				// Bootstrap chart-operator in the giantswarm namespace.
+				CatalogName:   key.DefaultCatalogName(),
+				KubeConfig:    kubeConfig,
+				Name:          key.ChartOperatorName(),
+				Namespace:     key.Namespace(),
+				ValuesYAML:    templates.ChartOperatorValues,
+				Version:       key.ChartOperatorVersion(),
+				WaitForDeploy: true,
```
rossf7, line 77: For the kubeconfig test we also create a chart-operator app CR to test the bootstrap logic.

rossf7:
Ping @tomahawk28 not super urgent but PTAL when you get time.

integration/setup/setup.go
tomahawk28: How did you fix coredns issue with helmclient's `wait` option? It seems fine for me but just don't get how it fixed this issue.

tomahawk28 (approved):
Just one comment :) 👍

integration/setup/setup.go
rossf7: There was a race condition because the test was running before coredns was fully up.

In the chart-operator logs it was failing to call the webhook URL. When I checked the coredns logs its because it wasn't up yet. Waiting for the operator pods to be ready adds some delay but makes the test more reliable.

PR #650, opened 2021-01-28
Update additional metadata fields 

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/14802

## Checklist


go.mod
```diff
@@ -6,10 +6,10 @@ require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/ghodss/yaml v1.0.0
-	github.com/giantswarm/apiextensions/v3 v3.15.0
+	github.com/giantswarm/apiextensions/v3 v3.15.2-0.20210127184902-85b45fef7268
```
tomahawk28, line 9: TODO: update to 3.16.0 before merge

service/controller/appcatalog/resource/appcatalogentry/create.go
```diff
@@ -140,34 +139,25 @@ func (r *Resource) newAppCatalogEntries(ctx context.Context, cr v1alpha1.AppCata
...
-					restrictions, err = parseRestrictions(metadata)
+				if rawMetadata != nil {
+					metadata, err = parseMetadata(rawMetadata)
 					if err != nil {
 						return nil, microerror.Mask(err)
 					}
 				}
 			}
 
-			createdTime, err := parseTime(entry.Created)
```
tomahawk28, line 162: Why not utilize `dateCreated` from metadata files directly? No need to use the entry date from the catalog.

tomahawk28: Skipping this arguing for simplicity.

tomahawk28: As we discussed in the team channel, resource would fall back to the date in the entry if no `dateCreated` is provided by meta.yaml.

service/controller/appcatalog/resource/appcatalogentry/create.go
```diff
@@ -230,28 +246,25 @@ func equals(current, desired *v1alpha1.AppCatalogEntry) bool {
...
 	}
 
 	// Using reflect.DeepEqual doesn't work for the 2 date fields due to time
 	// zones. Instead we compare the unix epoch and clear the date fields.
-	if current.Spec.DateCreated.Unix() != desired.Spec.DateCreated.Unix() {
-		return false
-	}
-	current.Spec.DateCreated = nil
-	desired.Spec.DateCreated = nil
+	timeComparer := cmp.Comparer(func(current, desired *metav1.Time) bool {
```
tomahawk28, line 259: Using `Comparer` is more elegant than reflect since you can compare by the type matching.

tomahawk28:
Tested in `ginger`

service/controller/appcatalog/resource/appcatalogentry/create.go
```diff
@@ -212,13 +212,29 @@ func (r *Resource) newAppCatalogEntries(ctx context.Context, cr v1alpha1.AppCata
...
+				entryCR.Spec.DateUpdated = m.DataCreated
+			}
+
+			if entryCR.Spec.Chart.APIVersion == "" {
+				// chartAPIVersion is default to `v1`
+				entryCR.Spec.Chart.APIVersion = "v1"
+			}
+
+			if entryCR.Spec.DateCreated == nil {
+				// If meta.yaml does not have dateCreated, use the timestamp from app
```
rossf7, line 233:
```suggestion
				// If meta.yaml does not have dateCreated, use the timestamp from app.
```

service/controller/appcatalog/resource/appcatalogentry/create.go
rossf7: Nice TIL about Comparer. But the comment also needs updating now we don't use the unix epoch.

service/controller/appcatalog/resource/appcatalogentry/types.go
```diff
@@ -18,5 +21,8 @@ type entry struct {
 }
 
 type metadata struct {
-	Restrictions v1alpha1.AppCatalogEntrySpecRestrictions `json:"restrictions"`
+	Annotations     map[string]string                         `json:"annotations"`
+	ChartAPIVersion string                                    `json:"chartApiVersion"`
```
rossf7, line 25:
```suggestion
	ChartAPIVersion string                                    `json:"chartAPIVersion"`
```

These abbreviations should be consistent and API is the correct casing IMO. But this is the casing in the metadata file isn't it?

If so lets go with this but I'll do a PR to the spec to fix the casing. The metadata files are still new so I'd like to get this right.

rossf7: Another idea. Should the struct be appMetadata instead of metadata? 

No need to export it but I got a bit confused earlier even though I think I chose the name. 😄  #NamingIsHard

rossf7 (commented):
Direction is nice. Just some minor stuff.

service/controller/appcatalog/resource/appcatalogentry/create.go
tomahawk28: I think we are still using unix epoch like before. right? 
In line number 261,
```
return current.Unix() == desired.Unix()
```

service/controller/appcatalog/resource/appcatalogentry/types.go
tomahawk28: > But this is the casing in the metadata file isn't it? 
This is from metadata actually, so yes, we should fix on spec as you said. Thanks for noticing!

> Should the struct be appMetadata instead of metadata?

You're right! Since we have `metadata` from upstream as well, it's better to change the name.

service/controller/appcatalog/resource/appcatalogentry/create.go
rossf7: Long week. 😂 Thanks for clarifying.

service/controller/appcatalog/resource/appcatalogentry/types.go
rossf7: Cool I'll do a spec PR. Thanks for renaming!

rossf7 (approved):
LGTM 🚀

rossf7:
@tomahawk28 I'll leave the dependabot PRs till you've merged this.

tomahawk28:
I don't understand why `appcatalogentry` testing is failing even though I changed almost nothing but revert apiextensions version to 3.16.1. I will take a further look next Monday. Thanks!

tomahawk28:
Strange, integration testing is fetching totally wrong version of app-operator, I checked this by ssh into circleCI env. 
```
circleci@default-12b803a1-faa5-4f2f-908a-98360294915b:~$ kubectl get deploy -n giantswarm -o wide
NAME                  READY   UP-TO-DATE   AVAILABLE   AGE   CONTAINERS     IMAGES                                                                           SELECTOR
app-operator-unique   1/1     1            1           90s   app-operator   quay.io/giantswarm/app-operator:3.1.0-daeded242d150cb2f72417841facd9d55db31f3c   app.kubernetes.io/instance=app-operator-unique,app.kubernetes.io/name=app-operator
```

tomahawk28:
I would merge this and fix the integration test on another PR.

PR #658, opened 2021-02-02
Enabling appcatalogentry to all catalog

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/14801

## Checklist

### DIFF
--- a/service/controller/appcatalog/resource/appcatalogentry/create.go
+++ b/service/controller/appcatalog/resource/appcatalogentry/create.go
@@ -169,10 +169,12 @@ func (r *Resource) newAppCatalogEntries(ctx context.Context, cr v1alpha1.AppCata
 
 			var rawMetadata []byte
 			{
-				rawMetadata, err = r.getMetadata(ctx, key.AppCatalogStorageURL(cr), entry.Name, entry.Version)
-				if err != nil {
-					r.logger.LogCtx(ctx, "level", "info", "message", fmt.Sprintf("failed to get appMetadata for entry %#q in catalog %#q", entry.Name, cr.Name), "stack", fmt.Sprintf("%#v", err))
-					continue
+				if url, ok := entry.Annotations[metadataAnnotation]; !ok {
+					rawMetadata, err = r.getMetadata(ctx, url, entry.Name, entry.Version)
+					if err != nil {
+						r.logger.LogCtx(ctx, "level", "info", "message", fmt.Sprintf("failed to get appMetadata for entry %#q in catalog %#q", entry.Name, cr.Name), "stack", fmt.Sprintf("%#v", err))
+						continue
+					}
 				}
 			}
 
--- a/service/controller/appcatalog/resource/appcatalogentry/create.go
+++ b/service/controller/appcatalog/resource/appcatalogentry/create.go
@@ -17,6 +17,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/giantswarm/app-operator/v3/pkg/annotation"
 	pkglabel "github.com/giantswarm/app-operator/v3/pkg/label"
 	"github.com/giantswarm/app-operator/v3/pkg/project"
 )
@@ -169,7 +170,7 @@ func (r *Resource) newAppCatalogEntries(ctx context.Context, cr v1alpha1.AppCata
 
 			var rawMetadata []byte
 			{
-				if url, ok := entry.Annotations[metadataAnnotation]; ok {
+				if url, ok := entry.Annotations[annotation.Metadata]; ok {
 					rawMetadata, err = r.getMetadata(ctx, url)
 					if err != nil {
 						r.logger.LogCtx(ctx, "level", "info", "message", fmt.Sprintf("failed to get appMetadata for entry %#q in catalog %#q", entry.Name, cr.Name), "stack", fmt.Sprintf("%#v", err))
--- a/service/controller/appcatalog/resource/appcatalogentry/create.go
+++ b/service/controller/appcatalog/resource/appcatalogentry/create.go
@@ -34,6 +34,13 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	// Skip creating appcatalogentry CRs if this is a community catalog.
+	if key.AppCatalogType(cr) == communityCatalogType {
+		r.logger.Debugf(ctx, "not creating CRs for catalog %#q with type %#q", cr.Name, communityCatalogType)
+		r.logger.Debugf(ctx, "canceling resource")
+		return nil
+	}
+
 	currentEntryCRs, err := r.getCurrentEntryCRs(ctx, cr)
 	if err != nil {
 		return microerror.Mask(err)
--- a/service/controller/appcatalog/resource/appcatalogentry/create.go
+++ b/service/controller/appcatalog/resource/appcatalogentry/create.go
@@ -169,8 +169,8 @@ func (r *Resource) newAppCatalogEntries(ctx context.Context, cr v1alpha1.AppCata
 
 			var rawMetadata []byte
 			{
-				if url, ok := entry.Annotations[metadataAnnotation]; !ok {
-					rawMetadata, err = r.getMetadata(ctx, url, entry.Name, entry.Version)
+				if url, ok := entry.Annotations[metadataAnnotation]; ok {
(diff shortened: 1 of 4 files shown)


tomahawk28:
NVM, I click wrong button and not it's PR!

tomahawk28:
```
releases-releases-kvm-1.0.0                                                                                    releases                     releases-kvm                         master                                            1.0.0                                                   305d
releases-test-releases-aws-0.0.0-ff7f7b9ec3deb622477d32d3d89b1c130d894993                                      releases-test                releases-aws                         master                                            0.0.0-ff7f7b9ec3deb622477d32d3d89b1c130d894993          4d1h
```

@rossf7 @piontec Do we need to keep `release` and `release-test` catalog for appcatalog entry?

tomahawk28:
Tested on `ginger` 
```
kubectl get appcatalogentry | wc -l
    1345
```

service/controller/appcatalog/resource/appcatalogentry/create.go
```diff
@@ -34,19 +34,6 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
 	}
 
-	// Skip creating appcatalogentry CRs if the catalog is not public.
-	if key.AppCatalogVisibility(cr) != publicVisibilityType {
-		r.logger.Debugf(ctx, "not creating CRs for catalog %#q with visibility %#q", cr.Name, key.AppCatalogVisibility(cr))
-		r.logger.Debugf(ctx, "canceling resource")
-		return nil
-	}
-	// Skip creating appcatalogentry CRs if this is a community catalog.
-	if key.AppCatalogType(cr) == communityCatalogType {
```
rossf7, line 44: I think we still need this. We don't want to create entries for the Helm stable catalog as its deprecated.

rossf7:
> Do we need to keep release and release-test catalog for appcatalog entry?

I don't think we care much about these catalogs but if we're generating entries for catalogs we might as well include them.

Code wise this looks fine but I think we should exclude Helm stable. My concerns are around performance and flooding the logs. 

- How many HTTP requests does it make for the metadata files and how often? Does that work in China?
- I can't remember what gets logged. Is it per app or per appcatalogentry?

I think it should log as either per app or even per catalog level to reduce noise and volume of logs.

service/controller/appcatalog/resource/appcatalogentry/create.go
tomahawk28: Even though we don't have any generated `AppCatalogEntry` belongs to helm-stable, yes, you're right. I will revert this back.

tomahawk28:
>How many HTTP requests does it make for the metadata files and how often? Does that work in China?

There are 400 HTTP pulling for metadata.yaml every 15 minutes. Most of them are returning 404. 

image

I think its the number of HTTP call would grow exponentially, so we need to reduce them.

What do you think about below option? > @piontec 

1. Adding metadata's sha1 to index.yaml in appcatalog.
2. Cachig appcatalog's index.yaml and only pulling when its hash is different. 

In this case, we will be pulling index.yaml every 5 minutes, but we will pull the meta.yaml only if we found a difference.

tomahawk28:
More about above: I see `annotations` in index.yaml contains `application.giantswarm.io/metadata` (e.g. https://giantswarm.github.io/giantswarm-catalog/loki-0.1.0-alpha.tgz-meta/main.yaml) 
We don't have that many app metadata available from index.yaml, so I will pull `meta.yaml` only if app presents `application.giantswarm.io/metadata`. We don't need the cache feature for now.

tomahawk28:
> I can't remember what gets logged. Is it per app or per appcatalogentry?

It's per appcatalogentry but its HTTP log is too much like below. 
```
D 02/02 15:56:19 /apis/application.giantswarm.io/v1alpha1/appcatalogs/giantswarm-playground appcatalogentry getting main.yaml from `https://giantswarm.github.io/giantswarm-playground-catalog/k8s-audit-metrics-0.0.3.tgz-meta/main.yaml` | app-operator/v3/service/controller/appcatalog/resource/appcatalogentry/resource.go:137 | controller=appCatalog | event=update | loop=735 | version=182847233
D 02/02 15:56:19 /apis/application.giantswarm.io/v1alpha1/appcatalogs/giantswarm-playground appcatalogentry no main.yaml generated at `https://giantswarm.github.io/giantswarm-playground-catalog/k8s-audit-metrics-0.0.3.tgz-meta/main.yaml` | app-operator/v3/service/controller/appcatalog/resource/appcatalogentry/resource.go:147 | controller=appCatalog | event=update | loop=735 | version=182847233
D 02/02 15:56:19 /apis/application.giantswarm.io/v1alpha1/appcatalogs/giantswarm-playground appcatalogentry getting main.yaml from `https://giantswarm.github.io/giantswarm-playground-catalog/k8s-initiator-app-0.9.1.tgz-meta/main.yaml` | app-operator/v3/service/controller/appcatalog/resource/appcatalogentry/resource.go:137 | controller=appCatalog | event=update | loop=735 | version=182847233
D 02/02 15:56:19 /apis/application.giantswarm.io/v1alpha1/appcatalogs/giantswarm-playground appcatalogentry no main.yaml generated at `https://giantswarm.github.io/giantswarm-playground-catalog/k8s-initiator-app-0.9.1.tgz-meta/main.yaml` | app-operator/v3/service/controller/appcatalog/resource/appcatalogentry/resource.go:147 | controller=appCatalog | event=update | loop=735 | version=182847233```

We can reduce the noise if we pull the `meta.yaml` only if we see this in the annotation as I suggested.

File: service/controller/appcatalog/resource/appcatalogentry/resource.go [hunk 2]
```diff
@@ -23,10 +23,12 @@ import (
...
-	apiVersion          = "application.giantswarm.io/v1alpha1"
-	kindAppCatalog      = "AppCatalog"
-	kindAppCatalogEntry = "AppCatalogEntry"
-	maxEntriesPerApp    = 5
+	apiVersion           = "application.giantswarm.io/v1alpha1"
+	communityCatalogType = "community"
+	kindAppCatalog       = "AppCatalog"
+	kindAppCatalogEntry  = "AppCatalogEntry"
+	maxEntriesPerApp     = 5
+	metadataAnnotation   = "application.giantswarm.io/metadata"
```
rossf7, line 31: Let's have this in pkg/annotation or apiextensions. There are discussions in sig-operator about having a metadata library instead of using apiextensions. So I think pkg is fine for now.

service/controller/appcatalog/resource/appcatalogentry/resource.go
```diff
@@ -124,13 +126,13 @@ func (r *Resource) getIndex(ctx context.Context, storageURL string) (index, erro
...
 
-func (r *Resource) getMetadata(ctx context.Context, storageURL, name, version string) ([]byte, error) {
+func (r *Resource) getMetadata(ctx context.Context, mainURL, name, version string) ([]byte, error) {
 	eventName := "pull_metadata_file"
 
 	t := prometheus.NewTimer(histogram.WithLabelValues(eventName))
 	defer t.ObserveDuration()
 
-	mainURL := fmt.Sprintf("%s/%s-%s.tgz-meta/main.yaml", strings.TrimRight(storageURL, "/"), name, version)
+	//mainURL := fmt.Sprintf("%s/%s-%s.tgz-meta/main.yaml", strings.TrimRight(storageURL, "/"), name, version)
```
rossf7, line 135: Should this be removed?

rossf7 (commented):
> More about above: I see annotations in index.yaml contains application.giantswarm.io/metadata (e.g. https://giantswarm.github.io/giantswarm-catalog/loki-0.1.0-alpha.tgz-meta/main.yaml)
> We don't have that many app metadata available from index.yaml, so I will pull meta.yaml only if app presents application.giantswarm.io/metadata. We don't need the cache feature for now.

Checking for the metadata annotation works for now but we should get the caching sorted out. As we'll need it at some point.

I think a 2nd annotation with the SHA is a good idea. We can use that to reduce both the number of HTTP requests and reduce noise in the logs. For this use case I think we should only log if we're adding or updating an appcatalogentry CR.

@piontec Let us know what you think about adding a digest to the metadata files. We can also do a quick call tomorrow if that helps.

service/controller/appcatalog/resource/appcatalogentry/resource.go
tomahawk28: Good catch, I will move this into pkg.

rossf7 (approved):
LGTM

Thanks this is cool. Let's go with this but not release the changes until we've discussed caching.

tomahawk28:
I will merge it for now, but would not release it until we clear about caching.

PR #662, opened 2021-02-03
Adding cordoned status into context

tomahawk28:
From the recent slack channel discussion in #area-managedapps, we found out cordoned apps are not updating their status accordingly. https://gigantic.slack.com/archives/CRZN9GLJW/p1612368461138500

This is because we delete `addStatusToContext` logic from configmap resource, which will propagate status update in the next reconciliation. 

## Checklist


rossf7 (approved):
Good catch, LGTM

PR #664, opened 2021-02-04
Rename watcher to appvalue watcher

rossf7:
Renames the current watcher package to appvalue. This is prep for adding the 2nd chartstatus watcher.


rossf7:
Tested on `gauss`.

tomahawk28 (approved):
LGTM

PR #644, opened 2021-01-18
WIP: v4-dev: Make app-operator instances per workload cluster

rossf7:
## Checklist

- [ ] Update changelog in CHANGELOG.md.

PR #668, opened 2021-02-05
For non-unique instance only watch the current namespace

rossf7:

Most recent test with cluster-operator was successful. 🎉

So I'm going to start breaking down my big ball of mud in https://github.com/giantswarm/app-operator/pull/644 into smaller PRs to merge into the v4-dev branch. Shouldn't take too many. The next PR will have the new chart CR watcher.

### DIFF
--- a/helm/app-operator/templates/deployment.yaml
+++ b/helm/app-operator/templates/deployment.yaml
@@ -44,6 +44,11 @@ spec:
         - daemon
         - --config.dirs=/var/run/{{ include "name" . }}/configmap/
         - --config.files=config
+        env:
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
         livenessProbe:
           httpGet:
             path: /healthz
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -1,0 +1,27 @@
+package env
+
+import (
+	"fmt"
+	"os"
+)
+
+const (
+	// EnvVarPodNamespace is the process environment variable representing the
+	// POD_NAMESPACE env var which is populated via the Downward API.
+	EnvVarPodNamespace = "POD_NAMESPACE"
+)
+
+var (
+	podNamespace string
+)
+
+func init() {
+	podNamespace = os.Getenv(EnvVarPodNamespace)
+	if podNamespace == "" {
+		panic(fmt.Sprintf("env var '%s' must not be empty", EnvVarPodNamespace))
+	}
+}
+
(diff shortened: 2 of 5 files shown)


helm/app-operator/templates/deployment.yaml
```diff
@@ -44,6 +44,11 @@ spec:
         - daemon
         - --config.dirs=/var/run/{{ include "name" . }}/configmap/
         - --config.files=config
+        env:
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
```
rossf7, line 51: Uses the Downward API to get the namespace the operator is running in. 

https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#capabilities-of-the-downward-api

pkg/env/env.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"fmt"
+	"os"
+)
+
+const (
+	// EnvVarPodNamespace is the process environment variable representing the
+	// POD_NAMESPACE env var which is populated via the Downward API.
+	EnvVarPodNamespace = "POD_NAMESPACE"
```
rossf7, line 11: I'm reusing the approach we use in integration tests to get the env var.

service/controller/app/app.go
```diff
@@ -116,6 +120,12 @@ func NewApp(config Config) (*App, error) {
 			Name: project.Name() + appControllerSuffix,
 		}
 
+		if !config.UniqueApp {
+			// Only watch app CRs in the current namespace. The label selector
+			// excludes the operator's own app CR which has the unique version.
+			c.Namespace = config.PodNamespace
```
rossf7, line 126: This is the main change. For the unique instance we continue to watch all namespaces.

pkg/env/env.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+var (
+	podNamespace string
+)
+
+func init() {
+	podNamespace = os.Getenv(EnvVarPodNamespace)
+	if podNamespace == "" {
```
tomahawk28, line 20: Is this going to be `giantswarm` in the unique app, right?

tomahawk28 (approved):
LGTM, just one question.

pkg/env/env.go
rossf7: Yes it will be giantswarm but because the unique flag is true we don't set the namespace in the operatorkit controller. 

This is so we can process the new app-operator app CR in the cluster namespace.

PR #673, opened 2021-02-08
Add watcher for chart CR status

rossf7:

This adds the new chart CR status watcher. It gets the kubeconfig to connect to the cluster from the chart-operator app CR.

It also replaces the webhook annotation with a new app-namespace annotation. This is needed because chart CRs are always in the giantswarm namespace but the app CR may be in the workload cluster namespace.


integration/test/app/basic/basic_test.go
```diff
@@ -96,6 +96,17 @@ func TestAppLifecycle(t *testing.T) {
...
+			{
+				// chart-operator app CR is used by the chart status watcher
+				// to get a kubeconfig.
+				AppCRName:     key.ChartOperatorUniqueName(),
+				CatalogName:   key.DefaultCatalogName(),
+				Name:          key.ChartOperatorName(),
+				Namespace:     key.Namespace(),
+				ValuesYAML:    templates.ChartOperatorValues,
+				Version:       key.ChartOperatorVersion(),
+				WaitForDeploy: false,
```
rossf7, line 108: Both tests now need a chart-operator app CR.

pkg/label/label.go
```diff
@@ -32,6 +33,13 @@ func AppVersionSelector(unique bool) labels.Selector {
 	return selector
 }
 
+func ChartOperatorAppSelector(unique bool) string {
+	return fmt.Sprintf("%s=%s,%s=%s", label.AppOperatorVersion,
+		GetProjectVersion(unique),
+		label.AppKubernetesName,
+		key.ChartOperatorAppName)
+}
```
rossf7, line 41: This selector gets the correct chart-operator app CR because it may be named chart-operator or chart-operator-unique.

service/watcher/chartstatus/chartstatus.go
```diff
@@ -0,0 +1,186 @@
...
+// watchChartStatus watches all chart CRs in the target cluster for status
+// changes. The matching app CR status is updated otherwise there can be a
+// delay of up to 5 minutes until the next resync period.
+func (c *ChartStatusWatcher) watchChartStatus(ctx context.Context) {
+	for {
+		// We need an active kubeconfig to connect to the cluster as it may be
+		// remote. We get this from the chart-operator app CR. The connection
+		// to the cluster may sometimes be down. So we wait with a backoff
+		// until we can connect.
+		g8sClient, err := c.waitForActiveKubeConfig(ctx)
```
rossf7, line 79: At times like on cluster creation or upgrades the workload cluster may not be available. The backoff retries until a connection can be made.

service/watcher/chartstatus/kubeconfig.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	if len(apps.Items) != 1 {
+		return nil, microerror.Maskf(executionFailedError, "expected 1 chart-operator app CR got %d", len(apps.Items))
+	}
+
+	// We have the chart-operator app CR. Now we get its kubeconfig.
+	app := apps.Items[0]
+
+	// App CR uses inCluster so we can reuse the existing client.
+	if key.InCluster(app) {
```
tomahawk28, line 37: Instead of chart-operator app CRs, can we use `unique` parameter instead to distinguish g8sclients for MC, WC?

service/watcher/chartstatus/kubeconfig.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+
+	return g8sClient, nil
+}
+
+// waitForActiveKubeConfig gets a kubeconfig for the chart-operator app CR.
+// If the target cluster is remote then sometimes the connection will be down
+// so we list all chart CRs to confirm the connection is active.
+func (c *ChartStatusWatcher) waitForActiveKubeConfig(ctx context.Context) (versioned.Interface, error) {
```
tomahawk28, line 78: What do you think divde a function in here, first, we wait for active connection and get g8sClient later. Or we can adjust function name like `getAvailableG8sClient`.

service/watcher/chartstatus/kubeconfig.go
rossf7: Yes I think that's better thanks.

If unique is true we can use the InCluster connection. If its false we use the namespace to generate the kubeconfig name,

service/watcher/chartstatus/kubeconfig.go
rossf7: Yes good to split it and if InCluster is true I don't think we need to list the chart CRs to check the connection.

rossf7 (commented):
@tomahawk28 Changes made. PTAL

tomahawk28 (approved):
LGTM, Thanks for the change.

PR #677, opened 2021-02-09
Replace authtoken with authtokenmigration resource

rossf7:
Towards giantswarm/giantswarm#14163

Renames the resource and its now used to delete the auth token secret.


service/controller/app/resource/authtokenmigration/resource.go
```diff
@@ -0,0 +1,114 @@
...
+		r.logger.Debugf(ctx, "workload cluster is unavailable")
+		r.logger.Debugf(ctx, "canceling resource")
+		return nil
+	}
+
+	r.logger.Debugf(ctx, "deleting secret %#q in namespace %#q", authTokenName, namespace)
+
+	err = cc.Clients.K8s.K8sClient().CoreV1().Secrets(namespace).Delete(ctx, authTokenName, metav1.DeleteOptions{})
+	if apierrors.IsNotFound(err) {
+		r.logger.Debugf(ctx, "already deleted secret %#q in namespace %#q", authTokenName, namespace)
```
tomahawk28, line 95: This log would be flooded in each reconciliation, what do you think to keep logs in case of actual deletion only?

rossf7: Good call. The logs are noisy enough already. I'll do that. 👍

rossf7: @tomahawk28 I updated the logs to make them less noisy. PTAL

tomahawk28 (approved):
LGTM!

PR #681, opened 2021-02-11
Enable VPA and set resources based on the unique flag

rossf7:

### DIFF
--- a/helm/app-operator/values.yaml
+++ b/helm/app-operator/values.yaml
@@ -38,11 +38,18 @@ project:
   commit: "[[ .SHA ]]"
 
 deployment:
-  requests:
-    cpu: 250m
-    memory: 250Mi
-  limits:
-    memory: 280Mi
+  management:
+    requests:
+      cpu: 250m
+      memory: 250Mi
+    limits:
+      memory: 280Mi
+  workload:
+    requests:
+      cpu: 100m
+      memory: 50Mi
+    limits:
+      memory: 50Mi
 
 verticalPodAutoscaler:
   enabled: true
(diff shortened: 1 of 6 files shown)


helm/app-operator/values.yaml
```diff
@@ -38,6 +38,18 @@ project:
...
+      cpu: 250m
+      memory: 250Mi
+    limits:
+      memory: 280Mi
+  workload:
+    requests:
+      cpu: 100m
+      memory: 50Mi
+    limits:
+      memory: 50Mi
```
rossf7, line 52: The resources will be managed by VPA but it uses the ratio between requests and limits.
https://medium.com/infrastructure-adventures/vertical-pod-autoscaler-deep-dive-limitations-and-real-world-examples-9195f8422724

These are set based on my testing and with this we put back the memory limit.

service/controller/app/resource/chart/resource.go
```diff
@@ -124,8 +125,13 @@ func copyChart(current *v1alpha1.Chart) *v1alpha1.Chart {
 // copyAnnotations copies annotations from the current to desired chart,
 // only if the key has a chart-operator.giantswarm.io prefix.
 func copyAnnotations(current, desired *v1alpha1.Chart) {
+	webhookAnnotation := fmt.Sprintf("%s/%s", annotation.ChartOperatorPrefix, annotation.WebhookURL)
+
 	for k, currentValue := range current.Annotations {
-		if !strings.HasPrefix(k, annotation.ChartOperatorPrefix) {
+		if k == webhookAnnotation {
+			// Remove webhook annotation that is no longer used.
+			continue
```
rossf7, line 133: Fixes a bug found in testing. We copy chart-operator annotations so we need to explicitly skip coping the webhook annotation to remove it.

rossf7:
Tested on `ginger`.

tomahawk28 (approved):
LGTM!

PR #687, opened 2021-02-15
Delete resources if app CR uses in cluster

rossf7:
The appnamespace resource checks if the namespace is being deleted and adds a flag to the controller context.

We do this for workload clusters as the cluster is being deleted. But for the app-operator app CR we need to delete these resources.


rossf7:
Tested on gauss

tomahawk28 (approved):

PR #689, opened 2021-02-15
Revert pause annotation

corest:
Towards https://github.com/giantswarm/giantswarm/issues/15502

## Checklist

### DIFF
--- a/service/controller/app/app.go
+++ b/service/controller/app/app.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"time"
 
-	"github.com/giantswarm/apiextensions/v3/pkg/annotation"
 	"github.com/giantswarm/apiextensions/v3/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/k8sclient/v5/pkg/k8sclient"
 	"github.com/giantswarm/microerror"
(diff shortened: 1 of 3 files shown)


service/controller/app/app.go
```diff
@@ -104,11 +103,8 @@ func NewApp(config Config) (*App, error) {
 			K8sClient:    config.K8sClient,
 			Logger:       config.Logger,
 			ResyncPeriod: config.ResyncPeriod,
-			Pause: map[string]string{
-				annotation.AppOperatorPaused: "true",
-			},
-			Resources: resources,
-			Selector:  label.AppVersionSelector(config.UniqueApp),
+			Resources:    resources,
+			Selector:     label.AppVersionSelector(config.UniqueApp),
```
kopiczko, line 107: I don't think we should remove that. It's enough if we remove adding pause annotation in app-admission-controller.

But I leave the final say to @giantswarm/team-batman-engineers

rossf7: Yes there could be other situations where we want to use the pause annotation. So I agree with @kopiczko about keeping it here and just removing from app-admission-controller.

corest:
Done here https://github.com/giantswarm/app-admission-controller/pull/87

PR #692, opened 2021-02-16
Fix semver err

tomahawk28:
Trying to sort appCatalogEntry by created timestamp. 

## Checklist

### DIFF
--- a/service/controller/appcatalog/resource/appcatalogentry/create.go
+++ b/service/controller/appcatalog/resource/appcatalogentry/create.go
@@ -306,17 +306,17 @@ func (r *Resource) sortEntry(ctx context.Context, entries []entry) ([]entry, err
 			return nil, microerror.Mask(err)
 		}
 
-		entries[i].SemVer = *v
+		entries[i].semVer = *v
 		newEntries = append(newEntries, entries[i])
 	}
 
 	sort.Slice(newEntries, func(i, j int) bool {
 		// Removing Prerelease from version since they are mostly SHA strings which we could not compare the size.
-		prevRelease, err := newEntries[i].SemVer.SetPrerelease("")
+		prevRelease, err := newEntries[i].semVer.SetPrerelease("")
 		if err != nil {
 			return false
 		}
-		nextRelease, err := newEntries[j].SemVer.SetPrerelease("")
+		nextRelease, err := newEntries[j].semVer.SetPrerelease("")
 		if err != nil {
 			return false
 		}
--- a/service/controller/appcatalog/resource/appcatalogentry/create.go
+++ b/service/controller/appcatalog/resource/appcatalogentry/create.go
@@ -318,7 +318,7 @@ func (r *Resource) getLatestVersion(ctx context.Context, entries []entry) (strin
 		}
 
 		if nextVersion.Equal(&latestVersion) {
-			if latestCreated.After(entries[i].Created.Time) {
+			if latestCreated != nil && latestCreated.After(entries[i].Created.Time) {
 				latestVersion = nextVersion
 				latestCreated = entries[i].Created.DeepCopy()
 			}
--- a/service/controller/appcatalog/resource/appcatalogentry/create.go
+++ b/service/controller/appcatalog/resource/appcatalogentry/create.go
@@ -179,7 +179,10 @@ func (r *Resource) newAppCatalogEntries(ctx context.Context, cr v1alpha1.AppCata
 			return entries[i].Created.After(entries[j].Created.Time)
 		})
 
-		latestVersion := entries[0].Version
+		latestVersion, err := r.getLatestVersion(ctx, entries)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
 
 		maxEntries := r.maxEntriesPerApp
 		if len(entries) < maxEntries {
--- a/service/controller/appcatalog/resource/appcatalogentry/create.go
+++ b/service/controller/appcatalog/resource/appcatalogentry/create.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strconv"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/giantswarm/apiextensions/v3/pkg/apis/application/v1alpha1"
@@ -169,7 +170,7 @@ func (r *Resource) updateAppCatalogEntry(ctx context.Context, entryCR *v1alpha1.
 	return nil
 }
 
-func (r *Resource) getDesiredAppCatalogEntryCR(ctx context.Context, cr *v1alpha1.AppCatalog, e entry) (*v1alpha1.AppCatalogEntry, error) {
+func (r *Resource) getDesiredAppCatalogEntryCR(ctx context.Context, cr *v1alpha1.AppCatalog, e entry, isLatest bool) (*v1alpha1.AppCatalogEntry, error) {
 	var err error
 	name := fmt.Sprintf("%s-%s-%s", cr.Name, e.Name, e.Version)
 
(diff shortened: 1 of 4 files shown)


tomahawk28:
Tested in `gauss` 

```
k get appcatalogentry -l app.kubernetes.io/name=config-controller,application.giantswarm.io/catalog=control-plane-test-catalog -o yaml | grep Created
    dateCreated: "2021-02-16T10:43:06Z"
    dateCreated: "2021-02-16T10:50:11Z"
    dateCreated: "2021-02-16T10:41:33Z"
    dateCreated: "2021-02-16T10:52:43Z"
    dateCreated: "2021-02-16T10:52:56Z"
```

service/controller/appcatalog/resource/appcatalogentry/create.go
```diff
@@ -288,34 +292,38 @@ func (r *Resource) newAppCatalogEntries(ctx context.Context, cr v1alpha1.AppCata
...
-		return nil, nil
-	}
+func (r *Resource) getLatestVersion(ctx context.Context, entries []entry) (string, error) {
+	var latestVersion semver.Version
+	var latestCreated *metav1.Time
+
+	for i := 0; i < len(entries); i++ {
+		v, err := semver.NewVersion(entries[i].Version)
+		if errors.As(err, &semver.ErrInvalidSemVer) {
+			r.logger.Debugf(ctx, "invalid semver from converting app entries %s, version is %s", entries[i].Name, entries[i].Version)
```
tomahawk28, line 302: We now get this by error for feedback.

service/controller/appcatalog/resource/appcatalogentry/create.go
```diff
@@ -288,34 +292,38 @@ func (r *Resource) newAppCatalogEntries(ctx context.Context, cr v1alpha1.AppCata
 	return entryCRs, nil
 }
 
-func sortEntry(entries []entry) ([]entry, error) {
-	if len(entries) == 0 {
-		return nil, nil
-	}
+func (r *Resource) getLatestVersion(ctx context.Context, entries []entry) (string, error) {
```
rossf7, line 295: This is the latest version by semver right? Can you add godoc for that so its clear its not the latest by create date.

service/controller/appcatalog/resource/appcatalogentry/create.go
```diff
@@ -171,22 +173,24 @@ func (r *Resource) updateAppCatalogEntry(ctx context.Context, entryCR *v1alpha1.
...
 	entryCRs := map[string]*v1alpha1.AppCatalogEntry{}
 
-	for name, entries := range index.Entries {
-		sortedEntries, err := sortEntry(entries)
+	for _, entries := range index.Entries {
+		sort.Slice(entries, func(i, j int) bool {
+			return entries[i].Created.After(entries[j].Created.Time)
+		})
+
+		latestVersion, err := r.getLatestVersion(ctx, entries)
```
rossf7, line 181: It should be an edge case but what happens if the latest version is not in the max entries? Will there still be a CR for it?

rossf7 (commented):
Looks good just a couple of questions.

service/controller/appcatalog/resource/appcatalogentry/create.go
tomahawk28: Oops, no, in these case, latest one would be not included. Do you think it makes sense to add `latest` always?

tomahawk28: So what about adding the latest `AppCatalogEntry` regardless the number of max entries?

rossf7: Yes I think always including latest makes sense. 

Otherwise if you filter by latest and its not there it would be weird for things like Happa when it eventually uses these CRs.

service/controller/appcatalog/resource/appcatalogentry/create.go
```diff
@@ -168,154 +170,178 @@ func (r *Resource) updateAppCatalogEntry(ctx context.Context, entryCR *v1alpha1.
...
+// getLatestEntry returns the entry with the highest version without considering the creation date.
+func (r *Resource) getLatestEntry(ctx context.Context, entries []entry) (entry, error) {
+	var latestIndex int
+	var latestVersion semver.Version
+	var latestCreated *metav1.Time
+
+	for i := 0; i < len(entries); i++ {
+		v, err := semver.NewVersion(entries[i].Version)
+		if errors.As(err, &semver.ErrInvalidSemVer) {
+			r.logger.Debugf(ctx, "invalid semver from converting app entries %s, version is %s", entries[i].Name, entries[i].Version)
```
rossf7, line 274:
```suggestion
			r.logger.Debugf(ctx, "invalid semver from converting app entry %s, version is %s", entries[i].Name, entries[i].Version)
```

I think singular instead of plural is better. As the log is just for this entry.

service/controller/appcatalog/resource/appcatalogentry/create.go
```diff
@@ -168,154 +170,178 @@ func (r *Resource) updateAppCatalogEntry(ctx context.Context, entryCR *v1alpha1.
...
+			continue
+		} else if err != nil {
+			return entry{}, microerror.Mask(err)
+		}
 
-			if entryCR.Spec.Chart.APIVersion == "" {
-				// chartAPIVersion default is `v1`.
-				entryCR.Spec.Chart.APIVersion = "v1"
-			}
+		// Removing Prerelease from version since they are mostly SHA strings which we could not compare the size.
```
rossf7, line 280:
```suggestion
		// Removing Prerelease from version since they are mostly SHA strings which we cannot compare.
```

service/controller/appcatalog/resource/appcatalogentry/create.go
```diff
@@ -168,154 +170,178 @@ func (r *Resource) updateAppCatalogEntry(ctx context.Context, entryCR *v1alpha1.
...
+			if err != nil {
+				return nil, microerror.Mask(err)
 			}
+
+			entryCRs[entryCR.Name] = entryCR
 		}
 
-		return false
-	})
+		// If the latest entry is not included in the desired CRs, we should add them so users could see the latest CR always.
```
rossf7, line 334:
```suggestion
		// If the latest entry is not included in the desired CRs, we add it so users can always see the latest CR.
```

Thanks for adding <3

rossf7 (approved):
LGTM! Just a few minor things about the text.

tomahawk28:
Tested in `geckon` again. 

Get the latest `promtail` chart
```
k get appcatalogentry -lapplication.giantswarm.io/catalog=giantswarm-test,app.kubernetes.io/name=promtail,latest=true
NAME                                                                            CATALOG           APP NAME   APP VERSION   VERSION
                                        AGE
giantswarm-test-promtail-0.1.0-alpha-d5e0a6b7db4f7650e970da164abf8ee1b95505b7   giantswarm-test   promtail   2.1.0         0.1.0-alpha-d5e0a6b7db4f7650e970da164abf8ee1b95505b7   26h
```

This is the latest chart from `index.yaml`

```
...
    name: promtail
...
    urls:
    - https://giantswarm.github.com/giantswarm-test-catalog/promtail-0.1.0-alpha-d5e0a6b7db4f7650e970da164abf8ee1b95505b7.tgz
    version: 0.1.0-alpha-d5e0a6b7db4f7650e970da164abf8ee1b95505b7
```

tomahawk28:
@rossf7 I'm sorry but can you take a look into changes that I made after you approve? it seems good to me but just to double-check with you :)

rossf7 (approved):
LGTM!

Recent changes look good but thanks for checking.

PR #715, opened 2021-03-11
change kubeconfig secret key from kubeConfig to values

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/16005

Uses https://github.com/giantswarm/kubeconfig/pull/60 so we can fetch kubeconfig from CAPI.

### DIFF
--- a/service/controller/app/resource/clients/resource.go
+++ b/service/controller/app/resource/clients/resource.go
@@ -121,7 +121,7 @@ func (r *Resource) addClientsToContext(ctx context.Context, cr v1alpha1.App) err
 
 	var restConfig *rest.Config
 	{
-		restConfig, err = kubeConfig.NewRESTConfigForApp(ctx, key.KubeConfigSecretName(cr), key.KubeConfigSecretNamespace(cr), "values")
+		restConfig, err = kubeConfig.NewRESTConfigForApp(ctx, key.KubeConfigSecretName(cr), key.KubeConfigSecretNamespace(cr), "value")
 		if kubeconfig.IsNotFoundError(err) {
 			// Set status so we don't try to connect to the workload cluster
 			// again in this reconciliation loop.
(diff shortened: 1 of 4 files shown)


service/watcher/chartstatus/g8sclient.go
```diff
@@ -41,7 +41,7 @@ func (c *ChartStatusWatcher) waitForG8sClient(ctx context.Context) (versioned.In
 
 	o := func() error {
 		kubeConfigName := fmt.Sprintf("%s-kubeconfig", c.appNamespace)
-		restConfig, err = kubeConfig.NewRESTConfigForApp(ctx, kubeConfigName, c.appNamespace)
+		restConfig, err = kubeConfig.NewRESTConfigForApp(ctx, kubeConfigName, c.appNamespace, "values")
```
tomahawk28, line 44: We have kubeconfig secrets which have a key `kubeConfig`, so I think we need a fallback mechanism. (this is why we are failing in the integration-test)
1. check `kubeConfig` key first 
2. if not, check `values` key

tomahawk28 (changes requested):

service/controller/app/resource/clients/resource.go
```diff
@@ -121,7 +121,7 @@ func (r *Resource) addClientsToContext(ctx context.Context, cr v1alpha1.App) err
 
 	var restConfig *rest.Config
 	{
-		restConfig, err = kubeConfig.NewRESTConfigForApp(ctx, key.KubeConfigSecretName(cr), key.KubeConfigSecretNamespace(cr))
+		restConfig, err = kubeConfig.NewRESTConfigForApp(ctx, key.KubeConfigSecretName(cr), key.KubeConfigSecretNamespace(cr), "values")
```
tomahawk28, line 124: Here also we need to update.

rossf7:
@TheoBrigitte Going to close this as it was done in 4.1.0. Hope thats OK. 🙏

PR #730, opened 2021-03-23
Drop k8sclient from server pkg

tomahawk28:
We are not using k8sclient from the server anymore after the webhook deletion. 

## Checklist


tomahawk28:
Since this is an internal change I will skip the changelog.

rossf7 (approved):
Nice catch. LGTM

PR #731, opened 2021-03-24
Caching k8sclient

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/16406

## Checklist

### DIFF
--- a/service/controller/app/app.go
+++ b/service/controller/app/app.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/giantswarm/apiextensions/v3/pkg/annotation"
 	"github.com/giantswarm/apiextensions/v3/pkg/apis/application/v1alpha1"
+	cachedk8sclient "github.com/giantswarm/app-operator/v4/service/internal/k8sclient"
 	"github.com/giantswarm/k8sclient/v5/pkg/k8sclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
@@ -22,9 +23,10 @@ import (
 const appControllerSuffix = "-app"
 
 type Config struct {
-	Fs        afero.Fs
-	K8sClient k8sclient.Interface
-	Logger    micrologger.Logger
+	CachedK8sClient *cachedk8sclient.Resource
+	Fs              afero.Fs
+	K8sClient       k8sclient.Interface
+	Logger          micrologger.Logger
 
 	ChartNamespace    string
 	HTTPClientTimeout time.Duration
@@ -42,6 +44,9 @@ type App struct {
 func NewApp(config Config) (*App, error) {
 	var err error
 
+	if config.CachedK8sClient == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.CachedK8sClient must not be empty", config)
+	}
 	if config.Fs == nil {
 		return nil, microerror.Maskf(invalidConfigError, "%T.Fs must not be empty", config)
 	}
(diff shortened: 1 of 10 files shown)


tomahawk28:
I'm testing this from `ghost`. I will check tomorrow morning to see whether we have a problem with a cache.

service/internal/k8sclient/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	if v, ok := r.cache.Get(k); ok {
+		clients, ok := v.(k8sclient.Interface)
+		if !ok {
+			return nil, microerror.Maskf(wrongTypeError, "expected '%T', got '%T'", k8sclient.Clients{}, v)
+		}
+
+		return clients, nil
+	}
+
+	fmt.Println("CACHE FAULT")
```
tomahawk28, line 87: TODO: remove these right before merge

service/internal/k8sclient/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	fmt.Println("CREATE CLIENT")
+	clients, err := r.generateK8sClient(ctx, kubeConfig)
```
tomahawk28, line 89: Only creating the k8sclient when we have a cache fault.

service/internal/k8sclient/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	fmt.Println("CREATE KEY")
+	r.cache.SetDefault(k, clients)
```
tomahawk28, line 95: I'm not sure TTL of each k8sclient, I left the testing operator overnight in `ghost` but it's okay.  Does that means it have more than 12 hours as TTL?

service/controller/app/app.go
```diff
@@ -17,14 +17,16 @@ import (
 	"github.com/giantswarm/app-operator/v4/pkg/label"
 	"github.com/giantswarm/app-operator/v4/pkg/project"
 	"github.com/giantswarm/app-operator/v4/service/controller/app/controllercontext"
+	cachedk8sclient "github.com/giantswarm/app-operator/v4/service/internal/k8sclient"
```
rossf7, line 20:
```suggestion
	"github.com/giantswarm/app-operator/v4/service/internal/k8sclientcache"
```

I found the naming a bit hard to follow here. What about this?

service/internal/k8sclient/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/giantswarm/kubeconfig/v4"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	gocache "github.com/patrickmn/go-cache"
+	"github.com/spf13/afero"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	expiration = 1 * time.Hour
```
rossf7, line 20: I think this TTL is a bit long. The problem is cert renewal for the app-operator cert as I think then the operator won't be able to connect.

What about reducing it to 10 mins or even 5 mins?

rossf7 (commented):
Looks good! Just a couple of queries.

service/controller/app/app.go
tomahawk28: Much better!

service/internal/k8sclient/cache.go
tomahawk28: I think reducing it to 10 mins makes more sense.

service/internal/k8sclientcache/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	"k8s.io/client-go/rest"
+)
+
+const (
+	expiration = 10 * time.Minute
+)
+
+type Config struct {
+	// Dependencies.
+	Fs        afero.Fs
```
rossf7, line 25: What do we need the filesystem for?

rossf7 (approved):
LGTM 

Once debug is removed and let's also remove the filesystem unless its needed.

service/internal/k8sclient/cache.go
rossf7: Cool thanks 👍

service/internal/k8sclientcache/cache.go
tomahawk28: Good catch, I coped it from `client` resource but we don't need it as you said.

PR #733, opened 2021-03-26
Save helmclient as cache

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/16406

## Checklist


service/controller/app/resource/clients/resource.go
```diff
@@ -112,7 +91,7 @@ func (r *Resource) addClientsToContext(ctx context.Context, cr v1alpha1.App) err
 		return nil
 	}
 
-	k8sClient, err := r.k8sClientCache.GetK8sClient(ctx, &cr.Spec.KubeConfig)
+	clients, err := r.k8sClientCache.GetClients(ctx, &cr.Spec.KubeConfig)
```
rossf7, line 94: Should the package now be clientcache instead of k8sclientcache?

Later when we remove the use of the controller context I think we can remove the clients resource and just have clientcache.

rossf7 (approved):
LGTM just a query on the package name

service/controller/app/resource/clients/resource.go
tomahawk28: Good point, yes we should use the correct name.

PR #745, opened 2021-04-07
Update architect-orb and app library

rossf7:
Fixes problems in dependbot PR https://github.com/giantswarm/app-operator/pull/741

### DIFF
--- a/integration/key/key.go
+++ b/integration/key/key.go
@@ -25,7 +25,7 @@ func ChartOperatorUniqueName() string {
 }
 
 func ChartOperatorVersion() string {
-	return "2.7.1"
+	return "2.13.1"
 }
 
 func ControlPlaneTestCatalogName() string {
--- a/integration/key/key.go
+++ b/integration/key/key.go
@@ -28,6 +28,10 @@ func ChartOperatorVersion() string {
 	return "2.13.1"
 }
 
+func ControlPlaneCatalogName() string {
+	return "control-plane-catalog"
+}
+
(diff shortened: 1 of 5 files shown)


integration/key/key.go
```diff
@@ -25,7 +25,7 @@ func ChartOperatorUniqueName() string {
 }
 
 func ChartOperatorVersion() string {
-	return "2.7.1"
+	return "2.13.1"
```
rossf7, line 28: Latest chart-operator for K8s 1.20 support.

integration/test/app/workload/workload_cluster_test.go
```diff
@@ -94,7 +94,7 @@ func TestWorkloadCluster(t *testing.T) {
 			{
 				// Bootstrap chart-operator in the giantswarm namespace.
 				AppCRName:     key.ChartOperatorUniqueName(),
-				CatalogName:   key.DefaultCatalogName(),
+				CatalogName:   key.ControlPlaneTestCatalogName(),
```
rossf7, line 97: This was a bit hard to find but I guess its OK since the unique chart-operator is installed from the CP catalog.

```
status:
  appVersion: ""
  release:
    reason: 'validation error: user configmap must be named `chart-operator-unique-user-values`
      for app in default catalog'
    status: resource-not-found
```

tomahawk28: Good catch. 👍

tomahawk28 (approved):
LGTM

PR #757, opened 2021-05-03
Set latest label to false in appcatalogentry test

rossf7:
Fix to make the appcatalogentry test more reliable.


tomahawk28 (approved):

integration/test/appcatalog/appcatalogentry/app_catalog_entry_test.go
```diff
@@ -98,6 +98,10 @@ func TestAppCatalogEntry(t *testing.T) {
 	}
 
 	{
+		// Set latest label to false to stop the test from flapping. This is
+		// because the latest release may not be the latet according to semver.
```
rossf7, line 102:
```suggestion
		// because the latest release may not be the latest according to semver.
```

PR #791, opened 2021-06-02
Apply LoadCRD into chartCRD resource and integration tests

tomahawk28:
Toward giantswarm/roadmap#219

This PR is a pre-step before updating app-operator to use catalog CRD.

## Checklist


go.mod
```diff
@@ -5,8 +5,9 @@ go 1.16
 require (
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/ghodss/yaml v1.0.0
-	github.com/giantswarm/apiextensions/v3 v3.22.0
+	github.com/giantswarm/apiextensions/v3 v3.26.0
 	github.com/giantswarm/app/v4 v4.13.0
+	github.com/giantswarm/app/v5 v5.0.0-20210603090121-ada93de65e7f
```
tomahawk28, line 10: TODO; update this to 5.0.0 before merge

tomahawk28: I will not release a new app-operator version until we get rid github.com/giantswarm/app/v4

service/internal/crdcache/cache.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/giantswarm/app/v5/pkg/crd"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	gocache "github.com/patrickmn/go-cache"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+const (
+	expiration = 4 * time.Hour
```
tomahawk28, line 16: This resource would fetch the crd once every 4 hours. 🚀

integration/setup/setup.go
```diff
@@ -70,7 +69,12 @@ func installResources(ctx context.Context, config Config) error {
 		for _, crdName := range crds {
 			config.Logger.Debugf(ctx, "ensuring %#q CRD exists", crdName)
 
-			err := config.K8sClients.CRDClient().EnsureCreated(ctx, crd.LoadV1("application.giantswarm.io", crdName), backoff.NewMaxRetries(7, 1*time.Second))
+			crd, err := config.CRDGetter.LoadCRD(ctx, "application.giantswarm.io", crdName)
```
tomahawk28, line 72: As you see, it supports almost the same convention we used it from apiextensions.

service/internal/crdcache/cache.go
rossf7: Thanks that TTL makes sense to me. If we need to refresh it sooner can we just restart the operator?

rossf7 (approved):

service/internal/crdcache/cache.go
tomahawk28: @rossf7 Yes, restarting the pod will load CRDs from GitHub again.

PR #792, opened 2021-06-04
Reconcile to catalog CRs

tomahawk28:
Toward https://github.com/giantswarm/roadmap/issues/219

This is the first step to use `Catalog` CRs in the app-operator.
I will create another PR to update `watcher` and other controllers.

## Checklist


service/controller/catalog/resource/appcatalogentry/resource.go
```diff
@@ -74,15 +74,15 @@ func (r Resource) Name() string {
...
+func (r *Resource) getCurrentEntryCRs(ctx context.Context, cr v1alpha1.Catalog) (map[string]*v1alpha1.AppCatalogEntry, error) {
+	r.logger.Debugf(ctx, "getting current appcatalogentries for catalog %#q", cr.Name)
 
 	currentEntryCRs := map[string]*v1alpha1.AppCatalogEntry{}
 
 	lo := metav1.ListOptions{
 		LabelSelector: fmt.Sprintf("%s=%s,%s=%s", label.ManagedBy, key.AppCatalogEntryManagedBy(project.Name()), label.CatalogName, cr.Name),
 	}
-	entries, err := r.k8sClient.G8sClient().ApplicationV1alpha1().AppCatalogEntries(metav1.NamespaceDefault).List(ctx, lo)
+	entries, err := r.k8sClient.G8sClient().ApplicationV1alpha1().AppCatalogEntries("").List(ctx, lo)
```
tomahawk28, line 85: To get all current AppCatalogEntries CRs across all namespaces, we should use the wildcard here.

service/controller/catalog/resource/appcatalogentry/create.go
```diff
@@ -230,8 +230,7 @@ func (r *Resource) getDesiredAppCatalogEntryCR(ctx context.Context, cr *v1alpha1
 			AppVersion: e.AppVersion,
 			Catalog: v1alpha1.AppCatalogEntrySpecCatalog{
 				Name: cr.Name,
-				// Namespace will be empty until appcatalog CRs become namespace scoped.
-				Namespace: "",
+				Namespace: cr.Namespace,
```
tomahawk28, line 233: `AppCatalogEntry` CRs should be located in the same namespace as the `Catalog` CR.

service/controller/catalog/resource/appcatalogentry/resource.go
rossf7: There is also `metav1.NamespaceAll` what about using that?

https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#pkg-constants

rossf7 (approved):
LGTM

service/controller/catalog/resource/appcatalogentry/resource.go
rossf7, line 85:
```suggestion
	entries, err := r.k8sClient.G8sClient().ApplicationV1alpha1().AppCatalogEntries(metav1.NamespaceAll).List(ctx, lo)
```

Forgot suggestion 🤦‍♂️

tomahawk28: Great! I never expected there would be a variable for "".

rossf7: It's client-go magic there is a constant for everything. 😄 

I just prefer it as it tells you why we're using empty string.

PR #794, opened 2021-06-15
Use catalog CR for app controller

tomahawk28:
## Checklist


go.mod
```diff
@@ -6,10 +6,9 @@ require (
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/ghodss/yaml v1.0.0
 	github.com/giantswarm/apiextensions/v3 v3.26.0
-	github.com/giantswarm/app/v4 v4.13.0
 	github.com/giantswarm/app/v5 v5.0.0
 	github.com/giantswarm/appcatalog v0.4.2
-	github.com/giantswarm/apptest v0.10.3
+	github.com/giantswarm/apptest v0.10.4-0.20210615115515-1c27004547ba
```
tomahawk28, line 11: TODO: update to v0.11.0 after release new version

service/controller/app/resource/catalog/resource.go
```diff
@@ -64,17 +64,36 @@ func (r *Resource) getCatalogForApp(ctx context.Context, customResource v1alpha1
...
+		break
+	}
+
+	if catalog == nil {
+		return microerror.Maskf(notFoundError, "catalog %#q", catalogName)
 	}
 
-	r.logger.Debugf(ctx, "found appCatalog %#q", catalogName)
-	cc.AppCatalog = *appCatalog
+	r.logger.Debugf(ctx, "found catalog %#q in namespace %#q", catalogName, namespace)
```
tomahawk28, line 95: I think we should log the namespace from now.

rossf7: Yes, nice. 👍

rossf7 (commented):
> I will change the integration test to use chart-operator name instead of chart-operator-unique

Good catch. This is because we're installing from the default catalog. We should only hit this in the integration tests.

rossf7 (approved):
Nice, LGTM!

Just needs a changelog unless there are more PRs needed?

tomahawk28:
Thanks, I added it :)

rossf7:
@tomahawk28 Going to go with this to fix tests in #795 for @kopiczko 🤞

PR #806, opened 2021-06-23
Create AppCatalog CRs for compatibility 

tomahawk28:
No issue yet but we briefly talked about how to create appcatalog CRs for compatibility with catalog CRs and we decided to create `appcatalogsync` resource as our solution.

https://gigantic.slack.com/archives/C76JX6YLQ/p1624451154111100
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Changed
 
-- Create `AppCatalog` CRs out of `Catalog` CRs in the cluster namespace for compatibility.
+- Create `AppCatalog` CRs from `Catalog` CRs for compatibility with existing app-operator releases.
 - Prepare helm values to configuration management.
 - Use `Catalog` CRs in `App` controller.
 - Reconcile to `Catalog` CRs instead of `AppCatalog`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Changed
 
+- Create `AppCatalog` CRs out of `Catalog` CRs in the cluster namespace for compatibility.
 - Prepare helm values to configuration management.
 - Use `Catalog` CRs in `App` controller.
 - Reconcile to `Catalog` CRs instead of `AppCatalog`.
--- a/service/controller/catalog/resource/appcatalogsync/create.go
+++ b/service/controller/catalog/resource/appcatalogsync/create.go
@@ -13,7 +13,7 @@ import (
 // EnsureCreated ensures appcatalog CRs are created for compatibility with catalog CRs
 func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	if !r.uniqueApp {
-		// Return early. Only unique instance manages appcatalogentry CRs.
+		// Return early. Only unique instance manages appcatalog CRs.
 		return nil
 	}
 
--- a/service/controller/catalog/resource/appcatalogsync/create.go
+++ b/service/controller/catalog/resource/appcatalogsync/create.go
@@ -27,7 +27,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return nil
 	}
 
-	appCatalogCR := &v1alpha1.AppCatalog{
+	newAppCatalogCR := &v1alpha1.AppCatalog{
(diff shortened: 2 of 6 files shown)


service/controller/catalog/resource/appcatalogsync/create.go
```diff
@@ -0,0 +1,56 @@
...
+		// Return early. Only unique instance manages appcatalogentry CRs.
+		return nil
+	}
+
+	cr, err := key.ToCatalog(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if namespace := cr.GetNamespace(); namespace == metav1.NamespaceDefault || namespace == "giantswarm" {
```
tomahawk28, line 25: I think customers should not care about catalog CRs in the default namespace or giantswarm namespace.  So I would skip creating appcatalog CRs in these cases. WDYT?

service/controller/catalog/resource/appcatalogsync/create.go
```diff
@@ -0,0 +1,56 @@
...
+		// Return early. No need to reconcile catalog CRs in default namespace or giantswarm namespace.
+		return nil
+	}
+
+	appCatalogCR := &v1alpha1.AppCatalog{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        cr.GetName(),
```
tomahawk28, line 32: Copy over the same name, the same labels, and the same annotation.

service/controller/catalog/resource/appcatalogsync/delete.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if namespace := cr.GetNamespace(); namespace == metav1.NamespaceDefault || namespace == "giantswarm" {
+		// Return early. No need to reconcile catalog CRs in default namespace or giantswarm namespace.
+		return nil
+	}
+
+	err = r.k8sClient.G8sClient().ApplicationV1alpha1().AppCatalogs().Delete(ctx, cr.GetName(), metav1.DeleteOptions{})
```
tomahawk28, line 29: I think we should clean up appCatalog CRs created by the app-operator.

service/controller/catalog/resource/appcatalogsync/create.go
rossf7: Yes, I think that makes sense.

service/controller/catalog/resource/appcatalogsync/create.go
```diff
@@ -0,0 +1,56 @@
...
+		Spec: v1alpha1.AppCatalogSpec{
+			Description: cr.Spec.Description,
+			Title:       key.CatalogTitle(cr),
+			Storage: v1alpha1.AppCatalogSpecStorage{
+				Type: "helm",
+				URL:  key.CatalogStorageURL(cr),
+			},
+		},
+	}
+	_, err = r.k8sClient.G8sClient().ApplicationV1alpha1().AppCatalogs().Create(ctx, appCatalogCR, metav1.CreateOptions{})
```
rossf7, line 45: I think we also need to handle updates. As I can see the situation where the catalog URL is wrong when the CR is first created. WDYT?

service/controller/catalog/resource/appcatalogsync/create.go
```diff
@@ -0,0 +1,56 @@
...
+	if apierrors.IsAlreadyExists(err) {
+		// no-op
+		return nil
+	} else if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.Debugf(ctx, "created appCatalog %#q for compatibility", cr.GetName())
```
rossf7, line 53: Can you add the usual logging with creating and created? Just for consistency.

rossf7 (commented):
Looks good. Just a couple of things.

service/controller/catalog/resource/appcatalogsync/create.go
tomahawk28: You're right, I will make it handle updates as well.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Changed
 
+- Create `AppCatalog` CRs out of `Catalog` CRs in the cluster namespace for compatibility.
```
rossf7, line 12:
```suggestion
- Create `AppCatalog` CRs from `Catalog` CRs for compatibility with existing app-operator releases.
```

rossf7 (approved):
LGTM

PR #840, opened 2021-08-09
Limit k8s version on package rules

tomahawk28:
## Checklist

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -16,7 +16,7 @@
       "groupName": "etc modules"
     },
     {
-      "matchPackageNames": ["^k8s.io"],
+      "matchPackagePatterns": ["^k8s.io"],
       "allowedVersions": "< 1.21.0"
     },
     {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -9,7 +9,8 @@
     },
     {
       "matchPackagePatterns": ["^k8s.io", "^sigs.k8s.io"],
-      "groupName": "k8s modules"
+      "groupName": "k8s modules",
+      "allowedVersions": "<= 1.20.0"
```
rossf7, line 13: I think we want to use the latest patch version of 1.20. Will that work with this?

Also I think we need to add controller-runtime as we need to use 0.6.X WDYT?

See https://github.com/giantswarm/app-operator/pull/831

tomahawk28: Then I think this should be `< 1.21.0` to include all patches on 1.20. Thanks

rossf7 (approved):
Thanks, LGTM

PR #846, opened 2021-08-13
Validate Catalog CRs instead of AppCatalog CRs

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/18384

## Checklist

### DIFF
--- a/service/controller/app/resource/catalog/resource.go
+++ b/service/controller/app/resource/catalog/resource.go
@@ -68,7 +68,7 @@ func (r *Resource) getCatalogForApp(ctx context.Context, customResource v1alpha1
 
 	var namespaces []string
 	{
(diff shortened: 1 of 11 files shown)


service/controller/app/resource/catalog/resource.go
```diff
@@ -87,7 +87,7 @@ func (r *Resource) getCatalogForApp(ctx context.Context, customResource v1alpha1
 		break
 	}
 
-	if catalog == nil {
+	if catalog == nil || catalog.Name == "" {
```
rossf7, line 90: @tomahawk28 I added the same check here for consistency. WDYT?

rossf7 (commented):
Tested on gauss

PR #899, opened 2021-12-21
Ensure .spec.namespace in app CRs is immutable

rossf7:
## Checklist


rossf7:
This can't be implemented. As we need to know the previous state of the app CR which is only available in app-admission-controller.

PR #904, opened 2022-01-14
Embed chart CRD to avoid hitting GitHub API rate limits

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/20379

I'd forgotten but there was already an internal [crdcache](https://github.com/giantswarm/app-operator/blob/e7deab43464e0b5cf99c9b475e2f89ac6e0f2764/service/internal/crdcache/cache.go) resource.

It caches the CRD for 4 hours. This helps but not if the initial API call fails. The resource supports setting a github token but configuring app-operator with a secret for this is a pain.

So this embeds the chart CRD. This is basically the same as the original approach when we vendored apiextensions.

## Checklist


Makefile.custom.mk
```diff
@@ -0,0 +1,6 @@
+# Directories.
+SCRIPTS_DIR := hack
+
+sync-chart-crd:
+	@echo "$(GEN_COLOR)Sync Chart CRD with apiextensions-application$(NO_COLOR)"
+	cd $(SCRIPTS_DIR); ./sync-chart-crd.sh
```
rossf7, line 6: This is the same approach we use in apptestctl.

piontec (approved):
freaking rate limiting, LGTM :)

rossf7:
> freaking rate limiting

:100: :)

Tested on gauss

PR #920, opened 2022-02-02
Fix error handling in chart CR watcher

rossf7:
Follow up for https://github.com/giantswarm/giantswarm/issues/20539

## Checklist

### DIFF
--- a/service/watcher/chartstatus/dynclient.go
+++ b/service/watcher/chartstatus/dynclient.go
@@ -65,7 +65,9 @@ func (c *ChartStatusWatcher) waitForAvailableConnection(ctx context.Context, dyn
 		// List all chart CRs in the target cluster to confirm the connection
 		// is active and the chart CRD is installed.
 		_, err = dynClient.Resource(chartResource).Namespace(c.chartNamespace).List(ctx, metav1.ListOptions{})
-		if err != nil {
+		if IsResourceNotFound(err) {
+			return microerror.Maskf(resourceNotFoundError, "chart CRD not found")
+		} else if err != nil {
 			return microerror.Mask(err)
 		}
 
--- a/service/watcher/chartstatus/dynclient.go
+++ b/service/watcher/chartstatus/dynclient.go
@@ -83,7 +83,8 @@ func (c *ChartStatusWatcher) waitForAvailableConnection(ctx context.Context, dyn
 		c.logger.Debugf(ctx, "failed to get available connection: %#v retrying in %s", err, t)
 	}
 
-	b := backoff.NewExponential(5*time.Minute, 30*time.Second)
+	// maxWait is 0 since cluster creation may fail.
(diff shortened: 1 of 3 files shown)


service/watcher/chartstatus/dynclient.go
```diff
@@ -76,7 +83,8 @@ func (c *ChartStatusWatcher) waitForAvailableConnection(ctx context.Context, dyn
 		c.logger.Debugf(ctx, "failed to get available connection: %#v retrying in %s", err, t)
 	}
 
-	b := backoff.NewExponential(5*time.Minute, 30*time.Second)
+	// maxWait is 0 since cluster creation may fail.
+	b := backoff.NewExponential(0, 30*time.Second)
```
rossf7, line 87: >	// After MaxElapsedTime the ExponentialBackOff returns Stop.
>	// It never stops if MaxElapsedTime == 0.

https://github.com/cenkalti/backoff/blob/6b0e4ad0cd65431b217393bff47d1dff727b264b/exponential.go#L59_L
61

Since cluster creation takes more than 5 mins the backoff then returns Stop and retries after 0 seconds. 🤦‍♂️ 

If there is a problem the cluster may never come up. So we should keep retrying every 30 secs.

service/watcher/chartstatus/dynclient.go
```diff
@@ -128,7 +136,8 @@ func (c *ChartStatusWatcher) waitForKubeConfig(ctx context.Context) (*corev1.Sec
 		c.logger.Debugf(ctx, "failed to get kubeconfig: %#v retrying in %s", err, t)
 	}
 
-	b := backoff.NewExponential(5*time.Minute, 30*time.Second)
+	// maxWait is 0 since kubeconfig creation may fail.
+	b := backoff.NewExponential(0, 30*time.Second)
```
rossf7, line 140: Same deal for the kubeconfig.

service/watcher/chartstatus/dynclient.go
```diff
@@ -65,7 +66,13 @@ func (c *ChartStatusWatcher) waitForAvailableConnection(ctx context.Context, dyn
 		// List all chart CRs in the target cluster to confirm the connection
 		// is active and the chart CRD is installed.
 		_, err = dynClient.Resource(chartResource).Namespace(c.chartNamespace).List(ctx, metav1.ListOptions{})
-		if err != nil {
+		if tenant.IsAPINotAvailable(err) {
+			c.logger.Debugf(ctx, "workload cluster is not available")
+			return microerror.Mask(err)
+		} else if IsResourceNotFound(err) {
+			c.logger.Debugf(ctx, "chart CRD is not installed")
+			return microerror.Mask(err)
```
rossf7, line 74: The errors can be cryptic. So handling each expected error and logging why the watcher can't proceeed.

MarcelMue (approved):
Makes sense to me!

PR #931, opened 2022-02-16
Speed up apps installation after bootstrapping chart-operator

ljakimczuk:
## Description

Towards: https://github.com/giantswarm/giantswarm/issues/20728

## Checklist

### DIFF
--- a/service/controller/app/resource/chartoperator/create.go
+++ b/service/controller/app/resource/chartoperator/create.go
@@ -136,6 +136,8 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			r.logger.Debugf(ctx, "release %#q deployed", releaseName)
 			r.logger.Debugf(ctx, "triggering charts reconciliation")
 
+			// Checks for App CRs without corresponding Chart CRs in the workload cluster,
+			// and then annotate them to trigger reconciliation and speed up bootstrapping.
 			err = r.triggerReconciliation(ctx, cr)
 			if err != nil {
 				return microerror.Mask(err)
--- a/service/controller/app/resource/chartoperator/create.go
+++ b/service/controller/app/resource/chartoperator/create.go
@@ -132,6 +132,16 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			}
 
 			r.logger.Debugf(ctx, "deleted release %#q", releaseName)
+		case helmclient.StatusDeployed:
+			r.logger.Debugf(ctx, "release %#q deployed", releaseName)
+			r.logger.Debugf(ctx, "triggering charts reconciliation")
+
+			err = r.triggerReconciliation(ctx, cr)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			r.logger.Debugf(ctx, "triggered charts reconciliation")
 		}
 	}
 
--- a/service/controller/app/resource/chartoperator/create_test.go
+++ b/service/controller/app/resource/chartoperator/create_test.go
@@ -98,8 +98,6 @@ func Test_Resource_triggerReconciliation(t *testing.T) {
 				t.Fatalf("error == %#v, want nil", err)
 			}
 
-			fakeLogger := microloggertest.New()
-
 			var r *Resource
 			{
 				var objs []runtime.Object
@@ -114,6 +112,7 @@ func Test_Resource_triggerReconciliation(t *testing.T) {
 					Build()
 
 				fakeK8sClient := clientgofake.NewSimpleClientset()
+				fakeLogger := microloggertest.New()
 
 				var valuesService *values.Values
 				{
--- a/service/controller/app/resource/chartoperator/create_test.go
+++ b/service/controller/app/resource/chartoperator/create_test.go
@@ -8,9 +8,8 @@ import (
 	"github.com/giantswarm/apiextensions-application/api/v1alpha1"
 	"github.com/giantswarm/app/v6/pkg/key"
 	"github.com/giantswarm/app/v6/pkg/values"
-	"github.com/giantswarm/helmclient/v4/pkg/helmclient"
-	"github.com/giantswarm/helmclient/v4/pkg/helmclienttest"
 	"github.com/giantswarm/k8sclient/v6/pkg/k8sclienttest"
+	"github.com/giantswarm/k8smetadata/pkg/annotation"
 	"github.com/giantswarm/k8smetadata/pkg/label"
 	"github.com/giantswarm/micrologger/microloggertest"
 	"github.com/spf13/afero"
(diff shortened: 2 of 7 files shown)


service/controller/app/resource/chartoperator/resource.go
```diff
@@ -20,7 +25,8 @@ import (
 )
 
 const (
-	Name = "chartoperator"
+	Name                             = "chartoperator"
+	AppOperatorTriggerReconciliation = "app-operator.giantswarm.io/trigger-reconciliation"
```
ljakimczuk, line 29: Should we move it to the https://github.com/giantswarm/k8smetadata/blob/be2bc263815cd4b8a8a76ed35ca1b15ab1cc9120/pkg/annotation/app.go? Currently I am setting a date as a value, so maybe `triggered-reconciliation-at` would be more appropriate name?

service/controller/app/resource/chartoperator/create_test.go
```diff
@@ -0,0 +1,400 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgofake "k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/giantswarm/app-operator/v5/service/controller/app/controllercontext"
+)
+
+func Test_Resource_EnsureCreated(t *testing.T) {
```
ljakimczuk, line 26: The two `Test_` functions are now redundant, sort of, though I believe we could keep them anyway in case we extend test cases for the `EnsureCreated`.

service/controller/app/resource/chartoperator/create_test.go
```diff
@@ -0,0 +1,400 @@
...
+					WithScheme(scheme.Scheme).
+					WithRuntimeObjects(tc.charts...).
+					Build()
+
+				fakeClient := k8sclienttest.NewClients(k8sclienttest.ClientsConfig{
+					CtrlClient: fakeCtrlClient,
+					K8sClient:  clientgofake.NewSimpleClientset(),
+				})
+
+				defaultContent := helmclient.ReleaseContent{
```
ljakimczuk, line 164: This will skip certain part of the `EnsureCreated` where this release is actually installed, though I don't believe it's necessarily bad since we are interested in the end result of having `StatusDeployed`, so that Apps can be annotated. And whether it is truly installed or not does not seem to depend on code.

service/controller/app/resource/chartoperator/create_test.go
```diff
@@ -0,0 +1,414 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/giantswarm/app-operator/v5/service/controller/app/controllercontext"
+)
+
+func Test_Resource_EnsureCreated(t *testing.T) {
+	tests := []struct {
+		name             string
+		apps             []runtime.Object
```
rossf7, line 29: Using `runtime.Object` is a bit strange here. Could we use the actual types and then convert them when setting up the fake client?

service/controller/app/resource/chartoperator/create.go
```diff
@@ -132,6 +132,16 @@ func (r Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			}
 
 			r.logger.Debugf(ctx, "deleted release %#q", releaseName)
+		case helmclient.StatusDeployed:
+			r.logger.Debugf(ctx, "release %#q deployed", releaseName)
+			r.logger.Debugf(ctx, "triggering charts reconciliation")
+
+			err = r.triggerReconciliation(ctx, cr)
```
rossf7, line 139: Could you add a comment here for why we need this?

The comments in the method are good but I think having something here too would be useful.

service/controller/app/resource/chartoperator/create_test.go
rossf7: I think just having a UT for `triggerReconciliation` is enough.

The tests for client-go become verbose quickly and often an integration test is better. The chart-operator bootstrapping is tested by the WorkloadClusterTest so I don't think we need coverage for `EnsureCreated`.

https://github.com/giantswarm/app-operator/blob/c72fc6ad2601b0908cd0ae294fd3166082c0d6f0/integration/test/app/workload/workload_cluster_test.go#L29-L31

service/controller/app/resource/chartoperator/create_test.go
```diff
@@ -0,0 +1,414 @@
...
+				}
+			}
+		})
+	}
+}
+
+func Test_Resource_triggerReconciliation(t *testing.T) {
+	tests := []struct {
+		name             string
+		apps             []runtime.Object
```
rossf7, line 205: Using `runtime.Object` is a bit strange here. Could we use the actual types and then convert them when setting up the fake client?

service/controller/app/resource/chartoperator/resource.go
MarcelMue: I think the name is fine but moving it to k8smetadata makes sense to me.

rossf7: Yes let's add it there. I think `triggered-reconciliation` is fine. We just need the update event so we don't care about the value. Although the timestamp is a good idea for debugging. 👍

service/controller/app/resource/chartoperator/resource.go
```diff
@@ -140,6 +149,79 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
 	return nil
 }
 
+func (r Resource) triggerReconciliation(ctx context.Context, operatorApp v1alpha1.App) error {
```
rossf7, line 152:
```suggestion
func (r Resource) triggerReconciliation(ctx context.Context, chartOperatorApp v1alpha1.App) error {
```

service/controller/app/resource/chartoperator/resource.go
```diff
@@ -140,6 +149,79 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	// For each App, check if the corresponding Chart CR exists.
+	// If not, annotate the App to trigger the reconciliation.
+	for i, app := range appList.Items {
+		// Skip for in-cluster apps and the chart-operator app itself.
+		if key.InCluster(app) || app.ObjectMeta.Name == operatorApp.ObjectMeta.Name {
```
MarcelMue, line 185: Why skip for incluster apps?

service/controller/app/resource/chartoperator/resource.go
```diff
@@ -140,6 +149,79 @@ func (r Resource) installChartOperator(ctx context.Context, cr v1alpha1.App) err
...
+
+			if len(app.GetAnnotations()) == 0 {
+				app.Annotations = map[string]string{}
+			}
+
+			modifiedApp := app.DeepCopy()
+			modifiedApp.Annotations[AppOperatorTriggerReconciliation] = metav1.Now().Format(time.RFC3339)
+
+			// Using indexing to fix the `G601: Implicit memory aliasing in for loop.`
+			err = r.ctrlClient.Patch(ctx, modifiedApp, client.MergeFrom(&appList.Items[i]))
```
MarcelMue, line 211: Is it maybe easier to generate the patch directly? instead of generating the patch from a deepcopy. I believe submitting the patch directly (as e.g. a JSONPatch) would guard us against possible race conditions.

rossf7 (commented):
Looks good but some changes are needed for the tests.

service/controller/app/resource/chartoperator/resource.go
ljakimczuk: Thanks guys, I'm introducing this new annotation here: https://github.com/giantswarm/k8smetadata/pull/32

service/controller/app/resource/chartoperator/create_test.go
ljakimczuk: Thanks @rossf7 , makes sense to me, I'm removing this

service/controller/app/resource/chartoperator/resource.go
ljakimczuk: Well, the honest answer is that I thought this logic matters only for bootstrapping WC and its apps, and then I also thought all of them are in-cluster apps, except for the app-operator 🤔

service/controller/app/resource/chartoperator/resource.go
ljakimczuk: Thanks for the suggestion @MarcelMue ! I will give it a try

service/controller/app/resource/chartoperator/resource.go
rossf7: Yes, this is just for speeding up the install of workload apps in new clusters. The MC apps in the collection are reconciled by flux and the new types in-cluster apps aren't affected.

service/controller/app/resource/chartoperator/resource.go
ljakimczuk: @MarcelMue fixed 🙏

ljakimczuk:
@MarcelMue @rossf7 I took care of the changes you suggested, would you like to take another look?

service/controller/app/resource/chartoperator/create_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+				if expectedNum != len(a.GetAnnotations()) {
+					t.Fatalf("%s: expected %d, got %d", a.ObjectMeta.Name, expectedNum, len(a.GetAnnotations()))
+				}
+			}
+		})
+	}
+}
+
+func newV1alpha1App(crName, appName, cluster, organization string, inCluster bool) *v1alpha1.App {
```
rossf7, line 204:
```suggestion
func newApp(crName, appName, cluster, organization string, inCluster bool) *v1alpha1.App {
```

Minor but I don't think we need the API version here.

service/controller/app/resource/chartoperator/create_test.go
```diff
@@ -0,0 +1,250 @@
...
+				InCluster: inCluster,
+			},
+			Name: appName,
+		},
+	}
+
+	return c
+}
+
+func newV1alpha1Chart(name, namespace string) *v1alpha1.Chart {
```
rossf7, line 237:
```suggestion
func newChart(name, namespace string) *v1alpha1.Chart {
```

Same here.

rossf7 (approved):
LGTM just couple of minor naming things

PR #960, opened 2022-03-09
Set app CR status to pending when resources do not exist yet

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/21038

When create default apps via default-apps-openstack or default-apps-aws we want the child app CRs to be in status `pending` until the kubeconfig and cluster values configmap exist.

Currently they are `failed` which is confusing for users.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  architect: giantswarm/architect@4.14.1
+  architect: giantswarm/architect@4.14.2
 
 workflows:
   build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  architect: giantswarm/architect@dev:bump-apptestctl
+  architect: giantswarm/architect@4.14.1
 
 workflows:
   build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  architect: giantswarm/architect@4.14.2
+  architect: giantswarm/architect@dev:bump-apptestctl
 
 workflows:
   build:
--- a/go.mod
+++ b/go.mod
@@ -8,12 +8,12 @@ require (
 	github.com/giantswarm/apiextensions-application v0.3.1
 	github.com/giantswarm/app/v6 v6.8.1
 	github.com/giantswarm/appcatalog v0.6.0
-	github.com/giantswarm/apptest v1.1.1
+	github.com/giantswarm/apptest v1.2.0
 	github.com/giantswarm/backoff v1.0.0
 	github.com/giantswarm/errors v0.3.0
 	github.com/giantswarm/helmclient/v4 v4.9.0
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  architect: giantswarm/architect@4.14.1
```
rossf7, line 3: This bumps apptestctl with the updated AppCatalogEntry CRD.

There is also a fix for this in https://github.com/giantswarm/app-operator/pull/961

go.mod
```diff
@@ -8,12 +8,12 @@ require (
 	github.com/giantswarm/apiextensions-application v0.3.1
 	github.com/giantswarm/app/v6 v6.8.1
 	github.com/giantswarm/appcatalog v0.6.0
-	github.com/giantswarm/apptest v1.1.1
+	github.com/giantswarm/apptest v1.2.0
```
rossf7, line 11: We need this fix to the apptest library.

I bumped the others too since renovate isn't working.

pkg/status/status.go
```diff
@@ -5,13 +5,17 @@ const (
 	// merge configmaps.
 	ConfigmapMergeFailedStatus = "configmap-merge-failed"
 
-	// ResourceNotFoundStatus is set in the CR status when there is an failure during
-	// finding dependents kubernete resources.
-	ResourceNotFoundStatus = "resource-not-found"
+	// PendingStatus is set in the CR status when the kubeconfig secret
+	// or cluster values configmap does not exist yet.
+	PendingStatus = "pending"
```
rossf7, line 10: This is the new status.

pkg/status/status.go
```diff
@@ -5,13 +5,17 @@ const (
...
 
 	// SecretMergeFailedStatus is set in the CR status when there is an failure during
 	// merge secrets.
 	SecretMergeFailedStatus = "secret-merge-failed"
+
+	// ValidationFailedStatus is set in the CR status when there is a validation error
+	// not related to dependent kubernetes resources.
+	ValidationFailedStatus = "validation-failed"
```
rossf7, line 18: Previously this was `resource-not-found` I changed it to `validation-failed` as that is more accurate.

service/controller/app/resource/validation/create.go
```diff
@@ -21,10 +21,22 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	}
 
 	_, err = r.appValidator.ValidateApp(ctx, cr)
-	if validation.IsValidationError(err) {
+
+	if validation.IsAppConfigMapNotFound(err) || validation.IsKubeConfigNotFound(err) {
```
MarcelMue, line 25: Can we maybe also check what the current status is? E.g. if there is no status -> set pending, if there was a status (e.g. deployed) then it should be error? Does that make sense?

rossf7: @MarcelMue Yes that makes total sense to check if it's already installed.

This didn't work out in testing. Currently the status is empty instead of resource-not-found. 

I'm going to move this back to draft until I've tested some more. Sorry for the noise.

PR #964, opened 2022-03-10
Default dns

whites11:
If coredns is installed as an app (that is the case in management clusters) there is a chicken-egg problem:
app operator needs coredns to resolve catalog URLs' hostnames, coredns needs app operator to be installed.

IF and this is an IF I ask to the reviewers app operator does NOT need to resolve cluster-internal names, than with this PR we solve this chicken-egg problem

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


rossf7 (approved):
I think we introduced new chicken egg problem with app-operator 5.6.0

We get the chart CR URLs from the index.yaml before we generated them

whites11:
closing as we decided to solve this differently

PR #995, opened 2022-05-20
Change VPA settings

uvegla:
Towards: https://github.com/giantswarm/giantswarm/issues/21980

Removing the `limits` from the `Deployment` lets the `VPA` figure out the proper resource limits (based on historic data it collects and analyses) that it sets when the pod starts. The controller keeps watching the pod and updates the limits / recreates the pod if needed within the max limits set in the `VPA`.

If the VPA is unavailable or not enabled we still set the the `limits` on the `Deployment`.

## References

- https://github.com/kubernetes/design-proposals-archive/blob/main/autoscaling/vertical-pod-autoscaler.md
- https://www.giantswarm.io/blog/vertical-autoscaling-in-kubernetes

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 *.swp
 
 # build result
-app-operator
+/app-operator
 app-operator-*
 
 # don't apply earlier restrictions under vendor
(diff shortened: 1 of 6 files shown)


uvegla:
The integration tests are failing with some Helm templating issue 😢  Will check on Monday 📅

kubasobon (approved):

PR #1002, opened 2022-06-01
Support Catalog with multiple Repositories

kubasobon:
- fixup: refactor some code
- fixup: refactor some code

## Checklist

### DIFF
--- a/service/controller/app/resource/chart/desired.go
+++ b/service/controller/app/resource/chart/desired.go
@@ -50,37 +50,9 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
-	var tarballURL, version string
-
-	if key.CatalogVisibility(cc.Catalog) == "internal" {
-		// For internal catalogs we generate the URL as its predictable
-		// and to avoid having chicken egg problems.
-		tarballURL, err = appcatalog.NewTarballURL(key.CatalogStorageURL(cc.Catalog), key.AppName(cr), key.Version(cr))
-		if err != nil {
-			r.logger.Errorf(ctx, err, "failed to generated tarball")
-		}
-
-		version = key.Version(cr)
-	} else {
-		// For all other catalogs we check the index.yaml for compatibility
-		// with community catalogs.
-		index, err := r.indexCache.GetIndex(ctx, key.CatalogStorageURL(cc.Catalog))
-		if err != nil {
-			r.logger.Errorf(ctx, err, "failed to get index.yaml")
-		}
-
-		version, tarballURL, err = getVersionAndTarballURL(index, key.AppName(cr), cr.Spec.Version)
-		if err != nil {
-			r.logger.Errorf(ctx, err, "failed to get tarball URL")
-		}
-
-		if !isValidURL(tarballURL) {
-			// URL may be relative. If so we join it to the Catalog Storage URL.
-			tarballURL, err = joinRelativeURL(cc.Catalog, tarballURL)
-			if err != nil {
-				r.logger.Errorf(ctx, err, "failed to join relative URL")
-			}
-		}
+	tarballURL, version, err := r.buildTarballURL(ctx, cc, cr)
+	if err != nil {
+		return nil, microerror.Mask(err)
 	}
 
 	chartCR := &v1alpha1.Chart{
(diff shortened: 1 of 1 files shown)


kubasobon:
@giantswarm/team-honeybadger PTAL at the new suggested flow. Tests are obviously failing for now. I will fix and extend them once we agree on the design.

service/controller/app/resource/chart/desired.go
```diff
@@ -111,6 +110,123 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
+	}
+
+	var chart v1alpha1.Chart
+	err := cc.Clients.K8s.CtrlClient().Get(
+		ctx,
+		types.NamespacedName{Name: chartName, Namespace: r.chartNamespace},
+		&chart,
+	)
+	if apierrors.IsNotFound(err) || tenant.IsAPINotAvailable(err) {
+		// Repositories is guaranteed by Custom Resource Definition to have at least one entry.
```
MarcelMue, line 125: Should we really rely on this? I think we should fall back to `repository` if `repositories[0]` is nil. Otherwise we will need to transition all app CRs somehow before we roll this out.

kubasobon: The `repositories` array is in Catalog CRD. We will need to transition all Catalog CRs (copy stuff in `storage` to `repositories[0]`), but I think that's an ok thing to do. WDYT?
I can add a fallback to `storage`, but that's incurring technical debt - we would have to update the code again after we deprecate `storage`.

MarcelMue: If we can control all catalog CRs, then that is fine to me. Some customers are creating catalogs now, so it might be some effort to try and get them to move their source of truth over. Maybe there is a metric or something that let's us check how many catalogs actually exist?

service/controller/app/resource/chart/desired.go
```diff
@@ -111,6 +110,123 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
+		}
+	}
+	if repositoryIndex == -1 {
+		// Could not match current tarballURL to any of Catalog's repositories.
+		// Maybe the list was updated. Let's pick any existing repository.
+		r.logger.Debugf(ctx, "could not match tarball URL %q to any of %q Catalog repositories; using default", chart.Spec.TarballURL, cc.Catalog.Name)
+		return cc.Catalog.Spec.Repositories[0].URL, nil
+	}
+
+	if chart.Status.Release.Status == releaseNotInstalledStatus {
```
uvegla, line 146: In the latest `chart-opererator` the `not-installed` status is set when:
- failed to pull, not found or timed out
- also when the chart failed to install and the release is not found

The `unknown-error` can be added when the release fails but there is an already deployed release. I am not sure we need this, but maybe we want to retry on those too.

What I was more concerned about is that there is a `FailedMaxAttempts` field stored in the controller context. The limit is currently [set to 5](https://github.com/giantswarm/chart-operator/blob/0ccda9890974f224b0eb032030d1b3887b39db48/pkg/project/project.go#L11-L16). It [checks the last 5 releases](https://github.com/giantswarm/chart-operator/blob/0ccda9890974f224b0eb032030d1b3887b39db48/service/controller/chart/resource/releasemaxhistory/create.go#L130-L143) in the release history. What [happens after this](https://github.com/giantswarm/chart-operator/blob/0ccda9890974f224b0eb032030d1b3887b39db48/service/controller/chart/resource/releasemaxhistory/create.go#L47-L64) is not super clear to me why we do that. Maybe it has no effects on the fallback charts / failing them multiple times even when the list is long and maybe you thought of it as well.

service/controller/app/resource/chart/desired.go
kubasobon: Let me check, but there should be a reasonable number of clusters for us to "migrate". I can update `kubectl-gs` to generate the new Catalog CR format.

kubasobon: I checked and there are currently 7 custom Catalog CRs (excluding default Giant Swarm ones):
```
===> bumblebee
NAMESPACE    NAME                                  CATALOG URL                                                                 AGE
default      nfs-subdir-external-provisioner       https://kubernetes-sigs.github.io/nfs-subdir-external-provisioner/          118d
--
===> fox
NAMESPACE    NAME                                  CATALOG URL                                                                 AGE
default      hephy                                 https://charts.teamhephy.com/                                               88d
--
===> gauss
NAMESPACE    NAME                                  CATALOG URL                                                                 AGE
default      demo-catalog-lj                       https://ljakimczuk.github.io/test-helm-repo/                                34d
default      machine-learning                      https://pipo02mix.github.io/ml-catalog/                                     206d
--
===> ginger
NAMESPACE    NAME                                  CATALOG URL                                                                 AGE
org-ross     hephy                                 https://charts.teamhephy.com/                                               125d
--
===> helios
NAMESPACE    NAME                                  CATALOG URL                                                                 AGE
default      thg-example-app-catalog               https://thehutgroup.github.io/example-app-catalog                           20d
--
===> tiger
NAMESPACE    NAME                                  CATALOG URL                                                                 AGE
default      hdi-apps-sandbox                      https://pg2000.github.io/tkp-app-catalog/                                   68d
```

service/controller/app/resource/chart/desired.go
kubasobon: See: https://github.com/giantswarm/chart-operator/pull/889

kubasobon: I just realized I don't have to check for `MaxFailedAttempts`. There won't be any Helm release to retry nor any history, since this is only for cases when pulling the chart has failed.

service/controller/app/resource/chart/desired.go
```diff
@@ -111,6 +110,124 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
+		// Could not match current tarballURL to any of Catalog's repositories.
+		// Maybe the list was updated. Let's pick any existing repository.
+		r.logger.Debugf(ctx, "could not match tarball URL %q to any of %q Catalog repositories; using default", chart.Spec.TarballURL, cc.Catalog.Name)
+		return cc.Catalog.Spec.Repositories[0].URL, nil
+	}
+
+	if chart.Status.Release.Status == chartPullFailedStatus {
+		// chart-operator had trouble pulling the chart -- this includes timeouts and chart not being found (404)
+		// Round-robin the repository.
+		repositoryIndex = repositoryIndex + 1%len(cc.Catalog.Spec.Repositories)
```
piontec:
```suggestion
		repositoryIndex = repositoryIndex + 1 % len(cc.Catalog.Spec.Repositories)
```

service/controller/app/resource/chart/desired.go
```diff
@@ -111,6 +110,124 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
+	}
+
+	// For all other catalogs we check the index.yaml for compatibility
+	// with community catalogs.
+	index, err := r.indexCache.GetIndex(ctx, repositoryURL)
+	if err != nil {
+		r.logger.Errorf(ctx, err, "failed to get index.yaml from %q: %v", repositoryURL)
+	}
+
+	if index == nil || len(index.Entries) == 0 {
```
piontec, line 176: Is it maybe better to give different error msg on index.yaml not existing and being empty?

piontec (approved):
LGTM

service/controller/app/resource/chart/desired.go
kubasobon: I fixed it anyways, it was missing parentheses.

service/controller/app/resource/chart/desired.go
kubasobon: Done!

kubasobon:
Ok, @MarcelMue & @uvegla you're free to take a look. This is pretty much the final shape of things.

service/controller/app/resource/chart/desired.go
kubasobon: In the end I added a `switch/case` which takes Catalogs without `.spec.repositories` but `.spec.storage` defined into account. Better be safe than sorry.

uvegla (approved):

kubasobon:
Merging. It won't be released for a while, but that's ok.

PR #1014, opened 2022-06-15
Only run podmonitor outside of bootstrap mode

MarcelMue:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Add support for Catalogs that define multiple repository mirrors to be used in case some of them are unreachable.
 
+### Changed
+
+- Only run `PodMonitor` outside of bootstrap mode.
+
 ## [6.0.0] - 2022-06-08
 
 ### Added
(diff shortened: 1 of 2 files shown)


kopiczko (approved):

kubasobon (approved):
LGTM :rocket:

MarcelMue:
@whites11 Can you check here? I have no idea how I could easily test if this fixes the current issues.

whites11:
> @whites11 Can you check here? I have no idea how I could easily test if this fixes the current issues.

not really IMHO. There might still be race conditions where app operator in "production mode" gets deployed before the CR exists, right?

In other places we used to check if the CRD exists before creating the CR.
Example: https://github.com/giantswarm/azure-operator/blob/master/helm/azure-operator/templates/vpa.yaml#L1

Does it make sense to use that approach instead?

MarcelMue:
> > In other places we used to check if the CRD exists before creating the CR. Example: https://github.com/giantswarm/azure-operator/blob/master/helm/azure-operator/templates/vpa.yaml#L1
> > Does it make sense to use that approach instead?
> 
> I think it may mean the PodMonitor is never created in some situations even though the CRD is eventually created.

Yes, I would rather have it fail in that case than have no monitoring of app-operator. Especially if we can't detect easily that monitoring is missing otherwise.

whites11 (approved):

PR #1023, opened 2022-06-30
Make `app-operator` PSS restricted compliant

stone-z:
Sets the default/runtime seccomp profile, and more explicitly restrictive securityContexts for the pod and container

## Checklist


piontec (approved):
LGTM :)

PR #1030, opened 2022-07-08
Check for default named user config not referenced

AverageMarcus:
Fixes: https://github.com/giantswarm/roadmap/issues/1195

If an App doesn't have a `ConfigMap` or `Secret` referenced in the `UserConfig` but the namespace does contain one that follows the standard naming convention of `*-user-values` or `*-user-secrets` then we'll update the App resource to reference those found resources.

## Checklist

### DIFF
--- a/service/controller/app/resource/configmap/desired.go
+++ b/service/controller/app/resource/configmap/desired.go
@@ -11,7 +11,6 @@ import (
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/operatorkit/v6/pkg/controller/context/resourcecanceledcontext"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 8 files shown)


service/controller/app/resource/configmap/desired.go
```diff
@@ -41,6 +42,19 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		return configMap, nil
 	}
 
+	// If no user-provided configmap name is present, check if a *-user-values config map exists and set the reference
+	if key.UserConfigMapName(cr) == "" {
+		userCM, err := cc.Clients.K8s.K8sClient().CoreV1().ConfigMaps(r.chartNamespace).Get(ctx, fmt.Sprintf("%s-user-values", cr.Name), metav1.GetOptions{})
+		if err == nil || !apierrors.IsNotFound(err) {
```
uvegla, line 48: I am thinking that I would only proceed if there are no errors. In theory any other error occurs - e.g. no permission or context deadline - it would update the CR and promotes these permanent / temporary issues further. Same apply for the secrets part.

uvegla:
Generally looks good to me, left a comment. Thanks a lot for doing this :)

service/controller/app/resource/configmap/desired.go
AverageMarcus: Hmm... yeah, not sure why I did it like that. You're totally right. Updated :)

uvegla: Cool, thank you! :) 👍🏻

PR #1053, opened 2022-09-26
Fix a bug that skips adding chart cr cm and / or secret when they are set only via extra configs

uvegla:
## Checklist


piontec (approved):
LGTM!

PR #1068, opened 2022-11-18
Fix race conditions

ljakimczuk:
## Description

Towards: https://github.com/giantswarm/giantswarm/issues/24582, https://github.com/giantswarm/giantswarm/issues/23441

## Checklist

### DIFF
--- a/integration/test/catalog/appcatalogentry/app_catalog_entry_test.go
+++ b/integration/test/catalog/appcatalogentry/app_catalog_entry_test.go
@@ -138,7 +138,7 @@ func TestAppCatalogEntry(t *testing.T) {
 				Namespace: metav1.NamespaceDefault,
 			},
 			Chart: v1alpha1.AppCatalogEntrySpecChart{
-				APIVersion:  "v2",
+				APIVersion:  "v1",
 				Description: latestEntry.Description,
 				Home:        latestEntry.Home,
 				Icon:        latestEntry.Icon,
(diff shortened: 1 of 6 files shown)


integration/test/catalog/appcatalogentry/app_catalog_entry_test.go
```diff
@@ -139,7 +139,7 @@ func TestAppCatalogEntry(t *testing.T) {
 				Namespace: metav1.NamespaceDefault,
 			},
 			Chart: v1alpha1.AppCatalogEntrySpecChart{
-				APIVersion:  "v2",
+				APIVersion:  "v1",
```
ljakimczuk, line 142: I've changed this in response to the missing metadata for the latest versions of `prometheus-operator-app` in the `giantswarm` catalog. Since the `chartAPIVersion` is taken from metadata, lack of it means setting the default value, being the `v1`. I don't know what have changed in the `prometheus-operator-app` and why it is missing, will create an issue for this for someone to investigate, but for now I need the test to reflect the reality.

uvegla: It is a known issue. It is missing because it was migrated away from ABS back to architect. Talked with Atlas and there is a PR to migrate it back to ABS that they will release as part of the next release soonish.

uvegla: Part of this PR: https://github.com/giantswarm/prometheus-operator-app/pull/207

ljakimczuk: ok, I will leave this change in place then, when they switch back, we can fix the test again then.

kubasobon (approved):

uvegla (approved):

PR #1127, opened 2023-03-29
Improving statuses

ljakimczuk:
## Description

First, let's quickly summarize what we currently have. The `chart` resource tries to find the app in different Helm repositories, starting with the primary, and  then falls back to secondary ones if the primary one fails for getting the app. And when, for whatever reason, the `chart` resource fails to find the app the whole resource returns an error and the further processing of CR is stopped. This means, we never get to the `status` resource responsible for updating the CR status. I believe returning the error, instead of gracefully falling through, has been done for some reason, but on the other hand the aforementioned logic seems to anticipate some of the errors, like the `notFoundError`, which means to me, it does not always need to really fail without giving any feedback in status.

So, what has been changed in this PR. Since I believe the controller is composed currently the way it is for a reason, I am a bit hesitant to change it completely. But, since the part of it that searches for an app in catalogs anticipates known errors, I believe it is safe to change this part. Hence, the `notFoundError` has been broken into more specific errors, like the `appVersionNotFoundError`,  and specific CR's statuses have been assigned to these errors. Then I have also thought that it should be safe to handle unknown errors coming from this part as well, for after all this means something is wrong on the AppOp - Helm Repository line and this is something we would like to share in the status field. So the `unknown-error` status has been introduced in addition. When any error is seen, the `chart` resource now does not fail, but uses the controller context to pass down the status, known or unknown, and then falls through to the `status` resource. For the other cases, it behaves like previously.

Towards: https://github.com/giantswarm/giantswarm/issues/26445

## Checklist


service/controller/app/resource/chart/desired.go
```diff
@@ -68,26 +70,26 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
+	repositories := []string{repositoryURL}
```
ljakimczuk, line 73: @kubasobon you have written this part I believe, so not sure if you are ok with this change, to me it feels bit more readable now, but also changes one thing in the logic. Previously it has been falling back to other repositories only on the `notFoundError` seen with the primary one. But then for the fallback repositories the type of error has not been checked. So this to me feels like we do not really need to check for this in the primary case as well, but maybe you had a specific reason for doing this that way.

kubasobon (approved):
LGTM :rocket:

service/controller/app/resource/chart/desired.go
kubasobon: Looks good. I think it's ok to round-robin through repositories on other kinds of errors as well. I was overly cautious with initial design, I think.

PR #1155, opened 2023-07-14
Introducing support for the Helm Controller as a downstream operator for installing Helm Charts

ljakimczuk:
Towards: https://github.com/giantswarm/roadmap/issues/2455

# Description

This PR introduces changes necessary for the App Operator to work with the Helm Controller. It carries some netorious changes in a form of new resources intrudced to the existing services and their controllers, as well as changes to the existing resources, and on top of that introduces new unit and integration tests. All of this is described in more detail below.

### App Controller

The App Controller receives two bare new resources with this PR, being the `helmrelease` and the `migration`. The former is obviously an equivalent of the `chart` resource which is being used when Chart Operator is configured as a downstream controller. The latter is intended to be used only for the migration time being and is in charge of deleting an app's Chart CR and its related configuration, and once all clusters are switched to the Helm Controller, the resource may get deleted, for then there should be no Chart CRs left. On top of that, the App Controller receives updates to the `clients`, `configmap`, `secret` and `status` resources to enable them to work with the HelmRelease CRs.

The order in which these resources are being executed when reconciling an object is presented below.

obraz

A brief description of each of the mentioned resources is presented in the next sections.

#### The `helmrelease` resource

The logic of translation of an App CR into a HelmRelease CR remains the same as to translation to a Chart CR. The obvious differences that follows from using this new type, are:

* HelmRelease CR is created in a management cluster, in the same namespace the App CR lives in, and in case app is being installed for a remote cluster this cluster's kubeconfig is being used.
* ConfigMap and Secret resources, carrying the Helm Chart values, are being configured with a HelmRelease CR's `.spec.valuesFrom` field. These values are however still a result of the merging procedure implemented by the App Operator.
* instead of choosing a Helm Chart tarball URL, the App Operator chooses one of the available HelmRepository CRs and configures it in a HelmRelease CR's `.spec.chart.spec.sourceRef` field. The fallback logic remains the same as for the Chart Operator-coupled App Operator.
* App CR dependencies annotatiotion results in setting `true` value for the `.spec.suspend` field of a HelmRelease CR, instead of setting `pause` annotation.
* App Operator sets the `app-operator.giantswarm.io/latest-configmap-version`, `app-operator.giantswarm.io/latest-secret-version` and the `reconcile.fluxcd.io/requestedAt` annotations on a HelmRelease CR, which carry the latest ConfigMap resource version, the latest Secret resource version, and concoatenation of the two respecively. The first two annotations are equivalents of a Chart CR `.spec.config.[configMap,secret].resourceVersion` fields, and the last once is to trigger a reconciliation in case either one of the config changes. Here an explanation is due. The version annotation are mostly informational here and have been left for consistency reasons, updating them does not trigger a reconciliation, as there are no predicates used for metadata changes detection in the [Helm Controller](https://github.com/fluxcd/helm-controller/blob/dcec6902c816e1929c9ad7c1fe5cb4860b76b1b1/internal/controller/helmrelease_controller.go#L121C14-L121C36). As you can see the only predicate used is the one that checks for `reconcile.fluxcd.io/requestedAt`. Since `values.yaml` changes do not change HelmRelease CR spec in any way, and no metadata except for the requested-at annotation is respected, it indicates we must use this annotation somehow smartly to notify the Helm Controller of config changes.
* the wait options are left enabled, despite they haven't been previously used with Chart Operator. The reason being, the Chart Operator could skip wait because it was capable of querying a Helm release directly, something that the Helm Controller is uncapable of doing. Hence the Chart Operator could eventually learn the release state, unlike the Helm Controller, which must wait in order to know it.

A brief summary of these changes is depicted on the below figure. **IMPORTANT**: in the meantime I decided to change the ConfigMap and Secret names from `<app>-chart-[values,secrets]` to `<app>-helmrelease-[values,secrets]`, the image below still caries the old names, for I have lost its source, hence can't update it.

obraz

#### The `migration` resource

For context, since the switch from the Chart Operator to the Helm Controller is intended for the CAPI clusters only, it makes sense to remove the Chart Operator in the Cluster Apps Operator who installs it for the CAPI clusters. How exactly it is going to be handled there is irrelevant now, what matters is when it is done for a particular workload cluster, its Chart CRs must be removed. Since these CRs were previously created by the App Operator, it makes sense for it to be in charge of removal. And this is exactly what the `migration` resource is responsible for. Its logic can brefily be described by the below points:

* it adds the pause annotation to the Chart CR. This effectively disables reconciliation of an app by the Chart Operator. It is an extra precaution in case the Chart Operator does not get removed.
* it checks for the Chart Operator deployment existence in the `giantswarm` namespace. If one is found, it skips further processing.
* if no Chart Operator is found, it removes a Chart CR created previously for an App CR being processed. This step requires also removing finalizers for this Chart CR, for there is no Chart Operator that can handle them.
* once done with Chart CR it checks for remaining configuration, either ConfigMap or Secret previously referenced by the Chart CR, and removes it. This is the step that influenced ConfigMap and Secrets names update from `-chart-` to `-helmrelease-`. To understand the nature of the potential problem consider re-using the `-chart-` names and migrating resources for different scenarios. First, we switch WC App Operator in some ORG namespace. It creates HelmRelease CRs and corresponding `-chart-` configuration resources in the ORG namespace, and removes Chart CR and `-chart-` configuration resources from the WC's `giantswarm` namespace. All good. Then consider we switch MC App Operator and watches in-cluster app in some ORG namespace, because it is a bundle for example. Then HelmRelease CR and corresponding `-chart-` configuration resources get created for such an app in the ORG namespace, and Chart CR and its `-chart-` configuration resources get deleted from the `giantswarm` namespace of MC, all good as well. But when we watch an in-cluster app in the `giantswarm` namespace, because it is a collection app for example, we see the App Operator creates HelmRelease CR and does not create the `-chart-` configuration resources, for they already exists in this namespace. But then the `migration` resources deletes the Chart CR and the `-chart-` configuration resources, which effectively breaks the HR. To avoid this case analysis, it has been decided to change the names. 

The App Operator coupled with Helm Controller has no business in a workload cluster except for the above procedure. To satisfy this extra needs, a new set of clients is being used, called the `MigrationClients`. Due to how the App Operator is currently implemented adding these extra clients is much easier than using other methods or conditional logic, and once there are no workload clusters left to migrate, removing this part should also be much simpler.

#### The `clients` resource

The App Operator supports two types of clients. First one that allows connectivity to a management cluster the App Operator runs in, call it local, and the other that allows connectivity to a workload cluster the App Operator provides apps for, call it remote. So far the logic has been simple, the App Operator has been using local client for in-cluster apps, and remote client for non in-cluster apps. This PR extends that logic as follow:

* a secondary set of clients is introduced for the migration procedure, while the primary set of clients is still being used for app-related operations.
* the primary clients set is always local  whenever an App CR is in-cluster or the Helm Controller backend is enabled
* the primary clients set is remote whenever an App CR is not in-cluster and the Helm Controller backend is disabled
* the secondary clients set is local whenever an App CR is in-cluster and Helm Controller backend is enabled
* the secondary clients set is remote whenever an App CR is not in-cluster and Helm Controller backend is enabled
* the secondary clients set is nil whenever an App CR is not in-cluster and Helm Controller backend is disabled

This extended logic is briefly depicted below.

obraz

#### The `configmap` resource

This resource has slightly been changed, so that instead of creating ConfigMap in a remote cluster, inside the `giantswarm` namespace, it is from now on created in a management cluster, next to the App CR and HelmRelease CR. The name of this ConfigMap has changed slightly, so now it is of the form `%s-helmrelease-values`. Semantically there is nothing wrong with the `%-chart-values` name, but this name causes troubles upon migration, if we want to do it automatically.

#### The `secret` resource

It is a sibling resource to the `configmap` resource and has undergone the same set of changes with the same reasoning behind. 

#### The `status` resource

The `status` resource from now on is capable of populating App CR status from HelmRelease CR status, on top of already being able to do it from a Chart CR status. Unlike in the Chart CR case, there is no 1:1 correspondence between App CR and HelmRelease CR statuses, hence a translation of a sort must be used, to get former from the latter. Some fields are easy to translate, like version for example, but to know the actual state of the Helm release, or reconciliation altogether, the `.status.conditions` array must be checked upon. There are four types of conditions offered by this array:

* `Ready` - status of the last reconciliation attempt
* `Released` - status of the last release attempt
* `TestSuccess` - status of the last test attempt
* `Remediated` - status of the last remediation attempt

The solution obviously translates the conditions array into an App CR-compatible status, but before discussing of how it works, to better understand decisions taken, let me share the three main premises the solution was based on:

* as like in the case of Chart Operator we are primarily interested in the Helm release state if there is information about it
* if something has interfered the release, for example a bad configuration, we are interested in this information instead of a release state
* the current set of statuses offered by the App Platform is to be maintained so that user switching from Chart Operator to Helm Controller is not confused by one App Platform reports about his apps

With these premises in mind, the solution can be characterized as follow:

* if there is the `Released` condition in the conditions array, grab its reason and follow one of the below sub-steps:
  + if there is a mapping for the grabbed reason to one of the App Platform **successful** statuses, use it and stop further processing
  + if there is a mapping for the grabbed reason to one of the App Platform **failed** statuses, follow the blow sub-steps:
    - if failure reason is **known** to App Platform use a pre-defined status for it and stops further processing
    - if failure reason is **unknown** to App Platform use the general status, the grabbed reason is mapped to, and stop further processing
  + the grabbed reason is **unknown** to us, hence use the `unknown` status and stop further processing
* if there is no `Released` condition in the conditions array, check for the `Ready` condition. If there is such, grab its reason and follow one of the below sub-steps:
  + if there is a mapping for the grabbed reason to one of the App Platform statuses, mostly failed, but not only, use it and stop further processing
  + the grabbed reason is unknown to us, hence use the `unknown` status and stop further processing
* if neither the `Released` nor the `Ready` status are available, use the empty status, for this means the HelmRelease CR either hasn't been picked up by the Helm Controller yet.

We check for the `Released` condition first because we are primarily interested in Helm release status. If we find mapping between this condition's reason and App Platform success status, we use this status. If we do not find anything, we try to find mapping between the reason and App Platform failure statuses and in case one is found, we try to narrow it down to one of the known error statuses. If there is a match, we use this known error status, otherwise we use status as given by the mapping. If no mappings are found for the reason, we use the `unknown` status.

When there is no `Released` condition we check for the `Ready` condition, for it may indicate problems on the way to executing Helm actions, like artifact failures. We again try to find mapping between this condition reason and known App Platform statuses. If one is found, we then use the status. If no mapping is found, we use the `unknown` status.

The `Remediated` and `TestSuccess` conditions are not used by us, because these actions are not enabled in HelmRelease CR created by the App Operator. Also, these hasn't been truely used previously with the Chart Operator, with some small exceptions for the remediation actions, hence to enable them right now we would need to know of how we want to configure them and what we want to show to a user in the App CR status. Yet, once we figure this out the algorithm should be easily extensible.

The brief summary of the algorithm is depicted below.

obraz

**IMPORTANT**: it seems that Helm Controller is more verbal in what it does, in context of status changes, than Chart Operator. It means that it changes status of HelmRelease CRs far more often than Chat Operator used to do for Chart CRs, at least it is what I have observed. To some extent it is a problem for us because `status` resource is not the only place we process statuses at. Another such place is status watcher, which is also capable of updating App CR status field from the HelmRelease CR status. But this, with often changes to HelmRelease CR status, may lead to conflicts, when one of these two places reports: 

```text
the object has been modified; please apply your changes to the latest version and try again
```

There seem to be no way to coordinate the two separate resources, hence I decided to account for this error in the `status` resource, which so far has been returning reconciliation failure on experiencing any error on updating App CR status. From now on it will be returning reconciliation error in case of any error, except the conflict above.In case of conflict it will return successful reconciliation, but will report the error so that we have clear information about it in the logs. The watcher already does not treat it like a problem.

### Catalog Controller

The Catalog Controller gets a new resource called `helmrepository` that is in charge of translating Catalog CR into HelmRepository CR(s).

#### The `helmrepository` resource

The resource takes Catalog CR as an input, and then uses its `.spec.storage` and `.spec.repositories` fields to create HelmRepository CR(s). Once created, it then updates the Catalog CR's status with a list of created HelmRepository CR(s), and deletes the ones that are no longer desired, based on the previous Catalog CR's status.

Naming convention used for HelmRepository CR, which is presented below, may look complicated at first:

`<Catalog CR name>-<repository type>-<repository hostname>-<repository path>`

It is however justified by these facts:

* one Catalog CR can point to multiple repository of the same hostname, but different paths
* one Catalog CR can point to multiple repositories of the same damain / hostname, but different paths

The mere fact that a single Catalog CR may create multiple HelmRepository CRs prohibits using the Catalog CR name solely, and implies using some of repositories' properties. A natural choice is to extend the base name, which is the Catalog CR name, with a path, which in our case, I believe, is always indicative of the content of the repository. For example, the `https://giantswarm.github.io/control-plane-catalog/` repository URL tells we can find MAPI components inside, so it would be a good choice for an extension, yet may in some instances be the same or very similar to the Catalog CR name. The problem is, if a single Catalog CR pointed to multiple repositories of different hostnames, but the same path, a simple naming convention like that would fail. Hence it is worth to add hostname as well to the naming. And to make it more complete, a scheme has been added as well yielding a final convention.

Once HelmRepository CRs are created, the Catalog CR status is going to be updated with a list of them, see an example of Catalog CR below. The status is to allow us to manage the HelmRepository CRs, so that we know what to delete when a Catalog CR gets deleted, or when it gets updated by removing some of previously configured repositories.

```yaml
apiVersion: application.giantswarm.io/v1alpha1
kind: Catalog
metadata:
  name: cluster
  namespace: giantswarm
spec:
  config: {}
  description: This catalog holds production revisions of helm charts defining clusters
    and their pre-installed apps.
  logoURL: /images/repo_icons/giantswarm.png
  repositories:
  - URL: https://giantswarm.github.io/cluster-catalog/
    type: helm
  - URL: oci://giantswarmpublic.azurecr.io/cluster-catalog/
    type: oci
  storage:
    URL: https://giantswarm.github.io/cluster-catalog/
    type: helm
  title: Giant Swarm Cluster Catalog
status:
  helmRepositoryList:
    entries:
      - name: cluster-helm-giantswarm.github.io-cluster-catalog
        namespace: giantswarm
      - name: cluster-oci-giantswarmpublic.azurecr.io-cluster-catalog
        namespace: giantswarm
```

## Watcher Service

The Watcher Service got a new watcher, called the `helmreleasestatus`, for watching changes in the HelmRelease CRs the App Operator has created. See a brief description of its logic in the next section.

### The `helmreleasestatus` watcher

The `helmreleasestatus` watcher works by watching HelmRelease CRs created by the App Operator, by the means of using the same labels selector the App Operator uses for App CRs it processes. For each of HelmRelease CRs, returned from a watch request, the watcher turns its status to an App CR desired status, and compares it with a current App CR status. If both statuses are the same, nothing obviously happens. If both are different, the watcher update an App CR status.

## Checklist


piontec:
Thanks for the detailed write-up (I read only the PR description), it is really a nice summary and I think everything you wrote about the changes makes sense.

I have a few extra questions:
1. I see that you left dependency management entirely on the app-operator side. What do you think about moving that already to flux, by applying its `dependsOn` feature?
2. How `HelmRepositories` are managed (when they are created/deleted/ in which namespace)?
3. Can I tell, by looking at the App CR only, if it is Char or HR backed?

ljakimczuk:
@piontec

> I see that you left dependency management entirely on the app-operator side. What do you think about moving that already to flux, by applying its dependsOn feature?

Well, good question and not that easy to answer.

I am not sure the Flux dependency feature does not have any loopholes to it that we haven't discovered. I think one, maybe not tied to the mechanism itself, is the status that gets reported, which is failed. Hence we would get failed status on App CR, while it is currently none. So both us and Flux sees this state differently. Failed status also implies risk of getting extra alerts on occasion. We could handle the status internally, to make it empty, but this is extra work, small but unnecessary, if we truly want it empty, for we have the solution for that already. We could also handle the status to mask it to something new, like the `waiting`, but this can be done even today without using Flux, and would probably be needed to should we choose to go down this path, for one of the postconditions that we wanted to establish after the migration, was to offer the same set of statuses for both types of clusters.

I think the most important thing to note is, using Flux directly gains us nothing really. As long as the App CR remains the user interface, we still need a way and hence a code in App Operator to get dependencies from a user and to process them. Now we check them in App Operator what translates to a set of annotations. Then we would need translate them to a slice, hence some code would still be there. And the footprint of this part of the code is small enough for us to not really worry about it. 

Other thing that worries me, related to App CR being the primary interface, is inconsistency between CAPI and legacy. For now, dependency mechanism means dependencies between App CRs for both types of environments. If we use Flux directly, this mechanism would have its nowaday meaning for legacy environments, and for CAPI environemts it would mean Flux dependencies between HelmRelease CRs, what introduces inconsistencies I am not a fan of and may not be something obvious for a user a user who works with App CR and is uninterested of Flux. It may be also not a good idea for us to do it if we are ever to tweak the naming convention for HelmRelease CRs, and decide to now follow App CRs names 1:1 for example, for we may then break something unintentionally. It sounds unlikely, but the recent situation with Chart CRs names for CAPI shows we are susceptible to these kinds of errors, especially we may anticipate a good deal of changes in the future.

Despite the above considerations, both ways should give the same outcome, yet I rather lean towards using App Operator-based dependencies for consistency reasons. In my humble opinion, once the heavy lifting related to migration is done, and an insight of the new setup and how it works, is gathered, I think then would be the time to consider possible improvements.

> How HelmRepositories are managed (when they are created/deleted/ in which namespace)?

The Catalog Controller of App Operator has a `helmrepository` resource that is in charge of creating / updating / deleting HelmRepository CRs. Actions are taken place in response to Catalog CR changes which are creation / update / deletion. HelmRepository CRs follow the Catalog CR and hence are created in the same namespace the latter lives in.

> Can I tell, by looking at the App CR only, if it is Char or HR backed?

No, you may check the App Operator configuration for that, or notice a lack of Chart Operator for your cluster. I think porting back this information to every single App CR, where you have at least two central places for getting it, is a little bit over the top.

PR #1177, opened 2023-09-14
Fixing helmclient

ljakimczuk:
## Checklist

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -31,6 +31,10 @@
       "allowedVersions": "< 0.7.0"
     },
     {
+      "matchPackageNames": ["helm.sh/helm/v3"],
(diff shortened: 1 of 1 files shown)


marians:
Will this not be overwritten by the Synchronize workflow? To change this, I think an override like this one is needed:

https://github.com/giantswarm/github/blob/main/repositories/override/app-build-suite/renovate.json

ljakimczuk:
@marians it probably will but right now I am in the trial and error stage.

marians (commented):
Go ahead then :)

ljakimczuk:
@marians can I get an approval pls?

marians (approved):

PR #1210, opened 2023-11-14
add-option-to-disable-k8s-client-cache

calvix:
the client cache does not work for EKS as the kubeconfig is pretty short-lived (around 10-15 minutes) and the caching interval is 10 minutes as well so if the app-operator  caches the kubeconfig very close to the EKS kubeconfig expiration then there is a situation that the app operator cannot work for several minutes, in E2E CI its around 5-8 minutes, this breaks e2e for EKS

towards https://github.com/giantswarm/roadmap/issues/2820

I assume same thing affect the imported EKS clusters

### DIFF
--- a/flag/service/kubernetes/kubernetes.go
+++ b/flag/service/kubernetes/kubernetes.go
@@ -1,0 +1,18 @@
+package kubernetes
+
+import (
+	"github.com/giantswarm/operatorkit/v8/pkg/flag/service/kubernetes/tls"
+	"github.com/giantswarm/operatorkit/v8/pkg/flag/service/kubernetes/watch"
+)
+
+// Kubernetes is a data structure to hold Kubernetes specific command line
+// configuration flags.
+type Kubernetes struct {
+	Address            string
+	DisableClientCache string
+	InCluster          string
+	KubeConfig         string
+	KubeConfigPath     string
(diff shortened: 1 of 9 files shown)


flag/service/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,18 @@
...
+// configuration flags.
+type Kubernetes struct {
+	Address            string
+	DisableClientCache string
+	InCluster          string
+	KubeConfig         string
+	KubeConfigPath     string
+	TLS                tls.TLS
+	Watch              watch.Watch
+}
```
calvix, line 18: this file was previously imported from `operatorkit` repo but since we need the extra property `DisableClientCache string` i redeclared it here

whites11: I would normally ask to make the change in operatorkit, but since operatorkit is not developed and won't be used anymore soon I guess it's ok

whites11 (approved):

flag/service/kubernetes/kubernetes.go
calvix: That was my thinking as well

mproffitt (approved):
LGTM

PR #1214, opened 2023-11-21
Use quay instead of docker

mnitchev:
- Use quay.io instead of docker.io
- Update CHANGELOG

## Checklist


marians:
As discussed, we are doing this differently

PR #1267, opened 2024-04-03
Add support for App resources having a dependency on HelmReleases

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/3119

## Motivation

Team Turtles is moving default apps for CAPI clusters from default-apps-aws to cluster chart and cluster-aws.

In order to be able to declare correct dependencies between the apps where some apps are deployed as HelmReleases (Flux) and some as App resources (Giant Swarm app platform), we need the possibility that an App has a dependency on a HelmRelease.

## Implementation

This PR adds support for App resources having a dependency on HelmReleases. The feature is opt-in and backward compatible and requires setting specific annotations.

e.g. In order to specify that the App `my-app` has a dependency on the HelmRelease `my-helmrelease`, the following annotations are required:

```yaml
metadata:
  name: my-app
  annotations:
    app-operator.giantswarm.io/depends-on-helmrelease: "true"
    app-operator.giantswarm.io/depends-on: "my-helmrelease"
```

Annotation `app-operator.giantswarm.io/depends-on-helmrelease` must be set to a non-empty string, otherwise app-operator will not check HelmReleases when checking if the dependencies are installed.

A dependency on a HelmRelease is considered to be satisfied when the following is true for a HelmRelease:
- `spec.chart.spec.version` is equal to `status.lastAppliedRevision`, and
- `status.conditions` has a condition with `type` equal to `Ready` and `status` equal to `True`.

The implementation is using the dynamic client and unstructured objects, in order not to add a large dependency tree from the upstream Flux project (which has major version v0, so there would be more chance that something breaks when go dependency is updated to a newer version).

## Testing

I have added unit tests for checking dependencies. As there were no tests before, I have added tests for both App and HelmRelease dependencies.

Will also test it on live clusters and update the PR afterwards (convert it to non-draft).

## Checklist


Gacko (approved):

nprokopic:
Log from app-operator when App observability-bundle depends on HelmRelease coredns which is not installed:

```
{"caller":"github.com/giantswarm/micrologger@v1.0.0/logger.go:66","controller":"app-operator-app","event":"update","function":"GetDesiredState","level":"debug","loop":"235","message":"Not creating chart for app \"defapps003-observability-bundle\": dependencies not satisfied [defapps003-coredns]","object":"org-nikola/defapps003-observability-bundle","resource":"chart","time":"2024-04-21T13:58:49.53521+00:00","version":"1030136723"}
```

mproffitt:
How does this align with, or conflict with the work being done on replacing chart operator with helmcontroller? https://github.com/giantswarm/roadmap/issues/2455

The expectation of #2455 is that chart-operator will be removed entirely from app platform and as a result of this, there will be an imperative relationship between app-operator and helm-controller in all future iterations of app-platform

cc: @ljakimczuk

nprokopic:
> How does this align with, or conflict with the work being done on replacing chart operator with helmcontroller?

I don't think this should conflict with the mentioned work, as this is simply a missing feature, a critical one that we actually need immediately in KaaS.

ljakimczuk (approved):

mproffitt (approved):

PR #1356, opened 2025-02-18
Making cordon-until expire automatically

ljakimczuk:
Towards: https://github.com/giantswarm/giantswarm/issues/32684

## Description

Previously, to cordon an app one had to set both `cordon-reason` and `cordon-until` annotations, and the cordoning itself had no real expiration date, in the sense that after reaching up the `cordon-until` date, nothing was automatically resumed and hence cordoning was still in power. To prevent apps suspencion forever the system was designed to fire an alert to notify an on-call engineer of expired cordon.

This PR changes the cordining behaviour as follows. The `cordon-reason` is not mandatory from now on, but it can still be set for informative purposes. The `cordon-until` is hence the only annotation required for cordoning. It remains the same, except, it won't suspend an app forever now, from now on when the App Operator sees the current time has passed the point given in the `cordon-until`, it will continue reconciliation as normal, as if `cordon-until` was not set.

There is one thing that to some may seem unusual, that is, when a `cordon-until` is set, but with a wrong time format, including a string that does not carry time in any format, results in app suspencion. There are two reasons for that, the first one is, when a user sets the annotation he has an intent to cordon his app. If the time format is wrong and we treat his app as not cordoned, this nullifies his intent. Now, if he has included the app's version upgrade in the same update operation he used to add the annotation, this would result in his app being immediately rolled out, which is not what we want. 

Another way of handling the above scenario, would be to not cordon an app in bad-time-format case, and instead return an immediate reconciliation error. In such case it would work like cordoning, for a reconciliation error stops processing further OperatorKit resources, so no changes would be applied, hence nothing would get rolled out. The problem, and the second reason for cordoning, I have with this approach is, I would like to let user know the time format is wrong and I would like to do it through the App CR status, I think that's reasonable taking we had complaints on visibility of errors over the last years. 

And now, if I bail on reconciliation on spotting the parsing error, I would need to update an App CR status at the point in code where I'm bailing out, but then we also have the `status` OperatorKit resource that is run last and which is put in charge of updating statuses. So setting status earlier would be repeating its logic, which feels unnecessary to me, after all we have it for a reason.

Since the quit-on-error & cordon-and-continue both work kind of the same, not technically but conceptually, I believe the above is fine. There would be no problem at all if we didn't have the `status` resource, but since we have, I guess it must have been added for a reason of managing statuses, hence I feel sort of obliged to use it, if not out of staying true to the designed, than for clarity. One may say, the `status` resource was added specifically for getting status from corresponding Chart CRs, but then I would counter this with the question of why other resources, like `configmap` cancel in case of errors and pass down statuses to this resource? `¯\_(ツ)_/¯`

## Checklist


mproffitt (approved):
Good work :-)

